PR #2, opened 2015-02-06
Updating translate task -sourcepath

jonnolen:
if you define custom matchers in your test tree that depend on a  tip in your source tree then the j2obcTranslate task would fail because only the source tree was set as the -sourcepath for the j2objc script (even though all java files are passed to it.)  this change updates the -sourcepath according to the j2objc documentation to be "source_tree:test_tree"


brunobowden:
Thanks for the change @jonnolen

brunobowden:
Merged manually, published in latest version. I also fixed the same issue for CycleFinder.

Thanks @jonnolen

PR #10, opened 2015-02-12
Added new configuration property buildSourceDir

confile:
When you use java annotation processing like in Dagger2 you need to specify where the additional generated source is located. 

Using ``buildSourceDir`` in your config lets you define it.

There one problem you might be able to solve quickly. When doing ``gradle j2objcTranslate`` the second time it only restarts the task if sources in ``src/mani/java``have been change. My buildSourceDir is: 

```
buildSourceDir "${projectDir}/build/source/apt"
```

The translation task is not done a second time if files have changed. Could you fix that?

### DIFF
--- a/j2objc.gradle
+++ b/j2objc.gradle
@@ -83,8 +83,8 @@ class J2objcPluginExtension {
     // Where to copy generated files (excludes test code and executable)
     String destDir = null
 
-	// Additional sources that should be included on the classpath, e.g., java annotation processor source files
-	String buildSourceDir = null 
+    // Additional sources that should be included on the classpath, e.g., java annotation processor source files
+    String buildSourceDir = null 
 
     // Path to j2objc distribution
     String j2objcHome = null
@@ -320,12 +320,12 @@ class J2objcTranslateTask extends DefaultTask {
         println "Deleting j2objc build dir: " + destDir.path
         project.delete destDir.path
 
-		def buildSourceDir = project.j2objcConfig.buildSourceDir
-		println "buildSourceDir: "+buildSourceDir		
-		if (buildSourceDir) {
-			def buildSrcFiles = project.files(project.fileTree(dir: buildSourceDir, includes: ["**/*.java"]))
-			srcFiles += buildSrcFiles
-		}
+	def buildSourceDir = project.j2objcConfig.buildSourceDir
+	println "buildSourceDir: "+buildSourceDir		
+	if (buildSourceDir) {
+		def buildSrcFiles = project.files(project.fileTree(dir: buildSourceDir, includes: ["**/*.java"]))
+		srcFiles += buildSrcFiles
+	}
 
         srcFiles = J2objcUtils.fileFilter(srcFiles,
                 project.j2objcConfig.translateIncludeRegex,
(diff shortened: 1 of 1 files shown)


j2objc.gradle
```diff
@@ -83,6 +83,9 @@ class J2objcPluginExtension {
     // Where to copy generated files (excludes test code and executable)
     String destDir = null
 
+	// Additional sources that should be included on the classpath, e.g., java annotation processor source files
+	String buildSourceDir = null 
+
```
brunobowden: Definitely use spaces instead of tabs. Tabs can be very inconsistent across platforms.

j2objc.gradle
```diff
@@ -317,6 +320,13 @@ class J2objcTranslateTask extends DefaultTask {
         println "Deleting j2objc build dir: " + destDir.path
         project.delete destDir.path
 
+		def buildSourceDir = project.j2objcConfig.buildSourceDir
+		println "buildSourceDir: "+buildSourceDir		
+		if (buildSourceDir) {
+			def buildSrcFiles = project.files(project.fileTree(dir: buildSourceDir, includes: ["**/*.java"]))
+			srcFiles += buildSrcFiles
+		}
+
```
brunobowden: Tabs => Spaces

j2objc.gradle
confile: okay thanks for the hint

confile:
Okay now it should be fine.

brunobowden:
@confile - thanks for the change, I need to get dagger working so I can
properly understand the change. I also want to make sure that it works
correctly with the gradle dependencies. I'm not sure if generated code
should be part of @InputFiles on each task - something that gradle uses to
check when files are out of date. My sense is that this shouldn't be an
issue as they're generated but that's what I'm trying to figure out.

confile:
Okay let me know when you get dagger2 running. From my point of view it is only a small step left. I posted you the configuration you need to get it running. Just use the atp plugin and it works and be sure to use the dagger sources with dependencies not the one without.

brunobowden:
I'm concerned the apt plugin may be an issue.... I'm using a shared project
without any android dependencies and I believe that apt depends on that. I
will get this working ;-)

j2objc.gradle
```diff
@@ -83,6 +83,9 @@ class J2objcPluginExtension {
     // Where to copy generated files (excludes test code and executable)
     String destDir = null
 
+    // Additional sources that should be included on the classpath, e.g., java annotation processor source files
+    String buildSourceDir = null
```
brunobowden: Rename:
buildSourceDir => generatedSourceDir

Change comment to:
   // Only generated source files, e.g. from dagger annotations. The script will
   // ignore changes in this directory so they must be limited to generated files.

brunobowden:
@confile - I had one change for you on naming and then I'll merge the request.

confile:
@brunobowden Okay I did the change. Did you make any progress on Dagger2 integration?

brunobowden:
Can you also change the comment as well - I put in a note about indicating
theat it's ONLY generated files, so people don't add other sources by
mistake.

brunobowden:
I haven't made progress on dagger. I'm being pulled away to something else
for a week or two and then I'll return. Annoying I haven't got it up and
running yet.

confile:
@brunobowden Comment changed

confile:
Thank you!

PR #11, opened 2015-02-12
cycleFinderExpectedCycles to warn on adding new cycles

brunobowden:
@confile - please take a look

This limits the cycleFinderTask so that if a new cycle is added then the build will fail.

It defaults to 40 cycles, which is what naturally occurs with the java runtime environment.

Instructions are also provided for doing the special j2objc build of the java runtime environment which eliminates all reported cycles by code modification and cycle whitelist.

### DIFF
--- a/j2objc.gradle
+++ b/j2objc.gradle
@@ -59,7 +59,7 @@
  *
  * Commands:
  * Each one depends on the previous command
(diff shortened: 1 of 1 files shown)


brunobowden:
I agree it's shorter groovy. I want to make the "== null" explicit so that another engineer looking at it will understand exactly what it means.

confile:
I am fine with that.

PR #13, opened 2015-02-15
Generate universal static libraries for iphone and simulator.

daschewie:
The task "j2objcLink" was added to the j2objc.gradle plugin. This
task allows the creation of static libraries/archives for linking against
iphone 32bit, iphone 64bit, and simulator i386 binaries.

Do the following to use the new task:

- Configure the j2objcConfig.compileFlags (i.e. "" for standard j2objc)

- Configure the j2objcConfig.distributionDir for output. This folder
  will contain include/ and lib/ with headers and universal lib.

- Optional configure j2objcConfig.staticLibName (i.e. "library.a")

### DIFF
--- a/j2objc.gradle
+++ b/j2objc.gradle
@@ -136,6 +136,10 @@ class J2objcPluginExtension {
     String testIncludeRegex = null
     // Warn if no tests are executed
     boolean testExecutedCheck = true
+
+    // Define static archive name for j2objcLink
+    String staticLibName = "library.a"
+    String distributionDir = null
 }
 
 
@@ -417,6 +421,121 @@ class J2objcTranslateTask extends DefaultTask {
     }
 }
 
+// TODO: Add better error handling
+// TODO: Remove dependency of sysroot_path.sh script.
+// TODO: compileFlags must be set to blank "", we get linker errors otherwise?
+// TODO: Document this feature.
+class J2objcLinkTask extends DefaultTask {
+    @InputDirectory File srcDir
+    @OutputDirectory File destDir
+
+    @TaskAction
+    def link() {
+        def flagMap = [:]
+        def output = new ByteArrayOutputStream()
+        def sysroot= ""
+        def script = project.file("scripts/sysroot_path.sh").path
+
+        // Add iphone build flags
+        project.exec {
+            executable script
+            args "--iphoneos"
+            standardOutput = output
+        }
+        sysroot = output.toString()
+        flagMap['armv7'] = "-arch armv7 -miphoneos-version-min=5.0 -isysroot ${sysroot}"
+        flagMap['arm64'] = "-arch arm64 -miphoneos-version-min=5.0 -isysroot ${sysroot}"
+        flagMap['armv7s'] = "-arch armv7s -miphoneos-version-min=5.0 -isysroot ${sysroot}"
+
+        // Add iphone sumulator build flags
+        output = new ByteArrayOutputStream()
+        project.exec {
+            executable script
+            args "--iphonesimulator"
+            standardOutput = output
+        }
+        sysroot = output.toString()
+        flagMap['i386'] = "-arch i386 -miphoneos-version-min=5.0 -isysroot ${sysroot}"
+
+        // Compile all platforms
+        def binary = J2objcUtils.j2objcHome(getProject()) + "/j2objcc"
+        for (arch in flagMap.keySet()) {
+            println "Building ${arch} object files."
+            def archDir = new File(destDir, arch)
+            if (!archDir.exists()) {
+                archDir.mkdirs()
+            }
+
+            project.exec {
+                workingDir "${archDir.path}"
+                executable binary
+                args "-I${srcDir}"
+                args "-c"
+
+                args flagMap[arch]
+                args "${project.j2objcConfig.compileFlags}".split()
+
+                def srcFiles = project.files(project.fileTree(
+                    dir: srcDir, includes: ["**/*.mm",  "**/*.m"]))
+                srcFiles.each { file ->
+                    args file.path
+                }
+            }
+
+            // Create arch specific archives 
+            println "Building ${arch} archive."           
+            project.exec {
+                workingDir "${archDir.path}"
+                commandLine "ar"
+                args  "rcs"
+                args  "${project.j2objcConfig.staticLibName}"
+
+                def srcFiles = project.files(project.fileTree(
+                    dir: archDir, includes: ["**/*.o"]))
+                srcFiles.each { file ->
+                    args file.path
+                }
+            }
+        }
+
+        def distDir = new File("${project.j2objcConfig.distributionDir}")
+        def includeDir = new File(distDir, "include")
+        def libDir = new File(distDir, "lib")
+        if (!includeDir.exists()) {
+            includeDir.mkdirs()
+        }
+        if (!libDir.exists()) {
+            libDir.mkdirs()
+        }
+
+        println "Copying header files to ${includeDir.path}"
+        project.copy {
+            includeEmptyDirs = false
+            from destDir
+            into includeDir
+            include "**/*.h"
+        }
+
+        println "Merging Universal archive to ${libDir.path}"   
+        project.exec {
+            workingDir "${libDir.path}"
+            commandLine "lipo"
+            args "-create"
+
+            for (arch in flagMap.keySet()) {
+                def archDir = new File(destDir, arch)
+                def archLib = new File(archDir, "${project.j2objcConfig.staticLibName}" )
+                if (archLib.exists()) {
+                    args "${archLib.path}"
+                }                
+            }
+
+            args "-output"
+            args "${project.j2objcConfig.staticLibName}"
+        }
+    }
+}
+
 
 class J2objcCompileTask extends DefaultTask {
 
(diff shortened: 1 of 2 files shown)


scripts/sysroot_path.sh
```diff
@@ -0,0 +1,73 @@
...
+
+  SDK_PATH=$(xcodebuild -sdk ${SDK_TYPE} -find-library system | sed 's/\/usr\/lib\/libsystem.dylib//')
+fi
+
+if [ "x${SDK_PATH}" = "x" ]; then
+  echo "No iOS SDKs located."
+  exit 1;
+fi
+echo ${SDK_PATH}
+exit 0
```
daschewie: This script was copied from the j2objc project. Eventually I would like to do away with it, and add the functionality directly to the gradle plugin.

j2objc.gradle
```diff
@@ -623,6 +742,13 @@ class j2objc implements Plugin<Project> {
                 destFile = file("${buildDir}/j2objc/testrunner")
             }
 
+            project.tasks.create(name: "j2objcLink", type: J2objcLinkTask,
+                    dependsOn: 'j2objcTranslate') {
```
brunobowden: Should be: ```dependsOn: 'j2objcTest'```. So if the tests fail then no library is built. Also, place it after the j2objcCopy task.

It sounds like it should either Copy or Link but not both? Also, should it be copying the library to the ios directory where it can be committed to the source repository? The expectation is that a lead engineer or build system will update the repo with the new binary, so that most iOS engineers don't need to run the j2objc.gradle script but automatically get the updates.

scripts/sysroot_path.sh
```diff
@@ -0,0 +1,73 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Returns the location of the latest installed Xcode Mac OS X, iPhoneOS,
+# or iPhoneSimulator SDK root directory.
+#
+# Usage: sysroot_path [--iphoneos | --iphonesimulator]
+
```
brunobowden: Add comment on where the file was originally from. So someone can look at the original and perhaps get updates. Also note if there were any changes to the original? If it's exactly the same then say "Unmodified Feb 15, 2015 copy of XXXX".

brunobowden: At comment on source:
Add note:

# Copied from j2objc repository as of Feb 13, 2015. No modifications made:
# https://github.com/google/j2objc/blob/master/scripts/sysroot_path.sh

j2objc.gradle
```diff
@@ -136,6 +136,10 @@ class J2objcPluginExtension {
     String testIncludeRegex = null
     // Warn if no tests are executed
     boolean testExecutedCheck = true
+
+    // Define static archive name for j2objcLink
+    String staticLibName = "library.a"
```
brunobowden: I believe settings should be:
boolean staticLibCreate = false
String staticLibName = "shared.a"
String staticLibDir = null

Do you also want to add another setting to disable the j2objcCopy task?

I think there's a growing convention around "shared" as the indicator for the shared code so use that.

Useful logic for first time users:
1) Default for translation only and no static lib (we might make the static lib the default in the future)
2) If (staticLibCreate && staticLibDir == null) {
    throw new InvalidUserDataException("staticLibDir path needs to be defined...")
}

j2objc.gradle
```diff
@@ -417,6 +421,121 @@ class J2objcTranslateTask extends DefaultTask {
...
+class J2objcLinkTask extends DefaultTask {
+    @InputDirectory File srcDir
+    @OutputDirectory File destDir
+
+    @TaskAction
+    def link() {
+        def flagMap = [:]
+        def output = new ByteArrayOutputStream()
+        def sysroot= ""
+        def script = project.file("scripts/sysroot_path.sh").path
```
brunobowden: I'm not sure if this always works. At the moment, I'm expecting people to put j2objc.gradle and perhaps the sysroot_path.sh script in the top of the project folder... and not is a scripts subdirectory. For now, you want to find out where the j2objc.gradle file is located and look next to that. In the future, it will be much better when this is all wrapped up in to a plugin.

j2objc.gradle
```diff
@@ -417,6 +421,121 @@ class J2objcTranslateTask extends DefaultTask {
...
+
+        // Add iphone build flags
+        project.exec {
+            executable script
+            args "--iphoneos"
+            standardOutput = output
+        }
+        sysroot = output.toString()
```
brunobowden: Put this in the J2objcUtils class. I think extracting the method makes it easier to understand.

Within that method, add an example of what the output may be:
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.9.sdk

j2objc.gradle
```diff
@@ -417,6 +421,121 @@ class J2objcTranslateTask extends DefaultTask {
...
+        flagMap['armv7'] = "-arch armv7 -miphoneos-version-min=5.0 -isysroot ${sysroot}"
```
brunobowden: Avoid the smell of duplicated code, so place " -miphoneos-version-min=5.0 -isysroot ${sysroot}" in a common variable, e.g.:

def archSuffix = " -miphoneos-version-min=5.0 -isysroot ${sysroot}"
flagMap['arm64'] = "-arch arm64" + archSuffix
flagMap['armv7'] = "-arch armv7" + archSuffix
...

j2objc.gradle
```diff
@@ -417,6 +421,121 @@ class J2objcTranslateTask extends DefaultTask {
...
+        project.exec {
+            executable script
+            args "--iphonesimulator"
+            standardOutput = output
+        }
+        sysroot = output.toString()
+        flagMap['i386'] = "-arch i386 -miphoneos-version-min=5.0 -isysroot ${sysroot}"
+
+        // Compile all platforms
+        def binary = J2objcUtils.j2objcHome(getProject()) + "/j2objcc"
```
brunobowden: Rename, so reading the code elsewhere, someone can guess the contents of the variable without needing this line:

binary => j2objccBin

j2objc.gradle
```diff
@@ -417,6 +421,121 @@ class J2objcTranslateTask extends DefaultTask {
...
+        for (arch in flagMap.keySet()) {
+            println "Building ${arch} object files."
+            def archDir = new File(destDir, arch)
+            if (!archDir.exists()) {
+                archDir.mkdirs()
+            }
+
+            project.exec {
```
brunobowden: This shares a lot of code with the project.exec inside the compile task. The only difference is the addition of the flagMap[arch] flag. Accordingly this should be converted to a shared method in J2objcUtils.compile(...)

    @TaskAction
    def link() {
        ...
        for (arch in flagMap.keySet()) {
            J2objcUtils.compile(getProject(), srcDir, destDir, flagMap[arch] + " " + project.j2objcConfig.compileFlags)
        }
    }

Then the compile task:

    @TaskAction
    def compile() {
        ...
        J2objcUtils.compile(getProject(), srcDir, destDir, project.j2objcConfig.compileFlags)
    }

Extracted Method:

    J2objcUtils.compile(Project project, File srcDir, File destDir, String extraFlags) {
        def j2objccBin = J2objcUtils.j2objcHome(project) + "/j2objcc"
        project.exec {
            executable j2objccBin
            args "-I${srcDir}"
            args "-o", "${destFile.path}"

            args extraFlags.split()

            def srcFiles = project.files(project.fileTree(
                    dir: srcDir, includes: ["**/*.h", "**/*.m"]))
            srcFiles.each { file ->
                args file.path
            }
        }
    }

j2objc.gradle
```diff
@@ -417,6 +421,121 @@ class J2objcTranslateTask extends DefaultTask {
...
+                srcFiles.each { file ->
+                    args file.path
+                }
+            }
+
+            // Create arch specific archives 
+            println "Building ${arch} archive."           
+            project.exec {
+                workingDir "${archDir.path}"
+                commandLine "ar"
```
brunobowden: Add comment what "ar" is

j2objc.gradle
```diff
@@ -417,6 +421,121 @@ class J2objcTranslateTask extends DefaultTask {
...
+                args  "rcs"
```
brunobowden: Add comment what "rcs" is

j2objc.gradle
```diff
@@ -417,6 +421,121 @@ class J2objcTranslateTask extends DefaultTask {
...
+            standardOutput = output
+        }
+        sysroot = output.toString()
+        flagMap['i386'] = "-arch i386 -miphoneos-version-min=5.0 -isysroot ${sysroot}"
+
+        // Compile all platforms
+        def binary = J2objcUtils.j2objcHome(getProject()) + "/j2objcc"
+        for (arch in flagMap.keySet()) {
+            println "Building ${arch} object files."
+            def archDir = new File(destDir, arch)
```
brunobowden: I think this should be under buildDir as it's a temporary creation and you want it to get deleted during a ```clean``` operation.

j2objc.gradle
```diff
@@ -417,6 +421,121 @@ class J2objcTranslateTask extends DefaultTask {
...
+        println "Merging Universal archive to ${libDir.path}"   
+        project.exec {
+            workingDir "${libDir.path}"
+            commandLine "lipo"
+            args "-create"
+
+            for (arch in flagMap.keySet()) {
+                def archDir = new File(destDir, arch)
+                def archLib = new File(archDir, "${project.j2objcConfig.staticLibName}" )
+                if (archLib.exists()) {
```
brunobowden: If archLib doesn't exist then it should fail... as that indicates something that should've been built but wasn't. Don't hide the error by ignoring it if it doesn't exist.

brunobowden:
Thanks @daschewie. I've reviewed all the changes and written up feedback. After you've gone through this, I'll try running it myself. This is a useful addition to the script.

daschewie:
@brunobowden , thanks for the feedback. I will be addressing these issues in the next couple of days.

brunobowden:
@daschewie - let me know when it's in good shape to merge it. Best not to leave this hanging for much longer. As master moves along, it'll make merging more complex for you.

brunobowden:
@daschewie - it would be great to merge these changes. There's already some merge conflicts and it'll get more painful as time goes by.

brunobowden:
@daschewie - we've moved in a different direction to use Cocoapods, so we're going to abandon this pull request. It was useful when we first got started. If you'd like to get involved with what we're working on now, then please reach out.

PR #19, opened 2015-04-03
New Task J2objcXcodeTask

confile:
This tasks links the generated objective-c files using https://github.com/developertown/integratej2objc in your Xcode project.

### DIFF
--- a/j2objc.gradle
+++ b/j2objc.gradle
@@ -636,21 +636,21 @@ class J2objcLinkTask extends DefaultTask {
     def link() {
 		
 	if (project.j2objcConfig.xcodeProjectDir == null) {
-		def message = "You must specify the xcodeProjectDir"
-		throw new InvalidUserDataException(message)
+	    def message = "You must specify the xcodeProjectDir"
+	    throw new InvalidUserDataException(message)
 	}
 	
 	if (project.j2objcConfig.xcodeJ2objcGeneratedDir == null) {
-		def message = "You must specify the xcodeJ2objcGeneratedDir"
-		throw new InvalidUserDataException(message)
+	    def message = "You must specify the xcodeJ2objcGeneratedDir"
+	    throw new InvalidUserDataException(message)
 	}
 	if (project.j2objcConfig.xcodeTarget == null) {
-		def message = "You must specify the xcodeTarget"
-		throw new InvalidUserDataException(message)
+	    def message = "You must specify the xcodeTarget"
+	    throw new InvalidUserDataException(message)
 	}
 	if (project.j2objcConfig.xcodeProject == null) {
-		def message = "You must specify the xcodeProject"
-		throw new InvalidUserDataException(message)
+	    def message = "You must specify the xcodeProject"
+	    throw new InvalidUserDataException(message)
 	}
 		
 	def xcodeProjectDir = project.j2objcConfig.xcodeProjectDir
@@ -662,16 +662,16 @@ class J2objcLinkTask extends DefaultTask {
 	def integratej2objcExec = "integratej2objc"
 		
 	try {
-		project.exec {
-			executable integratej2objcExec
+	    project.exec {
+		executable integratej2objcExec
 
-				args "integrate_source"
-				args "-p", "${xcodeProjectDir}" 
-				args "-x", "${xcodeProject}"
-				args "-s", "${srcFolder}"
-				args "-g", "${xcodeJ2objcGeneratedDir}"
-				args "-t", "${xcodeTarget}"
-			}
+		    args "integrate_source"
+		    args "-p", "${xcodeProjectDir}" 
+		    args "-x", "${xcodeProject}"
+		    args "-s", "${srcFolder}"
+		    args "-g", "${xcodeJ2objcGeneratedDir}"
+		    args "-t", "${xcodeTarget}"
+	    }
 
 	} catch (e) {
 		throw e
@@ -679,7 +679,7 @@ class J2objcLinkTask extends DefaultTask {
 		
 	def srcFilesSize = project.files(project.fileTree(dir: srcDir)).getFiles().size()
 	println "Linked ${srcFilesSize} files to Xcode project ${xcodeProject}"
-	}
+    }
 }
 
 
--- a/j2objc.gradle
+++ b/j2objc.gradle
@@ -674,7 +674,10 @@ class J2objcLinkTask extends DefaultTask {
 
 		} catch (e) {
 			throw e
-		}		
+		}
+		
+		def srcFilesSize = project.files(project.fileTree(dir: srcDir)).getFiles().size()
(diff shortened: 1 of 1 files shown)


j2objc.gradle
```diff
@@ -140,6 +140,17 @@ class J2objcPluginExtension {
     String testIncludeRegex = null
     // Warn if no tests are executed
     boolean testExecutedCheck = true
+    
+    // LINK
+    // the directory of the target Xcode project
+    String xcodeProjectDir = null
+    // the group where the objective-c files should be inserted to, e.g., "generated_source"
+    // note that the group will be deleted on every execution of the link command
+    String xcodeGroup = null
```
brunobowden: I prefer to minimize the setup, so someone can more easily start using the system. In this case, setting defaults. For example, I

```
// J2objc generated files are copied here, relative path to xcodeProjectDir
// Directory contents will be deleted on every execution of the link command
String xcodeJ2objcGeneratedDir = "j2objc-generated"
```

j2objc.gradle
```diff
@@ -140,6 +140,17 @@ class J2objcPluginExtension {
     String testIncludeRegex = null
     // Warn if no tests are executed
     boolean testExecutedCheck = true
+    
+    // LINK
+    // the directory of the target Xcode project
```
brunobowden: Avoid "the" at start of sentence, when someone is scanning, assume for the moment that they only read the first word - is that word descriptive or helpful? The goal is that the first word or few words is as helpful as possible so that they can then decide if they want to read the rest of the sentence.

Can the default be sensibly set to a folder named "Xcode" within the project? I'd like to setup good defaults for all of this. Makes setup easier for new projects and reduces variation.

// Xcode project directory, e.g. ${projectDir}/Xcode

j2objc.gradle
```diff
@@ -140,6 +140,17 @@ class J2objcPluginExtension {
...
+    String xcodeProjectDir = null
+    // the group where the objective-c files should be inserted to, e.g., "generated_source"
+    // note that the group will be deleted on every execution of the link command
+    String xcodeGroup = null
+    // the Xcode target the generated files should be linked to
+    String xcodeTarget = null
+    // the name of the target Xcode project, e.g., MyProject.xcodeproj
+    String xcodeProject = null
```
brunobowden: Should this be automatically found? Same for the Target as well?

j2objc.gradle
```diff
@@ -613,6 +624,64 @@ class J2objcCopyTask extends DefaultTask {
 }
 
 
+/**
+ * Note: This task requires https://github.com/developertown/integratej2objc
+ * 
+ */
+class J2objcLinkTask extends DefaultTask {
+	
+	@InputDirectory File srcDir
```
brunobowden: Please change all indentation to 4 spaces instead of 8 spaces. Thanks.

j2objc.gradle
```diff
@@ -664,6 +733,12 @@ class j2objc implements Plugin<Project> {
                 description 'Depends on j2objc translation and test, copies to destDir'
                 srcDir = file("${buildDir}/j2objc")
             }
+            
+            project.tasks.create(name: 'j2objcLink', type: J2objcLinkTask,
+		dependsOn: 'j2objcTranslate') {
```
brunobowden: Should depend on 'j2objcCopy' as that's what copies the files. Also needs to synchronize the input and output between those two tasks.

Main point is that if the tasks fail then they fail in the best state possible. For example, if the translate works but the j2objcc compile or test fail, then the generated files shouldn't be copied (as they were clearly broken). Instead, you force the developer to fix the tests first and then and only then do you copy and finally configure the Xcode project.

j2objc.gradle
```diff
@@ -613,6 +624,64 @@ class J2objcCopyTask extends DefaultTask {
...
+				args "integrate_source"
+				args "-p", "${xcodeProjectDir}" 
+				args "-x", "${xcodeProject}"
+				args "-s", "${srcFolder}"
+				args "-g", "${xcodeGroup}"
+				args "-t", "${xcodeTarget}"
+			}
+
+		} catch (e) {
+			throw e
```
brunobowden: Extend error message with explanation of what happens if integratej2objc doesn't exist. It should identify that case and explain the install and path setup to make it work. There are other examples in this file where it intercepts errors and then changes them.

brunobowden:
Hi @confile,
I'll have more feedback but will get back to that sometime tomorrow.

confile:
@brunobowden What about your additional feedback?

j2objc.gradle
```diff
@@ -613,6 +624,64 @@ class J2objcCopyTask extends DefaultTask {
 }
 
 
+/**
+ * Note: This task requires https://github.com/developertown/integratej2objc
+ * 
+ */
+class J2objcLinkTask extends DefaultTask {
```
brunobowden: Rename `J2objcLinkTask` => `J2objcXcodeTask` as I think that makes it's purpose clearer

brunobowden:
Thanks @Confile - just added a comment about changing the name to
J2objcXcodeTask as I think that's clearer than `Link`. Please address the
other feedback and I'll merge it pronto.

j2objc.gradle
confile: Are you sure about the dependency on ``j2objcCopy``? I think it is better to only link the files instead of copying them.

confile:
@brunobowden Done except one. You have to comment on that!

j2objc.gradle
```diff
@@ -664,6 +735,12 @@ class j2objc implements Plugin<Project> {
                 description 'Depends on j2objc translation and test, copies to destDir'
                 srcDir = file("${buildDir}/j2objc")
             }
+            
+            project.tasks.create(name: 'j2objcXcode', type: J2objcXcodeTask,
+		dependsOn: 'j2objcTranslate') {
```
brunobowden: Should depend on J2objcCopyTask

j2objc.gradle
```diff
@@ -565,7 +577,7 @@ class J2objcTestTask extends DefaultTask {
 
 class J2objcCopyTask extends DefaultTask {
 
-    @InputDirectory File srcDir
+	@InputDirectory File srcDir
```
brunobowden: Replace tab with 4 spaces

brunobowden:
@confile - what did I need to comment on?

brunobowden:
Now I understand... it should be dependsOn: `j2objcCopyTask`

The thing I missed is that it should take the output of the copy task, not
the input.
So instead of integrating from ${buildDir}/j2objc, it should be
from project.j2objcConfig.destDir
What I'm not sure about is how to hook up the dependencies since this is
project config setting

confile:
@brunobowden okay I did the change. Any thing else?

brunobowden:
I've merged #19 - thanks for the changes.

I still have some other changes but I'll send you that in a separate
request.

PR #20, opened 2015-04-04
Make source jars in sourcepath available

confile:
Make source jars in sourcepath available. Using the translateSourcepaths property makes it possible to add additional source jars for translation.

### DIFF
--- a/j2objc.gradle
+++ b/j2objc.gradle
@@ -128,6 +128,10 @@ class J2objcPluginExtension {
     String translateIncludeRegex = null
     // TODO: consider moving to include(s) / exclude(s) structure as used for filetree
 
+    // addition source paths for the translation task, 
+    // e.g., ${projectDir}/libSrc/dagger-2.0-SNAPSHOT-sources.jar:${projectDir}/libSrc/javax.inject-1-sources.jar
+    String translateSourcepaths = null
+
     // COMPILE
     // Flags copied verbatim to j2objcc command
     String compileFlags = "-ObjC -lguava -ljavax_inject -ljre_emul -ljsr305 -ljunit -lmockito"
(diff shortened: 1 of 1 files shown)


j2objc.gradle
```diff
@@ -128,6 +128,10 @@ class J2objcPluginExtension {
     String translateIncludeRegex = null
     // TODO: consider moving to include(s) / exclude(s) structure as used for filetree
 
+    // addition source paths for the translation task, 
+    // e.g., ${projectDir}/libSrc/dagger-2.0-SNAPSHOT-sources.jar:${projectDir}/libSrc/javax.inject-1-sources.jar
+    String translateSourcepaths = null
+
```
brunobowden: The ideal is to make it easy to use dagger2, so people can easily adopt it. What's the best way to do this? For example, can the script detect that dagger2 is being used and automatically make the change you described in the comment.

If you think it's appropriate to wait until dagger2 is out of beta, then add a comment to do this later and how it would be best done.

j2objc.gradle
```diff
@@ -377,6 +381,12 @@ class J2objcTranslateTask extends DefaultTask {
         project.delete destDir.path
 
 	def sourcepath = "${project.file("src/main/java").path}:${project.file("src/test/java").path}"
+	
+	// add additional sourcepaths
+	if (project.j2objcConfig.translateSourcepaths) {
+		println "Add to sourcepath: ${project.j2objcConfig.translateSourcepaths}"
+		sourcepath += ":${project.j2objcConfig.translateSourcepaths}"
+	}
```
brunobowden: I've noticed several repeated code sections that could have an extract method refactoring...

```
if (XXX) {
    println "Added: " + XXX
    return ":" + XXX
} else {
    return ""
}
```

brunobowden:
Lastly, just to confirm is this needed for the cycleFinder?

confile:
I do not think we need this for cycleFinder because We would not have any chance to change lib sources. If you include it for cycleFinder it would not matter if you find any cycles. You can't change them.

confile:
@brunobowden Still have have problems with translation of dependent source files for dagger2 (See: https://github.com/google/j2objc/issues/513) may be you have an idea what to do?

j2objc.gradle
confile: I did not see how to make this extern since the inner part is doing things different all the time. Considering the println and checking if the variable is set you are right.

Issue #12, opened 2015-02-12
Custom dependencies


PaNaVTEC:
Hi Bruno,
I'm trying to use the plugin but i find I need something more, in my case I have 2 modules: presentation and domain, Presentation depends on domain so I have in my build.gradle on presentation: 

```groovy
j2objcConfig {
    j2objcHome parent.ext.j2objcPath
    destDir iOsProjectPath + 'Presentation'
    testExecutedCheck false
}
```

But when I try to execute j2objcCopy it just crashes because does not find something of the classes that are on domain, I think I need to configure the classpath but there is no option to configure "j2objcConfig" with this, any hints? 

Thanks.

brunobowden:
I'm trying to understand this...

What error message do you see? Anything else helpful with --info? What
domain? What classes is it looking for?

brunobowden:
If the module uses a java plugin, it depends on "test" and if there's an
android plugin, it depends on "assemble". See the lines here:

https://github.com/brunobowden/j2objc-gradle/blob/master/j2objc.gradle#L640

            // Make sure the wider project builds successfully
            if (project.plugins.findPlugin('java')) {

project.tasks.findByName('j2objcCycleFinder').dependsOn('test')
            } else if
(project.plugins.findPlugin('com.android.application')) {

project.tasks.findByName('j2objcCycleFinder').dependsOn('assemble')
            } else {
                def message =
                        "j2objc plugin didn't find either 'java' or
'com.android.application'\n" +
                        "plugin (which was expected). When this is found,
the j2objc plugin\n" +
                        "will build and run that first to make sure the
project builds correctly.\n"
                        "This will not be done here as it can't be found."
                println message
            }

PaNaVTEC:
Both projects are java
Presentation depends on Domain and uses classes of the domain module,

presentation.build:
```groovy
dependencies {
    compile project(':domain')
}
```

I fix it as a temporary workarround adding a custom classpath arg to j2objc: 
/Users/panavtec/Documents/panavtecWorkspace/BBcleancontacts/domain/build/libs/domain.jar

I don't know if im explaining my problem well...

brunobowden:
I think I understand now. What you need to do ischange the code that
assembles the dependencies. At the moment, it will just collect the j2objc
jar's. You also need to add the other jar's.

The following code is used in the cycleFinder and translate tasks and need
the extra jar's:

                project.j2objcConfig.translateJ2objcLibs.each { library ->
[...]

PR #22, opened 2015-04-06
Automatically put java project dependencies on the sourcepath; #12

advayDev1:
Set appendProjectDependenciesToSourcepath = true to add the java srcDirs of all project dependencies to the sourcepath for j2objc.

### DIFF
--- a/j2objc.gradle
+++ b/j2objc.gradle
@@ -132,6 +132,10 @@ class J2objcPluginExtension {
     // e.g., ${projectDir}/libSrc/dagger-2.0-SNAPSHOT-sources.jar:${projectDir}/libSrc/javax.inject-1-sources.jar
     // TODO the script should detect that dagger2 is being used and automatically make the change to the sourcepath
     String translateSourcepaths = null
+    
+    // Set to true if java project dependencies of the current project should be appended to the sourcepath
+    // automatically.  You will most likely want to use --build-closure in the translateFlags as well.
+    boolean appendProjectDependenciesToSourcepath = false
 
     // COMPILE
     // Flags copied verbatim to j2objcc command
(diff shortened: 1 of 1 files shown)


confile:
+1 great idea

j2objc.gradle
```diff
@@ -394,6 +398,18 @@ class J2objcTranslateTask extends DefaultTask {
 	if (project.j2objcConfig.generatedSourceDir) {
 		sourcepath += ":${project.file(project.j2objcConfig.generatedSourceDir).path}"
 	}
+	
+	// Project depedencies.
+	if (project.j2objcConfig.appendProjectDependenciesToSourcepath) {
+		def depSourcePaths = []
```
brunobowden: Please use 4 spaces instead of 8

advayDev1: @brunobowden fixed in new squashed commit.  note however that this whole file has mixed tabs + spaces, which we should fix in a separate commit.

advayDev1: fixed in new squashed commit 395f7be7ba384a0a26b36e361c99c3dfc5aeee7b

brunobowden:
Thanks @advayDev1.

One thought, I haven't got a project with dependencies yet. Would it be sensible in the future to have this `true` by future?

advayDev1:
Yes in the future depending on how you version this code - it seems you
don't have release tags yet.  Also I'm trying to figure out how to do this
for all dependencies automatically, after that it should be default true.

confile:
If you use jar dependencies in your project they should only be added to the ``sourcepath`` if they are source jars.

confile:
@brunobowden You can merge the pull request.

PR #24, opened 2015-04-10
fix the $$ xcode bug, generate a file x$y for each x$$y files after

confile:
all files have been linked with xcode such that xcode finds the correct file.

### DIFF
--- a/j2objc.gradle
+++ b/j2objc.gradle
@@ -712,6 +712,20 @@ class J2objcXcodeTask extends DefaultTask {
 		
 	def srcFilesSize = project.files(project.fileTree(dir: srcDir)).getFiles().size()
 	println "Linked ${srcFilesSize} files to Xcode project ${xcodeProject}"
+	
+	// fix the $$ xcode bug, generate a file x$y for each x$$y files after 
+	// all files have been copied to Xcode
+	def srcFiles = project.files(project.fileTree(dir: srcDir)).getFiles()
+		
+	def col = srcFiles.findAll { file -> file.name.contains("\$\$") }
+	project.copy {
+	    from(col)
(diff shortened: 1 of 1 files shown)


confile:
See: http://stackoverflow.com/questions/29555225/xcode-no-such-file-or-directory-if-filename-contains-sign

j2objc.gradle
```diff
@@ -712,6 +712,21 @@ class J2objcXcodeTask extends DefaultTask {
 		
 	def srcFilesSize = project.files(project.fileTree(dir: srcDir)).getFiles().size()
 	println "Linked ${srcFilesSize} files to Xcode project ${xcodeProject}"
+	
+	// fix the $$ xcode bug, generate a file x$y for each x$$y files after
```
brunobowden: From the Stack Overflow answer, it appeared that for a x$$y file, it would add a x$$$$y file. You're adding a x$y file though, so I'm confused?

brunobowden: TODO(confile): remove once Xcode is able to correctly handle x$$y filename in compile.

Good that you included the reference to the SO issue

confile: @brunobowden When you have a file x$$y Xcode searches for x$y. I tested my approach with xcode and it worked fine.

confile:
@brunobowden The SO answer suggests the same. I think you misunderstood the answer. It approaches the problem from the other way round. If you want xcode to reference x$$y you need x$$$$y which is the same as when you have X$$$$y xcode will search for x$$y or another example when you have x$$y xcode searches for x$y.

brunobowden:
@confile - I have to merge this manually

confile:
ok

PR #27, opened 2015-04-13
Move testrunner to j2objcc directory

brunobowden:


brunobowden:
@confile - please take a look at this

confile:
I am okay with it

PR #29, opened 2015-04-13
integratej2objc to work with bundle

brunobowden:
- uses bundle for integratej2objc (prefer RubyGem in the future)
- comment improvements, particularly for dependency chains

### DIFF
--- a/j2objc.gradle
+++ b/j2objc.gradle
@@ -284,6 +284,8 @@ class J2objcUtils {
 class J2objcCycleFinderTask extends DefaultTask {
 
     @InputFiles FileCollection srcFiles
+    // TODO: declare @OutputXXX so gradle knows if task is up to date
+    // TODO: http://stackoverflow.com/questions/29599011/up-to-date-gradle-task-status-when-it-has-no-output
 
     @TaskAction
     def cycleFinder() {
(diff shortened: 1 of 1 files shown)


brunobowden:
@confile, @jonnolen @formvoltron

Please take a look. This gets integratej2objc working for now.

@jonnolen - please update the install instructions for now:
https://github.com/developertown/integratej2objc/#installation

formvoltron:
I'll give it a try, just one more thing I hit was setting the use-arc flag.  I'd read it somewhere else and could figure out where to change it, but it seems like something most people will want to set and so may as well expose that in your suggested settings that get placed in the build.gradle file.

formvoltron:
gradle --stacktrace j2objcXcode

It says "build successful", but I also get this output:

ERROR: "integratej2objc integrate_source" was called with arguments ["-p", "/Users/mattpease/Documents/basic"]
Usage: "integratej2objc integrate_source -g, --group=GROUP -s, --source-root=SOURCE_ROOT -t, --target=TARGET -x, --xcodeproj=XCODEPROJ"
Linked 20 files to Xcode project basic.xcodeproj

formvoltron:
There does not seem to be a -p option on the integratej2objc tool.

*     j2objcXcode       - Xcode project file import and target configuration
Also, based on the description inside j2objc.gradle, I'd expect the j2objcXcode target to ensure that Xcode is setup properly with link flags, etc, as described here:
https://github.com/google/j2objc/wiki/Xcode-Build-Rules#update-the-build-settings

 It does not seem to do that.  Or is that what integratej2objc will do?

confile:
@brunobowden okay you can merge this.

confile:
@formvoltron the ``-p`` option is a little bit confusing. It still exist but was removed from the docu.

formvoltron:
@confile I was getting this error

ERROR: "integratej2objc integrate_source" was called with arguments ["-p", "/Users/mattpease/Documents/fred"]

PR #32, opened 2015-04-14
Use only one classpath tag in j2objc command and added...

confile:
translateClasspath to add classpath jar like this:

```
translateClasspath "json-20140107.jar", "somelib.jar"
```

### DIFF
--- a/j2objc.gradle
+++ b/j2objc.gradle
@@ -462,8 +462,8 @@ class J2objcTranslateTask extends DefaultTask {
             J2objcUtils.filenameCollisionCheck(srcFiles)
         }
         
-        String[] allLClassPathsArs = project.j2objcConfig.translateJ2objcLibs + project.j2objcConfig.translateClassPaths
-        def classPathsArg = J2objcUtils.getClassPathsArg(project, allLClassPathsArs)
+        String[] allClassPathsArgs = project.j2objcConfig.translateJ2objcLibs + project.j2objcConfig.translateClassPaths
+        def classPathsArg = J2objcUtils.getClassPathsArg(project, allClassPathsArgs)
         
         try {
             project.exec {
--- a/j2objc.gradle
+++ b/j2objc.gradle
@@ -447,7 +447,7 @@ class J2objcTranslateTask extends DefaultTask {
             J2objcUtils.filenameCollisionCheck(srcFiles)
         }
         
-        def classpathArgs = ""
+        def classPathsArg = ""
         def classPathList = []
         project.j2objcConfig.translateJ2objcLibs.each { library ->
             def libPath = J2objcUtils.j2objcHome(getProject()) + "/lib/" + library
@@ -457,7 +457,7 @@ class J2objcTranslateTask extends DefaultTask {
             def libPath = "${project.projectDir}/lib/" + library
             classPathList += libPath
         }    
-        classpathArgs += classPathList.join(':')
+        classpathArg += classPathList.join(':')
 
         try {
             project.exec {
--- a/j2objc.gradle
+++ b/j2objc.gradle
@@ -285,7 +285,7 @@ class J2objcUtils {
     }
     
     // Translates an array of library names into a javac conform -classpath args String
-    static def getClasspathsArg(Project proj, String[] libraries) {
+    static def getClassPathsArg(Project proj, String[] libraries) {
         def classPathsArg = ""
         def classPathList = []
         libraries.each { library ->
@@ -358,7 +358,7 @@ class J2objcCycleFinderTask extends DefaultTask {
                 project.j2objcConfig.translateIncludeRegex,
                 project.j2objcConfig.translateExcludeRegex)
 
-        def classPathsArg = J2objcUtils.getClasspathsArg(project, project.j2objcConfig.cycleFinderClassPaths)
+        def classPathsArg = J2objcUtils.getClassPathsArg(project, project.j2objcConfig.cycleFinderClassPaths)
 
         def output = new ByteArrayOutputStream()
         try {
@@ -463,7 +463,7 @@ class J2objcTranslateTask extends DefaultTask {
         }
         
         String[] allLClassPathsArs = project.j2objcConfig.translateJ2objcLibs + project.j2objcConfig.translateClassPaths
-        def classPathsArg = J2objcUtils.getClasspathsArg(project, allLClassPathsArs)
+        def classPathsArg = J2objcUtils.getClassPathsArg(project, allLClassPathsArs)
         
         try {
             project.exec {
--- a/j2objc.gradle
+++ b/j2objc.gradle
@@ -461,9 +461,15 @@ class J2objcTranslateTask extends DefaultTask {
         if (project.j2objcConfig.filenameCollisionCheck) {
             J2objcUtils.filenameCollisionCheck(srcFiles)
         }
+     
+        def classPathsArg = J2objcUtils.getClassPathsArg(project, project.j2objcConfig.translateClassPaths)
         
-        String[] allClassPathsArgs = project.j2objcConfig.translateJ2objcLibs + project.j2objcConfig.translateClassPaths
-        def classPathsArg = J2objcUtils.getClassPathsArg(project, allClassPathsArgs)
+        // j2objc default libraries
+        def classPathList = []
+        project.j2objcConfig.translateJ2objcLibs.each { library ->
+            classPathList += j2objcHome(proj) + "/lib/" + library
+        }
+        classPathsArg += classPathList.join(':')
         
         try {
             project.exec {
--- a/j2objc.gradle
+++ b/j2objc.gradle
@@ -97,6 +97,8 @@ class J2objcPluginExtension {
     boolean cycleFinder = false
     // Flags copied verbatim to cycle_finder command
     String cycleFinderFlags = null
+    // Libraries in ${projectDir}/lib/ for classpath to cyclefinder command
+    String[] cycleFinderClassPaths = []
     // Expected number of cycles, defaults to all those found in JRE
     // TODO(bruno): convert to a default whitelist and change expected cyles to 0
     int cycleFinderExpectedCycles = 40
@@ -105,7 +107,7 @@ class J2objcPluginExtension {
     // TRANSLATE
     // Flags copied verbatim to j2objc command
     String translateFlags = "--no-package-directories"
-    // Libraries in ${projectDir}/lib/ for classpath to j2objc command
+    // Libraries in ${projectDir}/lib/ for classpath to j2objc command, e.g.: "json-20140107.jar", "somelib.jar"
     String[] translateClassPaths = []
     
     // Additional libraries that are part of the j2objc release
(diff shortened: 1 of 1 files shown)


confile:
This makes adding libs to classpath much more simple than adding the full path all the time.

brunobowden:
I'll get to this on Friday

formvoltron:
Hi guys.  Can j2objc-gradle automatically make an attempt at translating a dependency? 

For example, I've added a dependency on a simple url builder.  this class has no other dependencies:

so I add this dependency:
    compile 'com.jillesvangurp:urlbuilder:1.1'

However when I do

gradle j2objcTranslate

I get errors that those classes are not around.

gradle compile works just fine.

It would appear that this pull request has the user manually setup these jar files that will be translated.  

Wouldn't a more gradle way to do it be to have an j2objcTranslate dependency?

brunobowden:
I would welcome a patch for doing this.

For those with pending patches, I'll get to them tomorrow or Saturday.

brunobowden:
In general, I want to make the tool as automatic and easy to use as
possible. That's the best way for it to get wide adoption, which helps all
of us.

formvoltron:
I'd give it a shot if I had ever touched groovy or gradle before.  But I have not.  So I'll be a lazy programmer & wait & hope for someone else to tackle this.

Thank you for the clarification.

brunobowden:
Groovy is pretty easy to get in to. It's the gradle specific stuff that's
more involved. If you know how to change the j2objc command line, then it's
pretty easy to edit the code to do the same.

confile:
@formvoltron if you know Java you know Groovy. Java is a Subset of groovy. Everything from Java works in Groovy

formvoltron:
Thanks for the tips.   I need to ship an app update though, and it's been
slow going because I'm platinum plating the whole thing.  MVP was a
diversion, and then J2Objc, and so I've got to stop it there.  I'll finish
and ship the app, then maybe I'll circle back here.  Probably not though,
as I've got a big trip coming up in a couple weeks.

My "feature request" at the beginning of this thread doesn't make much
sense now that I've learned that j2objc can't take .class files as input.
You'd need the source files and the dependencies of the source.

Perhaps a more general tool that would be handy would be a j2objc
dependency analyzer that can tell you lickity split if a certain .jar file
can be used.  It would ensure that all dependencies are packages that
j2objc can handle.

confile:
@formvoltron please do not use this pull request for a discussion. Open an issue if there is a problem or something unclear.

j2objc.gradle
```diff
@@ -452,11 +467,10 @@ class J2objcTranslateTask extends DefaultTask {
                 args "-d", "${destDir}"
                 args "-sourcepath", sourcepath
 
-                project.j2objcConfig.translateJ2objcLibs.each { library ->
-                    def libPath = J2objcUtils.j2objcHome(getProject()) + "/lib/" + library
-                    args "-classpath", project.file(libPath).path
+                if (classPathList.size() > 0) {
```
brunobowden: Should be classpathArgs.size() > 0, since you're testing the very argument that's being added.

j2objc.gradle
```diff
@@ -105,6 +105,9 @@ class J2objcPluginExtension {
     // TRANSLATE
     // Flags copied verbatim to j2objc command
     String translateFlags = "--no-package-directories"
+    // Libraries in ${projectDir}/lib/ for classpath to j2objc command
+    String[] translateClasspath = []
```
brunobowden: Rename to translateClassPaths to make it plural and capitalize the "P". I think we should capitalize the P to make the code clearer, even though the command line is lowercase.

j2objc.gradle
```diff
@@ -443,6 +446,18 @@ class J2objcTranslateTask extends DefaultTask {
         if (project.j2objcConfig.filenameCollisionCheck) {
             J2objcUtils.filenameCollisionCheck(srcFiles)
         }
+        
+        def classpathArgs = ""
```
brunobowden: Rename classPathsArg to make it plural and capitalize the P

j2objc.gradle
```diff
@@ -443,6 +446,18 @@ class J2objcTranslateTask extends DefaultTask {
...
+        def classPathList = []
+        project.j2objcConfig.translateJ2objcLibs.each { library ->
+            def libPath = J2objcUtils.j2objcHome(getProject()) + "/lib/" + library
+            classPathList += project.file(libPath).path
+        }
+        project.j2objcConfig.translateClasspath.each { library ->
+            def libPath = "${project.projectDir}/lib/" + library
+            classPathList += libPath
+        }    
+        classpathArgs += classPathList.join(':')
```
brunobowden: Extract this lines 451-460 as a method and use the same code for J2objcTranslateTask AND the J2objcCycleFinderTask. This replaces lines 352-355 in J2objcCycleFinderTask.

brunobowden:
@confile - I've made comments, I'll merge right after you've resolved them

j2objc.gradle
brunobowden: Add example, e.g.: "json-20140107.jar", "somelib.jar"
Do these need to be source jars? If so, then add that in the comment as well.

confile:
@brunobowden I'm done!

brunobowden:
@confile - there were two other comments that I didn't see resolved:

Extract this lines 451-460 as a method and use the same code for J2objcTranslateTask AND the J2objcCycleFinderTask. This replaces lines 352-355 in J2objcCycleFinderTask.

Add example, e.g.: "json-20140107.jar", "somelib.jar"
Do these need to be source jars? If so, then add that in the comment as well.

confile:
Your where right. I missed that comment. Not its done.

j2objc.gradle
```diff
@@ -97,6 +97,8 @@ class J2objcPluginExtension {
     boolean cycleFinder = false
     // Flags copied verbatim to cycle_finder command
     String cycleFinderFlags = null
+    // Libraries in ${projectDir}/lib/ for classpath to cyclefinder command
+    String[] cycleFinderClassPaths = []
```
brunobowden: CycleFinder should use translateClassPaths - can you think of a situation where they would need to be different? So just eliminate cycleFinderClassPaths as a setting.

* translateIncludeRegex
* translateExcludeRegex
* translateJ2objcLibs

j2objc.gradle
```diff
@@ -105,6 +107,9 @@ class J2objcPluginExtension {
     // TRANSLATE
     // Flags copied verbatim to j2objc command
     String translateFlags = "--no-package-directories"
+    // Libraries in ${projectDir}/lib/ for classpath to j2objc command, e.g.: "json-20140107.jar", "somelib.jar"
```
brunobowden: Do these need to be source jars? If so, then add a comment on a new line explaining that.

j2objc.gradle
```diff
@@ -278,6 +283,18 @@ class J2objcUtils {
             return ""
         }
     }
+    
+    // Translates an array of library names into a javac conform -classpath args String
+    static def getClasspathsArg(Project proj, String[] libraries) {
```
brunobowden: getClasspathsArg => getClassPathsArg

j2objc.gradle
```diff
@@ -278,6 +283,18 @@ class J2objcUtils {
...
+        def classPathsArg = ""
+        def classPathList = []
+        libraries.each { library ->
+        def libPath = "${proj.projectDir}/lib/" + library
```
brunobowden: Indent 4 spaces

j2objc.gradle
```diff
@@ -443,7 +461,10 @@ class J2objcTranslateTask extends DefaultTask {
         if (project.j2objcConfig.filenameCollisionCheck) {
             J2objcUtils.filenameCollisionCheck(srcFiles)
         }
-
+        
+        String[] allLClassPathsArs = project.j2objcConfig.translateJ2objcLibs + project.j2objcConfig.translateClassPaths
```
brunobowden: allLClassPathsArs => allClassPathsArgs

Is there meant to be an uppercase "L" in the variable name? If it's meant to be "Local" or "Logical", then spell it out in full camel case.

j2objc.gradle
```diff
@@ -443,7 +461,10 @@ class J2objcTranslateTask extends DefaultTask {
...
+        def classPathsArg = J2objcUtils.getClasspathsArg(project, allLClassPathsArs)
```
brunobowden: getClassPathsArg should only apply to translateClassPaths and not translateJ2objcLibs? The translateJ2objcLibs are in the J2ObjC directory, which is an entirely different location.

So I think the logic should be:

    // -classpath flag generation for j2objc and custom libraries
    static def getClassPathsArg(Project proj, String[] libraries) {
        def classPathList = []
        // custom libraries
        libraries.each { library ->
            classPathList += "${proj.projectDir}/lib/" + library
        }
        // j2objc default libraries
        project.j2objcConfig.translateJ2objcLibs.each { library ->
            classPathList += j2objcHome(proj) + "/lib/" + library
        }
        return classPathList.join(':')
    }

brunobowden:
@confile - I think we're getting very close. There's a couple of needed corrections. Thanks for being understanding on the back and forth. It's now much better code.

j2objc.gradle
confile: @brunobowden Yes cycleFinder does not need source jars so there might be different dependencies.

If you think this is not important I will change it and set the translate classpath and sourcepath for cyclefinder.

j2objc.gradle
confile: This comment is for ``translateClassPaths`` not for source path.

confile:
@brunobowden No problem. Please tell me if I overlooked something. I have some problems with the intend they seems to work different in the online editor.

j2objc.gradle
confile: You are right. Thanks. I changed it.

j2objc.gradle
brunobowden: @confile - this is about making the whole process easier for people to setup. It sounds correct that CycleFinder doesn't need certain jars but since adding them is harmless, it's easier to make them the same. We can separate them out at a later point but I'd prefer to only do that when multiple people say "I really need this". Adding complexity makes it more likely for a project to fail or make it harder to change. So it's best to avoid complexity as much as possible.

Please look reworking the getClassPathsArg as I described in the previous comment:

    // -classpath flag generation for j2objc and custom libraries
    static def getClassPathsArg(Project proj, String[] libraries) {
        def classPathList = []
        // custom libraries
        libraries.each { library ->
            classPathList += "${proj.projectDir}/lib/" + library
        }
        // j2objc default libraries
        project.j2objcConfig.translateJ2objcLibs.each { library ->
            classPathList += j2objcHome(proj) + "/lib/" + library
        }
        return classPathList.join(':')
    }

confile: You are right!

Concerning the ``getClassPathsArg`` function I still did an update it is now:

```
    // Translates an array of library names into a javac conform -classpath args String
    static def getClassPathsArg(Project proj, String[] libraries) {
        def classPathsArg = ""
        def classPathList = []
        libraries.each { library ->
        def libPath = "${proj.projectDir}/lib/" + library
            classPathList += libPath
        }
        classPathsArg += classPathList.join(':')
        return classPathsArg
    }
```

confile:
@brunobowden I am done!

brunobowden:
Thanks @confile for sticking with this

confile:
@brunobowden Thank you for your great feedback. By the way. Could you add me on linkedIn. I sent you an request.

brunobowden:
@confile - please give me a url for your LinkedIn. I get a lot of requests,
so I reject anything I don't recognize. If you didn't put a not in the
request, I've probably already rejected it.

confile:
Here it is: https://de.linkedin.com/in/michaelgorski

PR #35, opened 2015-04-21
Refactor classPathArg method

brunobowden:
- classPathsArg => classPathArg
- move j2objc default libs to getClassPathArg(…)
- indentation


brunobowden:
@confile - please take a look

confile:
Looks good to me!

PR #36, opened 2015-04-22
incremental translation implemented

confile:
translation task now:

- detects source file changed in source files and generated source files, it only translates source files which have been changed since the last compile 
- removes .java source files and remove translated counterpart .m and .h files 

This pretty cool and makes development a lot faster.

### DIFF
--- a/j2objc.gradle
+++ b/j2objc.gradle
@@ -769,26 +769,16 @@ class J2objcXcodeTask extends DefaultTask {
         try {
             project.exec {
                 // TODO: executable integratej2objcExec
-//                executable "bundle"
-//
-//                args "exec"
-//                args "integratej2objc"
-//                args "integrate_source"
-//                args "-p", "${xcodeProjectDir}"
-//                args "-x", "${xcodeProject}"
-//                args "-s", "${srcFolder}"
-//                args "-g", "${xcodeJ2objcGeneratedDir}"
-//                args "-t", "${xcodeTarget}"
-              
-              // TEST
-                executable "integratej2objc"
+                executable "bundle"
+
+                args "exec"
+                args "integratej2objc"
                 args "integrate_source"
                 args "-p", "${xcodeProjectDir}"
                 args "-x", "${xcodeProject}"
                 args "-s", "${srcFolder}"
                 args "-g", "${xcodeJ2objcGeneratedDir}"
-                args "-t", "${xcodeTarget}"
-                
+                args "-t", "${xcodeTarget}"                
             }
 
         } catch (Exception exception) {
@@ -843,6 +833,7 @@ class j2objc implements Plugin<Project> {
                                 exclude: project.relativePath(buildDir)))
             }
 
+            // TODO @Bruno "build/source/apt" must be project.j2objcConfig.generatedSourceDir no idea how to set it there
             // Dependency may be added in project.plugins.withType for Java or Android plugin
             tasks.create(name: "j2objcTranslate", type: J2objcTranslateTask,
                     dependsOn: 'j2objcCycleFinder') {
@@ -851,8 +842,8 @@ class j2objc implements Plugin<Project> {
                         fileTree(dir: projectDir,
                                 include: "**/*.java",
                                 exclude: project.relativePath(buildDir)) +
-                        fileTree(dir: "build/source/apt",     // TODO @Bruno must be project.j2objcConfig.generatedSourceDir no idea how to set it there
-                            include: "**/*.java")
+                        fileTree(dir: "build/source/apt",
+                                include: "**/*.java")
                         )
                 destDir = file("${buildDir}/j2objc")
             }
@@ -879,7 +870,7 @@ class j2objc implements Plugin<Project> {
             }
             
             project.tasks.create(name: 'j2objcXcode', type: J2objcXcodeTask,
-                dependsOn: 'j2objcTranslate') {
+                dependsOn: 'j2objcCopy') {
                 description 'Depends on j2objc translation, link generated files to Xcode project'
                 srcDir = file("${buildDir}/j2objc")
             }
--- a/j2objc.gradle
+++ b/j2objc.gradle
@@ -431,21 +431,17 @@ class J2objcTranslateTask extends DefaultTask {
         println "Source files: "+srcFiles.getFiles().size()
         FileCollection srcFilesChanged = project.files()
         inputs.outOfDate { change ->
-        //  println "New or Updated file: "+change.file
-          srcFilesChanged += project.files(change.file)
+            println "New or Updated file: "+change.file
+            srcFilesChanged += project.files(change.file)
         }
         def removedFileNames = []
         inputs.removed { change ->
-          println "Removed file: "+change.file.name
-          def nameWithoutExt = file.name.toString().replaceFirst("\\..*","")
-          removedFileNames += nameWithoutExt
+            println "Removed file: "+change.file.name
+            def nameWithoutExt = file.name.toString().replaceFirst("\\..*","")
+            removedFileNames += nameWithoutExt
         }
         println "Removed files: "+removedFileNames.size()
         
-        boolean freshBuild = false
-        if (srcFilesChanged.getFiles().size() == srcFiles.getFiles().size()) {
-          freshBuild = true
-        }
         println "New or Updated files: "+srcFilesChanged.getFiles().size()
         FileCollection translatedSrcFiles = srcFiles - srcFilesChanged
         println "Unchanged files: "+translatedSrcFiles.getFiles().size()
@@ -457,10 +453,10 @@ class J2objcTranslateTask extends DefaultTask {
             
             // remove translated .h and .m files which has no corresponding .java files anymore
             destFiles.each {File file ->
-              def nameWithoutExt = file.name.toString().replaceFirst("\\..*","")
-              if (removedFileNames.contains(nameWithoutExt)) {
-                file.delete()
-              }
+                def nameWithoutExt = file.name.toString().replaceFirst("\\..*","")
+                if (removedFileNames.contains(nameWithoutExt)) {
+                    file.delete()
+                }
             }
         }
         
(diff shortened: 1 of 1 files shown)


confile:
@brunobowden May be you have an idea how to add the .class files correctly to the classpath here https://github.com/confile/j2objc-gradle/commit/af138abd9a986c2c27b895d6193dfed0cbf79653#diff-cd56e928e1096973b671c75e5e0c6073R525

When I uncomment the code I get these output messages: 
```
warning: error in opening zip file
```

j2objc.gradle
```diff
@@ -417,8 +418,8 @@ class J2objcCycleFinderTask extends DefaultTask {
         }
     }
 }
-
-
+     
+
```
brunobowden: Avoid adding spaces. Same for previous line as well.

j2objc.gradle
```diff
@@ -426,7 +427,48 @@ class J2objcTranslateTask extends DefaultTask {
     @OutputDirectory File destDir
 
     @TaskAction
-    def translate() {
+    def translate(IncrementalTaskInputs inputs) {
+        println "Source files: "+srcFiles.getFiles().size()
+        FileCollection srcFilesChanged = project.files()
+        inputs.outOfDate { change ->
+        //  println "New or Updated file: "+change.file
+          srcFilesChanged += project.files(change.file)
```
brunobowden: Should be 4 space tabs. Same issues repeats several times in the task.

j2objc.gradle
```diff
@@ -434,12 +476,14 @@ class J2objcTranslateTask extends DefaultTask {
...
         // Clean build directory to delete files that are no longer generated
         // Print command before operation to be helpful log if operation fails
-        println "Deleting j2objc build dir: " + destDir.path
-        project.delete destDir.path
-
+        if (freshBuild) {
+          println "Deleting j2objc build dir: " + destDir.path
+          project.delete destDir.path
+        }
+
```
brunobowden: Remove "freshBuild" logic, this shouldn't be needed and it's better to have a single code path. If you have a rarely used path like this, it invites bugs that will only be rarely seen.

j2objc.gradle
```diff
@@ -449,7 +493,7 @@ class J2objcTranslateTask extends DefaultTask {
         }
 
         // Generated Files
-        srcFiles = J2objcUtils.addJavaFiles(project, srcFiles, project.j2objcConfig.generatedSourceDir)
+       // srcFiles = J2objcUtils.addJavaFiles(project, srcFiles, project.j2objcConfig.generatedSourceDir)
```
brunobowden: Indent one more space

brunobowden: Actually is this just a temporary comment that should be removed entirely?

j2objc.gradle
```diff
@@ -426,7 +427,48 @@ class J2objcTranslateTask extends DefaultTask {
...
+        println "Unchanged files: "+translatedSrcFiles.getFiles().size()
+       
+        
+        if (destDir.exists()) {
+            FileCollection destFiles = project.files(project.fileTree(
+                        dir: destDir, includes: ["**/*.h", "**/*.m"]))
+            
+            // remove translated .h and .m files which has no corresponding .java files anymore
+            destFiles.each {File file ->
+              def nameWithoutExt = file.name.toString().replaceFirst("\\..*","")
```
brunobowden: I would prefer that the matching is done on full filenames, rather than comparing the stripped version. Leave this code as is and I'll rework it myself. The code needs to be refactored so that it takes `J2objcUtils.prefixProperties(project)` in to account which mangles the filename.

j2objc.gradle
```diff
@@ -426,7 +427,48 @@ class J2objcTranslateTask extends DefaultTask {
...
+              if (removedFileNames.contains(nameWithoutExt)) {
+                file.delete()
+              }
+            }
+        }
+        
+               
+        // set the srcFiles to the files which need to be translated
+        srcFiles = srcFilesChanged
```
brunobowden: Can we eliminate srcFiles and only use srcFilesChanges?

j2objc.gradle
```diff
@@ -468,13 +512,45 @@ class J2objcTranslateTask extends DefaultTask {
...
         if (project.j2objcConfig.filenameCollisionCheck) {
             J2objcUtils.filenameCollisionCheck(srcFiles)
         }
      
         def classPathArg = J2objcUtils.getClassPathArg(
                 project, project.j2objcConfig.translateClassPaths)
-
+          
+        
+        // TODO add translated .class files to classpath for incremental translation
```
brunobowden: Is this temporary test code that should be removed?

brunobowden:
@confile - can you help explain this code and what you're trying to do?

```
fileTree(dir: "build/source/apt",
     // TODO @Bruno must be project.j2objcConfig.generatedSourceDir no idea how to set it there
     include: "**/*.java")
```

j2objc.gradle
confile: done!

j2objc.gradle
confile: Ok. I leave this up to you.

j2objc.gradle
confile: No because ``@InputFiles FileCollection srcFiles`` is the input to the tasks containing all source files for this task. The files which have changed are computed after the task has started.

j2objc.gradle
confile: done

j2objc.gradle
confile: done

j2objc.gradle
confile: no this is code your wrote!

confile:
@brunobowden I am done!

confile:
@brunobowden 

```
fileTree(dir: "build/source/apt",
     // TODO @Bruno must be project.j2objcConfig.generatedSourceDir no idea how to set it there
     include: "**/*.java")
```
This code takes the generated source files from an annotation processor which ran in the ``compile-java`` task and appends them to the source files which are the input of the translation task. I have to do it at this point in order to enable the task to track changes.

brunobowden:
I've had the same problem before with trying to setup the correct state from the j2objcConfig object. See the comment from the StackOverflow question I asked. I'm not sure if SO "bigguy" is the same as GitHub @bigguy but he's been very helpful. Still not clear how to wire up the global config but we'll get there...

_... In general, I'd try to stay away from an extension class that contains configuration for a task. There's nothing wrong with that, but you have to do more wiring yourself._ –  bigguy Nov 23 '14 at 19:44

http://stackoverflow.com/questions/27075123/gradle-plugin-best-practices-for-tasks-that-depend-on-extension-objects

j2objc.gradle
```diff
@@ -803,7 +850,10 @@ class j2objc implements Plugin<Project> {
                 srcFiles = files(
                         fileTree(dir: projectDir,
                                 include: "**/*.java",
-                                exclude: project.relativePath(buildDir)))
+                                exclude: project.relativePath(buildDir)) +
+                        fileTree(dir: "build/source/apt",     // TODO @Bruno must be project.j2objcConfig.generatedSourceDir no idea how to set it there
```
brunobowden: Put TODO on separate line so it formats better

j2objc.gradle
```diff
@@ -803,7 +850,10 @@ class j2objc implements Plugin<Project> {
...
+                            include: "**/*.java")
```
brunobowden: Should be indented 8 spaces as function parameter on separate line

j2objc.gradle
```diff
@@ -829,7 +879,7 @@ class j2objc implements Plugin<Project> {
             }
             
             project.tasks.create(name: 'j2objcXcode', type: J2objcXcodeTask,
-                dependsOn: 'j2objcCopy') {
+                dependsOn: 'j2objcTranslate') {
```
brunobowden: Why this change? It should be as it was before.... though I'm thinking of merging the Copy and Xcode tasks together.

j2objc.gradle
```diff
@@ -732,16 +769,26 @@ class J2objcXcodeTask extends DefaultTask {
...
+//                executable "bundle"
+//
+//                args "exec"
+//                args "integratej2objc"
+//                args "integrate_source"
+//                args "-p", "${xcodeProjectDir}"
+//                args "-x", "${xcodeProject}"
+//                args "-s", "${srcFolder}"
+//                args "-g", "${xcodeJ2objcGeneratedDir}"
+//                args "-t", "${xcodeTarget}"
```
brunobowden: You should only need to comment out a couple of lines. We'll have to figure out how to make these installations consistent. Let's do that in a separate thread.

brunobowden:
@confile - just a few last nits

j2objc.gradle
confile: Sorry this was a local change to make the process faster should not be pushed I change it back

confile:
@brunobowden I am done!

j2objc.gradle
brunobowden: Watch out for the trailing spaces. I'll merge this for now and correct it afterwards.

j2objc.gradle
```diff
@@ -741,7 +778,7 @@ class J2objcXcodeTask extends DefaultTask {
                 args "-x", "${xcodeProject}"
                 args "-s", "${srcFolder}"
                 args "-g", "${xcodeJ2objcGeneratedDir}"
-                args "-t", "${xcodeTarget}"
+                args "-t", "${xcodeTarget}"
```
brunobowden: Watch out for the trailing spaces. I'll merge this for now and correct it afterwards.

confile:
@brunobowden Thanks Bruno. Question: Do you have any idea how we could add the already translated classes which have not changed to the ``-classpath`` of the ``j2objc`` command. I tried it here: 
https://github.com/confile/j2objc-gradle/commit/af138abd9a986c2c27b895d6193dfed0cbf79653#diff-cd56e928e1096973b671c75e5e0c6073R525

But when running this code I get the error: ``warning: error in opening zip file``.

brunobowden:
I might be misunderstanding what you're looking but j2objc doesn't yet
translate class files.

PR #40, opened 2015-04-24
Make incremental compilation working

confile:
// add java classpath base to classpath for incremental translation
        // we have to check if translation has been done before or not
        // if it is only an incremental build we must remove the --build-closure
        // argument to make fewer translations of dependent classes 
        // NOTE: TODO there is one case which fails, when you have translated the code
        // make an incremental change which refers to a not yet translated class from a 
        // source lib. In this case due to not using --build-closure the dependent source 
        // we not be translated, this can be fixed with a clean and fresh build

### DIFF
--- a/j2objc.gradle
+++ b/j2objc.gradle
@@ -521,11 +521,7 @@ class J2objcTranslateTask extends DefaultTask {
         def classPathArg = J2objcUtils.getClassPathArg(
                 project, project.j2objcConfig.translateClassPaths)
         
-        def javaClassPath = "${project.buildDir}/classes" 
-        if (classPathArg.size() > 0) {
-            javaClassPath += ":"
-        }
-        classPathArg = javaClassPath + classPathArg  
+        classPathArg += ":${project.buildDir}/classes"  
         
         // add java classpath base to classpath for incremental translation
         // we have to check if translation has been done before or not
--- a/j2objc.gradle
+++ b/j2objc.gradle
@@ -457,7 +457,7 @@ class J2objcTranslateTask extends DefaultTask {
         FileCollection translatedSrcFiles = srcFiles - srcFilesChanged
         println "Unchanged files: "+translatedSrcFiles.getFiles().size()
        
-        
+        def translatedFiles = 0
         if (destDir.exists()) {
             FileCollection destFiles = project.files(project.fileTree(
                         dir: destDir, includes: ["**/*.h", "**/*.m"]))
@@ -469,6 +469,8 @@ class J2objcTranslateTask extends DefaultTask {
                     file.delete()
                 }
             }
+            // compute the number of translated files
+            translatedFiles = destFiles.getFiles().size()
         }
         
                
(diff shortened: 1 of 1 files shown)


brunobowden:
If the build fails, should it do an automatic clean and build again?

j2objc.gradle
```diff
@@ -518,10 +520,25 @@ class J2objcTranslateTask extends DefaultTask {
      
         def classPathArg = J2objcUtils.getClassPathArg(
                 project, project.j2objcConfig.translateClassPaths)
-          
         
-        // TODO add translated .class files to classpath for incremental translation
-   
+        def javaClassPath = "${project.buildDir}/classes" 
+        if (classPathArg.size() > 0) {
+          javaClassPath += ":"
```
brunobowden: 4 space indentation

confile:
The build will not fail. It just don't translates the dependent classes from the source jars. I do not think we can track that.

j2objc.gradle
```diff
@@ -518,10 +520,25 @@ class J2objcTranslateTask extends DefaultTask {
...
+        // we have to check if translation has been done before or not
+        // if it is only an incremental build we must remove the --build-closure
+        // argument to make fewer translations of dependent classes 
+        // NOTE: TODO there is one case which fails, when you have translated the code
+        // make an incremental change which refers to a not yet translated class from a 
+        // source lib. In this case due to not using --build-closure the dependent source 
+        // we not be translated, this can be fixed with a clean and fresh build
+        def j2obcArgs = "${project.j2objcConfig.translateFlags}"
+        if (translatedFiles > 0) {
+          j2obcArgs = j2obcArgs.toString().replaceFirst("--build-closure","").trim()
```
brunobowden: 4 space indentation

j2objc.gradle
```diff
@@ -518,10 +520,25 @@ class J2objcTranslateTask extends DefaultTask {
...
+        
+        // add java classpath base to classpath for incremental translation
+        // we have to check if translation has been done before or not
+        // if it is only an incremental build we must remove the --build-closure
+        // argument to make fewer translations of dependent classes 
+        // NOTE: TODO there is one case which fails, when you have translated the code
+        // make an incremental change which refers to a not yet translated class from a 
+        // source lib. In this case due to not using --build-closure the dependent source 
+        // we not be translated, this can be fixed with a clean and fresh build
+        def j2obcArgs = "${project.j2objcConfig.translateFlags}"
```
brunobowden: j2obcArgs => translateFlags

It's helpful to reuse the same variable name, to make it easier for new engineers to follow the code. When a single search matches all the relevant places, then it's easier to figure out what's going on.

confile:
@brunobowden Done

j2objc.gradle
```diff
@@ -518,10 +520,25 @@ class J2objcTranslateTask extends DefaultTask {
...
+        if (classPathArg.size() > 0) {
+          javaClassPath += ":"
+        }
+        classPathArg = javaClassPath + classPathArg  
+        
+        // add java classpath base to classpath for incremental translation
+        // we have to check if translation has been done before or not
+        // if it is only an incremental build we must remove the --build-closure
+        // argument to make fewer translations of dependent classes 
+        // NOTE: TODO there is one case which fails, when you have translated the code
```
brunobowden: What is the error in this case? If it fails with a distinctive signature, you should identify that during the catch(e) on line 562 onwards. Then give someone the suggestion of forcing a fresh non-incremental build:

```
Incremental build likely failed due to adding new dependency that hasn't
been translated. Try fixing this by cleaning the build and re-running the command:

gradlew base:clean
```

Bonus points to force non-incremental build of just that specific task by perhaps a command line parameter: http://stackoverflow.com/a/19256949/1509221

confile: It will not fail. It will just not translate all the dependent classes from linked source jars. java files from your project will be translated. You cannot track this case. You get no error no warning. You just get fewer classes being translated that's all.

j2objc.gradle
```diff
@@ -518,10 +520,25 @@ class J2objcTranslateTask extends DefaultTask {
...
                 project, project.j2objcConfig.translateClassPaths)
-          
         
-        // TODO add translated .class files to classpath for incremental translation
-   
+        def javaClassPath = "${project.buildDir}/classes" 
+        if (classPathArg.size() > 0) {
+          javaClassPath += ":"
+        }
+        classPathArg = javaClassPath + classPathArg
```
brunobowden: Does it work to leave a trailing ":" on the classPathArg? If so, then simplify to:

    classPathArg = "${project.buildDir}/classes:" + classPathArg

j2objc.gradle
brunobowden: Does it potentially fail when it comes to the j2objcc compiler and linking?

confile: I never use j2objcc. I alway compile in xcode.

confile: You have to execute the link task when you think that some new files have been created but this is always the case.

brunobowden:
I always do an j2objcc compile and then run all the unit tests. It keeps my code in good working order.

brunobowden:
@confile - this is ready to merge now?

confile:
@brunobowden yes you can merge

confile:
Thanks

brunobowden:
I tested it and it works

PR #48, opened 2015-05-05
Support multiple source set directories 

advayDev1:
Support multiple source set directories by using appropriate project properties instead of hard-coding them.

This is useful when you have multiple source-sets, such as:

sourceSets {
  main {
    java.srcDirs += ['otherJavaRoot/']
  }
}

### DIFF
--- a/j2objc.gradle
+++ b/j2objc.gradle
@@ -181,6 +181,17 @@ class J2objcUtils {
         return System.getProperty("os.name").toLowerCase().contains("windows")
     }
 
+    static def sourcepathJava(Project proj) {
+        def javaRoots = []
+        proj.sourceSets['main'].java.srcDirs.each {
+            javaRoots += it.path
+        }
+        proj.sourceSets['test'].java.srcDirs.each {
+            javaRoots += it.path
+        }
+        return javaRoots.join(':')
+    }
+
     static def j2objcHome(Project proj) {
         def result = proj.j2objcConfig.j2objcHome
         if (result == null) {
@@ -360,7 +371,7 @@ class J2objcCycleFinderTask extends DefaultTask {
             throw new InvalidUserDataException(message)
         }
 
-        def sourcepath = "${project.file("src/main/java").path}:${project.file("src/test/java").path}"
+        def sourcepath = J2objcUtils.sourcepathJava(project)
 
         // Generated Files
         srcFiles = J2objcUtils.addJavaFiles(
@@ -492,7 +503,7 @@ class J2objcTranslateTask extends DefaultTask {
             windowsOnlyArgs = "-jar ${j2objcHome}/lib/j2objc.jar"
         } 
        
-        def sourcepath = "${project.file("src/main/java").path}:${project.file("src/test/java").path}"
+        def sourcepath = J2objcUtils.sourcepathJava(project)
 
         // Additional Sourcepaths, e.g. source jars
         if (project.j2objcConfig.translateSourcepaths) {


j2objc.gradle
```diff
@@ -492,7 +492,14 @@ class J2objcTranslateTask extends DefaultTask {
...
        
-        def sourcepath = "${project.file("src/main/java").path}:${project.file("src/test/java").path}"
+        def javaRoots = []
+        project.sourceSets['main'].java.srcDirs.each {
+            javaRoots += it.path
+        }
+        project.sourceSets['test'].java.srcDirs.each {
+            javaRoots += it.path
+        }
+        def sourcepath = javaRoots.join(':')
```
brunobowden: This code looks good... but if you search for "src/main/java", you'll notice that the change also needs to be shared with the CycleFinderTask - line 363.

Best thing is to extract the new code to a static method on the J2objcUtils class. It'll accept "project" as a parameter and return javaRoots.join(':'). The method with then be used in both CycleFinderTask and TranslateTask. Name the method as follows:

    static def sourcepathJava(Project proj) {...

advayDev1:
Thanks - commit amended to d7d5081552e13c563633ddd1b3d721ef5fa0c4cd

Also for my own reference, is it intentional that you use return to return values from methods, and not just leaving the value as the last statement in the method?  Want to stay consistent with your coding style.

brunobowden:
I didn't realize that groovy could do returns without a return statement. I'll definitely keep the style of adding "return" as I think it helps clarity for developers not as familiar with groovy. That also seems to be the consensus of StackOverflow:

http://programmers.stackexchange.com/questions/152405/when-to-write-an-explicit-return-statement-in-groovy

maxbritto:
Hello, how should we update projects that dont use sourceSets to be able to translate again. Since I've updated to latest version I have this error :
```
Error:FAILURE: Build failed with an exception.

* Where:
Script 'j2objc.gradle' line: 216

* What went wrong:
Execution failed for task ':rainbowapi:j2objcTranslate'.
> SourceSet with name 'main' not found.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.
```

Thanks

maxbritto:
I managed to make the build pass by adding this : 
```gradle
sourceSets {
    main {
        java.srcDirs = java.srcDirs
    }
    test {

    }
}
```
in the build.gradle, above the *dependencies* section.

advayDev1:
I'll make a fix for this later today.  BTW how is it that you don't use
source sets?  Which plugins are you using if any?

maxbritto:
I don't use any plugin, when I migrated my project from ant to gradle I created a blank Android studio project to see where it would put the sources and I followed those conventions. I also used the generated build.gradle file from the blank project and it doesn't contains any sourceSets entry. My project build without any issue without this.
BTW, the fix I mentioned above doesn't seem to work : no error is displayed but nothing is translated anymore.

advayDev1:
It seems we assume Java plugin convention in various parts of j2objc-gradle.  
Let me send a PR for flags for the source directories, which when not overridden will default to java plugin conventions (which use sourcesets).

Do you have a link to a build.gradle file you are willing to let us use?  It might be helpful when in creating a diverse testset some day.

advayDev1:
@maxbritto , filed issue #75 to track the fix.

maxbritto:
Thanks I'll track it there. I was wrong before, it is actually working with creating the empty sourceSets like this : 
```gradle
sourceSets {
    main {
    }
    test {
    }
}
```
I'm still exploring/discovering J2OBJC so I haven't added any test yet but for the sources they are now translated like they were before this change. Thanks for the quick answer.

PR #50, opened 2015-05-09
Fix matcher for prefix(es) arguments; fixes #49

### LINKED ISSUE #49: Prefix files with hyphens in path crash build
Because the matcher pattern on L245 identifies argument values by 'inverse-hyphen' class, paths are interpreted as terminating early.  Instead, argument values should continue until whitespace is found.

Example failure when there is a hyphen:
```

* Where:
Script '/Users/XX/clean/YY-ZZ/third_party/j2objc-gradle/j2objc.gradle' line: 245

* What went wrong:
Execution failed for task ':ZZ:j2objcTest'.
> java.io.FileNotFoundException: /Users/XX/clean/YY (No such file or directory)
```

PR incoming.

advayDev1:
Currently, prefixes files with hyphens in their path will crash the build.  Fix #49

### DIFF
--- a/j2objc.gradle
+++ b/j2objc.gradle
@@ -232,7 +232,7 @@ class J2objcUtils {
     //   --prefixes dir/prefixes.properties --prefix com.ex.dir=Short --prefix com.ex.dir2=Short2
     static def prefixProperties(Project proj) {
         Properties props = new Properties()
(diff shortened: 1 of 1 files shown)


brunobowden:
Nice fix @advayDev1 - I'm sure the script will need a lot more fixes like, so thanks for helping out.

PR #53, opened 2015-05-10
Use proper logger levels instead of println.

advayDev1:
This should prevent debug spew on standard build configs.

Errors are used for cycle finder or translation issues.

A warning is used for the last message about lack of tasks to hook up j2objc to,
while debug level is used for all else.  Add --debug to your gradle or gradlew command
to see this messages.

Translation and test output is routed to debug, unless an error occurs, in which
case it is routed to error.

TESTED=yes

### DIFF
--- a/j2objc.gradle
+++ b/j2objc.gradle
@@ -250,7 +250,7 @@ class J2objcUtils {
             props.putAll(newProps)
         }
 //        for (key in props.keys()) {
-//            println key + ": " + props.getProperty(key)
+//            logger.debug key + ": " + props.getProperty(key)
 //        }
 
         return props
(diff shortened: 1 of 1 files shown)


j2objc.gradle
```diff
@@ -596,12 +600,15 @@ class J2objcTranslateTask extends DefaultTask {
...
+                    logger.warn message
                 }
             }
-
+            logger.error 'Error during translation:'
+            logger.error output.toString()
             throw e
         }
+        logger.debug 'Translation output:'
+        logger.debug output.toString()
```
brunobowden: Nice job on this, you've fixed up the logic well. Please check if there's any other tasks that could do with the same error / debug changes.

PR #56, opened 2015-05-11
Add destDirTest option to save away test objc files.

advayDev1:
(Please first do PR #54 and #53 for easy merging)

You can make this a different directory or the same as destDir - it
will work either way.

This allows a convenient way to setup objc source sets
just like your java ones (think:
src-gen/main/objc
src-gen/test/objc
etc.) for easy gradle compilation.

TESTED=yes

### DIFF
--- a/j2objc.gradle
+++ b/j2objc.gradle
@@ -84,6 +84,11 @@ class J2objcPluginExtension {
     // Where to copy generated files (excludes test code and executable)
     String destDir = null
 
+    // Where to copy generated test files (excludes executable)
+    // If null, generated test files are discarded for final output.
+    // Can be the same directory as destDir.
+    String destDirTest = null
+
     // Only generated source files, e.g. from dagger annotations. The script will
     // ignore changes in this directory so they must be limited to generated files.
     String[] generatedSourceDirs = []  
@@ -280,7 +285,7 @@ class J2objcUtils {
     static def addJavaFiles(Project proj, FileCollection files, String[] generatedSourceDirs) {
         if (generatedSourceDirs.size() > 0) {
             generatedSourceDirs.each { sourceDir ->
-                logger.debug "include generatedSourceDir: "+sourceDir
+                logger.debug "include generatedSourceDir: " + sourceDir
                 def buildSrcFiles = proj.files(proj.fileTree(dir: sourceDir, includes: ["**/*.java"]))
                 files += buildSrcFiles
             }
@@ -431,7 +436,7 @@ class J2objcCycleFinderTask extends DefaultTask {
                 def cyclesFound = cycleCountStr.toInteger()
                 if (cyclesFound != project.j2objcConfig.cycleFinderExpectedCycles) {
                     logger.error out
-                    logger.error ("Cycles found (" + cyclesFound + ") != cycleFinderExpectedCycles (" +
+                    logger.error("Cycles found (" + cyclesFound + ") != cycleFinderExpectedCycles (" +
                             project.j2objcConfig.cycleFinderExpectedCycles + ")    ")
                     throw e
                 }
@@ -456,23 +461,23 @@ class J2objcTranslateTask extends DefaultTask {
 
     @TaskAction
     def translate(IncrementalTaskInputs inputs) {
-        logger.debug "Source files: "+srcFiles.getFiles().size()
+        logger.debug "Source files: " + srcFiles.getFiles().size()
         FileCollection srcFilesChanged = project.files()
         inputs.outOfDate { change ->
-            logger.debug "New or Updated file: "+change.file
+            logger.debug "New or Updated file: " + change.file
             srcFilesChanged += project.files(change.file)
         }
         def removedFileNames = []
         inputs.removed { change ->
-            logger.debug "Removed file: "+change.file.name
+            logger.debug "Removed file: " + change.file.name
             def nameWithoutExt = file.name.toString().replaceFirst("\\..*","")
             removedFileNames += nameWithoutExt
         }
-        logger.debug "Removed files: "+removedFileNames.size()
-        
-        logger.debug "New or Updated files: "+srcFilesChanged.getFiles().size()
+        logger.debug "Removed files: " + removedFileNames.size()
+
+        logger.debug "New or Updated files: " + srcFilesChanged.getFiles().size()
         FileCollection translatedSrcFiles = srcFiles - srcFilesChanged
-        logger.debug "Unchanged files: "+translatedSrcFiles.getFiles().size()
+        logger.debug "Unchanged files: " + translatedSrcFiles.getFiles().size()
        
         def translatedFiles = 0
         if (destDir.exists()) {
@@ -562,7 +567,7 @@ class J2objcTranslateTask extends DefaultTask {
                 args windowsOnlyArgs.split()
                 args "-d", "${destDir}"
                 args "-sourcepath", sourcepath
-                
+
                 if (classPathArg.size() > 0) {
                     args "-classpath", classPathArg
                 }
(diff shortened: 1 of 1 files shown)


j2objc.gradle
```diff
@@ -751,33 +806,39 @@ class J2objcCopyTask extends DefaultTask {
...
             exclude "**/*Test.h"
             exclude "**/*Test.m"
         }
+
+        if (project.j2objcConfig.destTestDir != null) {
+            def destTestDir = project.file(project.j2objcConfig.destTestDir)
+            if (destTestDir != destDir) {
+                // If we want main source and test source in one directory, then don't
+                // re-delete the main directory where we just put files!
+                prepDestDir destTestDir, 'destTestDir'
```
brunobowden: I find it easier to understand when it's function style rather than groovy. This is helpful for other engineers making changes:

    prepDestDir(destTestDir, 'destTestDir')

j2objc.gradle
```diff
@@ -751,33 +806,39 @@ class J2objcCopyTask extends DefaultTask {
...
-                    "files that are nolonger generated. Please check the directory and remove\n" +
-                    "any files that don't end with Objective-C extensions '.h' and '.m'.\n" +
-                    "destDir: ${project.j2objcConfig.destDir}\n" +
-                    "Unexpected file for deletion: ${file.path}"
-            throw new InvalidUserDataException(message)
-        }
-        // TODO: better if this was a sync operation as it does deletes automatically
-        println "Deleting destDir to fill with generated objc files... " + destDir.path
-        project.delete destDir
+        prepDestDir destDir, 'destDir'
```
brunobowden: See comment below...

    prepDestDir(destTestDir, 'destTestDir')

advayDev1:
@brunobowden - currently j2objc.gradle has a mix of groovy top-level method calls and java ones.  for example, near the lines you commented on there was 'project.delete destDir'.  also your readme uses groovy style in its example of how to configure the plugin.  do you want to actively go towards java style top-level method calls?  thanks!

j2objc.gradle
```diff
@@ -739,6 +775,25 @@ class J2objcCopyTask extends DefaultTask {
     File srcDir
     // TODO: declare @OutputXXX so gradle knows if task is up to date
 
+    private def prepDestDir(File destDir, String name) {
```
brunobowden: Rename to `clearDestDirWithChecks` as that makes it purpose clearer.

j2objc.gradle
```diff
@@ -739,6 +775,25 @@ class J2objcCopyTask extends DefaultTask {
...
+        def destFiles = project.files(project.fileTree(
+                dir: destDir, excludes: ["**/*.h", "**/*.m"]))
```
brunobowden: Does this cause a problem if the test output is placed in the same folder?

j2objc.gradle
```diff
@@ -132,6 +137,30 @@ class J2objcPluginExtension {
...
+    // Excludes a fully qualified class name from translation.
+    // Example: translateExcludeClass 'java.lang.String'
+    def translateExcludeClass(def className) {
+        translateExcludeRegex = regexOr(translateExcludeRegex, classNameToFileRegex(className))
+    }
+    // Includes a fully qualified class name in translation.
+    // Example: translateIncludeClass 'java.lang.String'
+    def translateIncludeClass(def className) {
+        translateIncludeRegex = regexOr(translateIncludeRegex, classNameToFileRegex(className))
+    }
```
brunobowden: I can't see any place that this code is used. It would be best to propose this change alongside some code (and an example) that uses it.

j2objc.gradle
advayDev1: @brunobowden no problem, see L826.  also i've tested both destTestDir == destDir and destTestDir != destDir.

however, if you're doing something more wacky like destTestDir is a sub dir of destDir and there are some classes that happen to have a package prefix equaling the difference in path between the two, or if destDir is a sub dir of  destTestDir, then in the first class you may be confused by the paths that result, and in the second case the original non-test files will be deleted.  worth warnings to catch these edge cases?

brunobowden:
Hi @advayDev1 - I think the Java style function calls is better as that makes it clearer for other developers proposing changes. I realize that there's multiple places where that needs to be fixed up.

Please rebase / merge this with master and I'll take another look. Also, is the include / exclude classes still needed?

advayDev1:
@brunobowden - the include/exclude classes is from PR #54 (that commit is in both this branch and that one).  i do use that code in my project as it is more convenient than making regexes for each class I exclude, and an example is provided on the declaration.

also, I will repair the java-style calls affected by this change.  thanks for the feedback!

j2objc.gradle
advayDev1: see PR #54

j2objc.gradle
advayDev1: fixed PTAL

j2objc.gradle
advayDev1: fixed PTAL

j2objc.gradle
advayDev1: fixed PTAL

advayDev1:
rebased a new version for all comments except PR #54 and the open comment re: https://github.com/brunobowden/j2objc-gradle/pull/56#discussion_r30204264 - will modify further if you make a decision on the other PR and want to see the latter added complexity.

j2objc.gradle
```diff
@@ -84,6 +84,11 @@ class J2objcPluginExtension {
     // Where to copy generated files (excludes test code and executable)
     String destDir = null
 
+    // Where to copy generated test files (excludes executable)
+    // If null, generated test files are discarded for final output.
+    // Can be the same directory as destDir.
+    String destTestDir = null
```
brunobowden: Rename to 'destDirTest' as that signifies closer association with 'destDir'

j2objc.gradle
```diff
@@ -132,6 +137,30 @@ class J2objcPluginExtension {
     //     translateIncludeRegex ".*/TranslateOnlyMeAnd(|Test)\\.java"
     String translateExcludeRegex = null
     String translateIncludeRegex = null
+
+    private def classNameToFileRegex(def className) {
```
brunobowden: Add TODO explaining planned changes to switch to glob / pattern match. It would be best to link to an open issue which writes this up in more detail.

j2objc.gradle
advayDev1: done

j2objc.gradle
advayDev1: i've rebased this branch to exclude #54, since this other comment thread indicates we might want to go a different way with that  functionality.

brunobowden:
Just add the TODO and then I'll merge. I'm working on some changes with confile early tomorrow morning, so I'd like to get this in first.

advayDev1:
Per https://github.com/brunobowden/j2objc-gradle/pull/56#discussion_r30206228, TODO is no longer needed (cleaner this way anyway).

brunobowden:
I merged in #54, thinking that you needed that for now ;-) Never mind, we'll clean this up later.

advayDev1:
Yeah actually there's no hurry in PRs on this repo I guess.  I have various alterations to this plugin in my proprietary repo, and am extricating my changes (to skip private or thoroughly entangled changes) into my public fork and eventually your source repo.

brunobowden:
In the future, I think of this as not just a tool for shared build but also
for partial migration. Some of which you may be already dealing with.

advayDev1:
Yep, I see what you mean.  I see that @confile is working on making this easier to use with Xcode projects with integratej2objc.   My workflow is a bit different in that I'm trying to use the xcode toolchain via Gradle[1] and with no prebuilts or checked-in generated source, as I do not want to deal with Xcode's workspace management for many of the reasons that make integratej2objc necessary or with denormalized iOS source (I have a single codebase that I want to run './gradlew build' from the root and have it build all my custom Groovy toolchain, my common Java libraries, my Android apk, generated objc, and my iOS app all at once in that order.)  Also working on getting this plugin to work with ObjC++.  I'll file issues so we can talk over things in an organized manner.

Also, wishing github PRs had batched/atomic code review comments/revisions right now =)

[1] https://gradle.org/docs/current/userguide/nativeBinaries.html

Issue #60, opened 2015-05-13
Better error logging


advayDev1:
Right now if translation or something else fails, you get the whole process output dumped and have to use scrollback to find your mistake.  I will add to the end of the output a copy of any lines that begin with 'error: '

PR #62, opened 2015-05-13
When translation fails, display only the error messages.

advayDev1:
Full output remains available with --debug.

Issue #60

### DIFF
--- a/j2objc.gradle
+++ b/j2objc.gradle
@@ -345,6 +345,10 @@ class J2objcUtils {
         }
         return classPathList.join(':')
     }
+
+    static def filterJ2objcOutputForErrorLines(String processOutput) {
+        return processOutput.tokenize('\n').grep(~/^(.*: )?error:.*/).join('\n')
+    }
 }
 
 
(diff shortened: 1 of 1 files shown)


j2objc.gradle
```diff
@@ -345,6 +345,10 @@ class J2objcUtils {
         }
         return classPathList.join(':')
     }
+
+    static def filterForErrorLines(String processOutput) {
+        return processOutput.tokenize('\n').grep(~/^error:.*/).join('\n')
+    }
```
brunobowden: Is it more complex than this, e.g. that showing a stack trace on an error, only the first line may have the error prefix?

advayDev1: They don't seem to be consistent (https://github.com/google/j2objc/search?p=3&q=ErrorUtil.error&utf8=%E2%9C%93).  Sometimes they include a stack trace in their own output, but usually not.  I assume the former is for internal errors in their compiler.  I think, on the balance, it is better to print a few single lines of error for developer convenience in the normal case (i.e. developer fault, bug in their Java input) than have additional complexity to discover stack traces, especially since full error info is available by passing --debug, and gradle by default gives a message re: adding that flag for all build failures.

Thoughts?

I did go ahead and add another case from https://github.com/google/j2objc/blob/master/translator/src/main/java/com/google/devtools/j2objc/util/ErrorUtil.java#L83 though - I've never seen one of those clang style errors myself but might as well be compatible.

brunobowden: I've merged it in. Let's see how it goes and then change it later on if
needed.

What I'd love was a user opt-in for sharing debug error messages, so you
could find out where the pain points were. Unless you gather stats like
that, you have no idea what's going on.

advayDev1: Great idea.  IMO this should be a j2objc thing though.  No matter where you use j2objc, you want to see issues like that, not just with the gradle plugin.

PR #46, opened 2015-05-03
J2objcPodTask for Xcode integration using CocoaPods

confile:
- generate a podspec file
- integrate the pod in the podfile in the project dir
- install the pod

PR #64, opened 2015-05-14
Start CONTRIBUTING.md with how to finish a PR

advayDev1:
Using PR #46 as an example.

@confile, I'd appreciate your comments on anything to clarify. It's been a while since I've done this for the first time so I may have elided steps that are second nature to me.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -59,3 +59,6 @@ git merge master
 # resolve and finish the merge as usual
 git push
 ```
+
+For guidance on doing the merge itself, see
+http://git-scm.com/book/en/v2/Git-Branching-Basic-Branching-and-Merging
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,18 +22,23 @@ We'll assume you are working from your local repository
 which is a clone of a fork of brunobowden/j2objc-gradle, that your feature
 branch is called 'patch-1' and that your pull request is number 46.
 
+### Preperation
 ```
 # have a clean working directory and index
 # from the tip of your patch-1 branch: first save away your work
 git branch backup46
-# now don't checkout or touch branch backup46 from here on out.
-# if things go terribly wrong below, you can then return to a state of LKG sanity with:
-#git rebase --abort
-#git merge --abort
-#git checkout patch-1
-#git reset --hard backup46
-# this will return your patch-1 branch to the state before merging & rebasing
 ```
+
+Don't checkout or touch branch backup46 from here on out.  If things
+go terribly wrong below, you can return to a state of LKG sanity with:
+
+1.  `git rebase --abort`
+2.  `git merge --abort`
+3.  `git checkout patch-1`
+4.  `git reset --hard backup46`
+
+This will return your patch-1 branch to the state before merging & rebasing
+
 First ensure you've setup your upstream remote as per 
 https://help.github.com/articles/configuring-a-remote-for-a-fork/,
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,64 @@
+## Style guide
+TODO
+
+## Testing your code
```
brunobowden: There's a lot we need to rework here

advayDev1: Definitely, I think we should do those as separate commits though - just laying out some TODOs for what we'll need in the future.   Determining a testing strategy for this project will be more involved than a guide on how to merge, while the latter could be immediately helpful.

CONTRIBUTING.md
```diff
@@ -0,0 +1,64 @@
...
+```
+
+Now you can work on merging in master to your branch.  We'll assume a simple
+branch history here, where patch-1 first diverged from master at commit
+@5665f0a5ff774005ce30e088ab2e1b76caca5a43.
+```
+git checkout patch-1
+# condense this down to one commit to preserve proper project history
+git rebase -i 5665f0a5ff774005ce30e088ab2e1b76caca5a43
+# within the rebase editor, replace the word 'pick' with 'fixup' on
```
confile: You should claryfy where you get this number ``5665f0a5ff774005ce30e088ab2e1b76caca5a43`` from. I know it defines the request but write if it is the first or the last. 

Also: 
'pick' and  'fixup' I was totally lost here. What does this step do?

CONTRIBUTING.md
```diff
@@ -0,0 +1,64 @@
...
+best position to resolve these correctly.  Even if the merge is a fast-forward,
+doing the merge yourself allows you to test your code after incorporating others'
+changes.
+
+If you are new to github, here's an example workflow.
+We'll assume you are working from your local repository
+which is a clone of a fork of brunobowden/j2objc-gradle, that your feature
+branch is called 'patch-1' and that your pull request is number 46.
+
+```
```
confile: Say that the user has to checkout his cloned repo and the correct branch with.

```
git checkout --track origin/branch_branch
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,64 @@
...
+# have a clean working directory and index
+# from the tip of your patch-1 branch: first save away your work
+git branch backup46
+# now don't checkout or touch branch backup46 from here on out.
+# if things go terribly wrong below, you can then return to a state of LKG sanity with:
+#git rebase --abort
+#git merge --abort
+#git checkout patch-1
+#git reset --hard backup46
+# this will return your patch-1 branch to the state before merging & rebasing
```
brunobowden: Convert lines 30-35 to a numbered list, more markdown in nature.

CONTRIBUTING.md
advayDev1: good points.
Do you believe https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history is a good explanation?  If so, that is the reference I will give.  (I think best to link to experts here, no need to repeat how git works in the file).

confile: yes you can add this!

advayDev1: ptal

CONTRIBUTING.md
advayDev1: ptal

confile:
May be could could add a link how to configure the default git editor that helped me a lot.

brunobowden:
Merged

@confile - feel free to discuss with @advayDev1 about a PR for your other suggestions

PR #65, opened 2015-05-14
update docs j2objcPod

confile:


brunobowden:
@confile - I'd like to not merge this in. Let's wait until we've got J2objcPodTask in good working order, then rename it to J2objcXcodeTask and have it replace the existing Xcode task. As it's only a temporary name, it shouldn't make it in to the documentation.

confile:
okay

PR #77, opened 2015-05-21
testSkip configuration added #73

confile:
... enables to skip Test

rename the cycleFinder to skipCycleFinder and negate the logic

### DIFF
--- a/j2objc.gradle
+++ b/j2objc.gradle
@@ -183,9 +183,9 @@ class J2objcPluginExtension {
 
 
     // TEST
-    // Flags copied verbatim to testrunner command
     // Skip test task if true
     boolean testSkip = false
+    // Flags copied verbatim to testrunner command
     String testFlags = ""
     // Filter tests, applied in addition to translate filters (see above)
     String testExcludeRegex = null
--- a/j2objc.gradle
+++ b/j2objc.gradle
@@ -185,7 +185,7 @@ class J2objcPluginExtension {
     // TEST
     // Flags copied verbatim to testrunner command
     // Skip test task if true
-    boolean skipTest = false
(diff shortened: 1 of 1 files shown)


j2objc.gradle
```diff
@@ -99,7 +99,7 @@ class J2objcPluginExtension {
 
     // CYCLEFINDER
     // TODO(bruno): consider enabling cycleFinder by default
-    boolean cycleFinder = false
+    boolean skipCycleFinder = true
```
brunobowden: Thinking about it again and I think that `cycleFinderSkip` and `testSkip` would be more consistent with the other flags that already exist. Please change that and then I'll merge it.

j2objc.gradle
```diff
@@ -184,6 +184,8 @@ class J2objcPluginExtension {
 
     // TEST
     // Flags copied verbatim to testrunner command
```
brunobowden: The "Flags copied verbatim..." comment should be next to the testFlags line of code

confile:
done

advayDev1:
LGTM, if indent on L748 is fixed

brunobowden:
@confile - fix up L748 and then I'll merge

confile:
@brunobowden What is wrong with L748?

brunobowden:
See @advayDev1 comment about the indentation being off. Line 748 needs to be indented another space. See image:

screen shot 2015-05-21 at 8 38 52 am

confile:
@advayDev1 @brunobowden  Thanks for comments I could not see the missing indentation in the github online editor.

brunobowden:
It's now fixed. I might have had a stale copy before but it certainly
needed the indentation fix.

Issue #52, opened 2015-05-09
Use idiomatic sourceSets for all listings of source files


advayDev1:
The idiomatic way in Gradle to specify what sources a project is using is to use source sets.  Usually 'main' for production code and 'test' for test code.

Currently, source files are found by aggregating globs of .java files, ex:
```
srcFiles = files(
                        fileTree(dir: projectDir,
                                include: "**/*.java",
                                exclude: project.relativePath(buildDir)) +
                        fileTree(dir: "build/source/apt",
                                include: "**/*.java")
                        )
```

Instead we should always be using the sourceSets in the project.  Currently, we will erroneously find files that are not part of the project build but just happen to be in the directory, miss files the user has added to the source set (specifically generated source directories, which have to be added manually but should not have to be), and causes special casing like the above for apt (this should also allow us to fix #45.).

In addition for test sources:
```
 // Doesn't use 'buildDir' as missing full path with --no-package-directories flag
                srcFiles = files(fileTree(dir: projectDir, includes: ["**/*Test.java"]))
```

we should be using the 'test' source set.

confile:
Please make a pull request if you have any improvements!

advayDev1:
Yep, I'm working on it (generally all the bugs I file I have in progress solutions for, but github prevents me from assigning the bug to myself like most workflows I'm used to).

My own gradle plugins use project afterEvaluation stage to configure the tasks' input directories to be equal to the projects' source sets.  Do y'all have a preference on when to configure the tasks?  It means some non-trivial changes to how the tasks and plugin get setup.

confile:
@advayDev1 Any progress on that issue?

advayDev1:
Now that I've got Bruno's opinion on #75 , I can get this done.  It'll be one pull request.

We'll be supporting sourceSets, a default hard-coded convention, and user overrides it looks like.

brunobowden:
That sounds like a good balance.

advayDev1:
Ok see pull #78  - we can't finish this until the transition to real plugin allows us to configure the inputs to certain tasks using source sets.

PR #78, opened 2015-05-21
    Flexibly configure source directories.

### LINKED ISSUE #75: Gradle projects lacking sourceSets will not compile
We've assumed Java plugin (and its derivatives, like Groovy plugin) usage, which for example use sourceSets.  Where we use Java plugin convention by default, we should provide a way to skip that default by providing flags.

As reported by @maxbritto in #48.

At minimum we should allow overriding of the java, test, and resource source directories.

advayDev1:
In general, a user can specify source directories with one of:
        // 1. customSourceDirectories.  You may override 0 to 4 of the directories
        //    for {main, test} x {java, resources}.  The plugin will use any non-null
        //    list values, and backoff to the remaining options for ones that are null.
        //    Example:
        //    customSourceDirectories = [main: [java: ["src-gen/core"], resources: null],
        //                               test: [java: ["src-gen/test"], resources: null]]
        // 2. Project SourceSets.
        // 3. "Conventional" defaults for each directory, namely:
        //    src/{main|test}/{java|resources}
        //
        // If you use the standard java plugin, this works automatically, and will
        // take into account any customizations you have made of the 'main' and/or
        // 'test' SourceSets, and you do not need to repeat such customization here.

    Fixes #75; improves #52.

Tested options 1 and 2, we can ask @maxbritto to test option 3.

### DIFF
--- a/j2objc.gradle
+++ b/j2objc.gradle
@@ -98,21 +98,24 @@ class J2objcPluginExtension {
 
     // Input source files for this plugin can be configured in a few ways.  The
     // plugin will use the first valid option:
-    // 1. The flags below.  You may override 0 to 4 of the directories
-    //    for {main, test} x {java, resources}.  The plugin will use any non-null
-    //    list values, and backoff to the remaining options for ones that are null.
-    //    Example:
-    //    customSourceDirectories = [main: [java: ["src-gen/core"], resources: null],
-    //                               test: [java: ["src-gen/test"], resources: null]]
-    // 2. Project SourceSets.
+    // 1. The flags below.  You can override 0 to 4 of the flags; any that are null
+    //    will backoff to the following.
+    // 2. Project SourceSets.  For example, if you use the 'java' or related plugins.
     // 3. "Conventional" defaults for each directory, namely:
-    //    src/{main|test}/{java|resources}
+    //    src/{main|test}/{java|resources}.  For example, if you use an automatically
+    //    converted Ant build.
     //
-    // If you use the standard java plugin, this works automatically, and will
-    // take into account any customizations you have made of the 'main' and/or
-    // 'test' SourceSets, and you do not need to repeat such customization here.
-    def customSourceDirectories = [main: [java: null, resources: null],
-                                   test: [java: null, resources: null]]
+    // WARNING:
+    // We recommend using the standard java plugin (or derivatives) instead of
+    // configuring this plugin's source directories manually:
+    //     apply plugin: 'java'
+    // If you use the standard java plugin, any source set customization will be
+    // applied to this plugin as well, and many other plugins will understand
+    // your customizations without duplicating configuration.
+    String[] customMainSrcDirs = null  // defaults to ['src/main/java']
+    String[] customTestSrcDirs = null  // defaults to ['src/test/java']
+    String[] customMainResourcesSrcDirs = null  // defaults to ['src/main/resources']
+    String[] customTestResourcesSrcDirs = null  // defaults to ['src/test/resources']
 
 
     // CYCLEFINDER
@@ -232,15 +235,26 @@ class J2objcUtils {
     }
 
     // Retrieves the configured source directories per specification indicated
-    // in documentation of 'customSourceDirectories'.
+    // in documentation of 'custom.*SrcDirs'.
     private final static def DEFAULT_SOURCE_DIRECTORIES =
             [main: [java: ['src/main/java'], resources: ['src/main/resources']],
              test: [java: ['src/test/java'], resources: ['src/test/resources']]]
     static def srcDirs(Project proj, String sourceSetName, String fileType) {
+        assert fileType == 'java' || fileType == 'resources'
+        assert sourceSetName == 'main' || sourceSetName == 'test'
         // Note that subprojects can also be passed to this method; such projects
         // may not have j2objcConfig.
         if (proj.hasProperty('j2objcConfig')) {
-            def custom = proj.j2objcConfig.customSourceDirectories[sourceSetName][fileType]
+            String custom = null
+            if (fileType == 'java') {
+                custom = sourceSetName == 'test' ?
+                    proj.j2objcConfig.customTestSrcDirs :
+                    proj.j2objcConfig.customMainSrcDirs
+            } else if (fileType == 'resources') {
+                custom = sourceSetName == 'test' ?
+                    proj.j2objcConfig.customTestResourcesSrcDirs :
+                    proj.j2objcConfig.customMainResourcesSrcDirs
+            }
             // We intentionally check for nulls only.  It is valid to desire
             // an empty list of (i.e. no) source directories.
             if (custom != null) {
@@ -257,19 +271,13 @@ class J2objcUtils {
         }
         return DEFAULT_SOURCE_DIRECTORIES[sourceSetName][fileType].collect({ proj.file(it) })
     }
-    static def javaSrcDirs(Project proj, String sourceSetName) {
-        return srcDirs(proj, sourceSetName, 'java')
-    }
-    static def resourcesSrcDirs(Project proj, String sourceSetName) {
-        return srcDirs(proj, sourceSetName, 'resources')
-    }
 
     static def sourcepathJava(Project proj) {
         def javaRoots = []
-        javaSrcDirs(proj, 'main').each {
+        srcDirs(proj, 'main', 'java').each {
             javaRoots += it.path
         }
-        javaSrcDirs(proj, 'test').each {
+        srcDirs(proj, 'test', 'java').each {
             javaRoots += it.path
         }
         return javaRoots.join(':')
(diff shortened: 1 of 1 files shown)


advayDev1:
i tested this on projects of mine that do actually have custom source sets with additional srcDirs using both customSourceDirectories and source set configuration.

j2objc.gradle
```diff
@@ -213,12 +231,45 @@ class J2objcUtils {
         return System.getProperty("os.name").toLowerCase().contains("windows")
     }
 
+    // Retrieves the configured source directories per specification indicated
+    // in documentation of 'customSourceDirectories'.
+    private final static def DEFAULT_SOURCE_DIRECTORIES =
+            [main: [java: ['src/main/java'], resources: ['src/main/resources']],
+             test: [java: ['src/test/java'], resources: ['src/test/resources']]]
+    static def srcDirs(Project proj, String sourceSetName, String fileType) {
```
brunobowden: Reorder parameters to: String fileType, String sourceSetName, Project proj... this is intended to read better when writing out the logic:

    srcDirs('main', 'java', proj)

The idea is that `'main', 'java'` more easily mentally translates to `src/main/java`

j2objc.gradle
```diff
@@ -213,12 +231,45 @@ class J2objcUtils {
...
+            }
+        }
+        return DEFAULT_SOURCE_DIRECTORIES[sourceSetName][fileType].collect({ proj.file(it) })
+    }
+    static def javaSrcDirs(Project proj, String sourceSetName) {
+        return srcDirs(proj, sourceSetName, 'java')
+    }
+    static def resourcesSrcDirs(Project proj, String sourceSetName) {
+        return srcDirs(proj, sourceSetName, 'resources')
+    }
```
brunobowden: I'd eliminate the javaSrcDirs and resourcesSrcDirs as the indirection doesn't save much text. Furthermore, writing out the full function call without an intermediate layer - I think that makes it easier to understand.

j2objc.gradle
```diff
@@ -213,12 +231,45 @@ class J2objcUtils {
...
+
     static def sourcepathJava(Project proj) {
         def javaRoots = []
-        proj.sourceSets['main'].java.srcDirs.each {
+        javaSrcDirs(proj, 'main').each {
```
brunobowden: Replace with:

    srcDirs('main', 'java', proj).each {

Repeat below as well

j2objc.gradle
```diff
@@ -1104,6 +1155,7 @@ class j2objc implements Plugin<Project> {
 
             tasks.create(name: "j2objcCycleFinder", type: J2objcCycleFinderTask) {
                 description "Run the cycle_finder tool on all Java source files"
+                // TODO: Once this is a proper plugin, we can switch this to use SourceSets.
```
brunobowden: Thanks for the whitespace cleanup and TODOs

j2objc.gradle
```diff
@@ -91,11 +91,29 @@ class J2objcPluginExtension {
...
+    //                               test: [java: ["src-gen/test"], resources: null]]
+    // 2. Project SourceSets.
+    // 3. "Conventional" defaults for each directory, namely:
+    //    src/{main|test}/{java|resources}
+    //
+    // If you use the standard java plugin, this works automatically, and will
+    // take into account any customizations you have made of the 'main' and/or
+    // 'test' SourceSets, and you do not need to repeat such customization here.
+    def customSourceDirectories = [main: [java: null, resources: null],
+                                   test: [java: null, resources: null]]
```
brunobowden: I would enumerate the individual srcDirs to make it easier for developers to work with and also get closer to the naming conventions of the java plugin. I can see why you did it the other way though for easier code logic.

    String[] mainSrcDirs = ['src/main/java']
    String[] testSrcDirs = ['src/test/java']
    String[] mainResourcesSrcDirs = ['src/main/resources']
    String[] testResourcesSrcDirs = ['src/test/resources']

screen shot 2015-05-21 at 9 31 20 am

j2objc.gradle
```diff
@@ -213,12 +231,45 @@ class J2objcUtils {
...
     }
 
+    // Retrieves the configured source directories per specification indicated
+    // in documentation of 'customSourceDirectories'.
+    private final static def DEFAULT_SOURCE_DIRECTORIES =
+            [main: [java: ['src/main/java'], resources: ['src/main/resources']],
+             test: [java: ['src/test/java'], resources: ['src/test/resources']]]
+    static def srcDirs(Project proj, String sourceSetName, String fileType) {
+        // Note that subprojects can also be passed to this method; such projects
+        // may not have j2objcConfig.
```
brunobowden: Should j2objcConfig apply in that case? Don't know and don't need to work that out for this PR

j2objc.gradle
```diff
@@ -91,11 +91,29 @@ class J2objcPluginExtension {
...
+    //    Example:
+    //    customSourceDirectories = [main: [java: ["src-gen/core"], resources: null],
+    //                               test: [java: ["src-gen/test"], resources: null]]
+    // 2. Project SourceSets.
+    // 3. "Conventional" defaults for each directory, namely:
+    //    src/{main|test}/{java|resources}
+    //
+    // If you use the standard java plugin, this works automatically, and will
+    // take into account any customizations you have made of the 'main' and/or
+    // 'test' SourceSets, and you do not need to repeat such customization here.
```
brunobowden: Add a serious warning to the code:

    // NOTE: we recommend not using this and instead using automatic setup with:    apply plugin: 'java'

j2objc.gradle
```diff
@@ -91,11 +91,29 @@ class J2objcPluginExtension {
...
     String j2objcHome = null
 
+    // Input source files for this plugin can be configured in a few ways.  The
+    // plugin will use the first valid option:
+    // 1. The flags below.  You may override 0 to 4 of the directories
+    //    for {main, test} x {java, resources}.  The plugin will use any non-null
+    //    list values, and backoff to the remaining options for ones that are null.
+    //    Example:
+    //    customSourceDirectories = [main: [java: ["src-gen/core"], resources: null],
+    //                               test: [java: ["src-gen/test"], resources: null]]
```
brunobowden: With the simpler config logic, I think you can shorten this section

j2objc.gradle
brunobowden: Should it be prefixed with custom so that people understand that it's special?

    String[] customMainSrcDirs = ['src/main/java']
    String[] customTestSrcDirs = ['src/test/java']
    String[] customMainResourcesSrcDirs = ['src/main/resources']
    String[] customTestResourcesSrcDirs = ['src/test/resources']

brunobowden:
@advayDev1 - I've added a bunch of comments. If you're able to get to this, I'll merge it in and handle any fallout. I think it mainly came down to making it easier for developers.

Last thought, should we instead have 'src/main/java' and so on be internal defaults only? Then if someone wants to to do something custom, tell them that they should manually configure the source sets, e.g.:

```
sourceSets {
  main {
    java {
      exclude 'some/unwanted/package/**'
    }
  }
}
```

I think that only works through if you've added the java plugin.

j2objc.gradle
advayDev1: actually it should.  the usecase is --build-closure with subprojects, and we should use their configured j2objc directories when creating the translateSourcePaths for the superproject as well.

advayDev1:
@brunobowden - did everything except the changing of the order of srcDirs params.  Since this function is basically a static helper standing in for a method on Project, I think general convention is to leave the 'explicit this' as the first parameter (ex. Groovy extension methods).

Not completely understanding your point about internal defaults,  you could be talking about 2 things:
- I think you wanted to be able to support common non-java-plugin configs (i.e. the Ant one @maxbritto tried).  In that case, our default must be part of our API.  Which is what I've implemented in the updated PR (which means the flags in the extension must have null default values, to allow fallback to the sourceSets, and eventually fall back to the default conventions.)
- If you are talking about how to improve our j2objc-specific inclusion/exclusion flags in the future, I think creating source sets is one way to go.  However, they would have to be source sets within our plugin extension.  I.e. we would add support for a SourceSetContainer within our extension, and then you can define a main sourceSet in our DSL like:
```
j2objcConfig {
    sourceSets {
      // exclude the JNI stuff, for example.
   }
}
```
We would then default our source sets to the project's overall sources sets, unless configured in our plugin extension.

All I've got time for at this moment, anything else will have to wait.

Thanks!

brunobowden:
Thanks @advayDev1 

@maxbritto - please try this newly merged version, it should fix the problem you were seeing

maxbritto:
It works indeed! Thanks for the quick fix and sorry about yesterday : I was out of the office so I couldn't be of any help

brunobowden:
@advayDev1 is the gentleman to thank here ;-)

maxbritto:
Thanks @advayDev1 :+1:

PR #81, opened 2015-06-01
compileSkip setting analogous to testSkip

advayDev1:


confile:
+1 I like the idea!

advayDev1:
hopefully when we make this a real plugin starting tomorrow, we will use normal task.enabled = false.

this causes some issues, which I'm fixing (one sec...)

brunobowden:
@advayDev1 - thanks for contributing, I only changed the description to make "compileSkip" the first word in the description.

advayDev1:
@brunobowden  - sorry, commit race.  please merge in #82  as well

PR #83, opened 2015-06-02
J2objcPodTask fixes and error message improvements

brunobowden:
- Fixed depProj => project which didn’t run before
- Fixed j2objcHome which didn’t run before

Error Message Improvements:
- Missing podfile
- Missing pod command
- project.version unspecified


brunobowden:
@confile - I still haven't gotten J2objcPodTask to work yet. In particular, see lines 1170-76. I couldn't figure out how to specify project.version. Because it was unspecified, s.version was unspecified:

    "s.version = '${podVersion}'" + lineSeparator +

confile:
Here is the solution: 

```
def podVersion = "${project.version}"
```

brunobowden:
I don't think that should make any difference. I tried the two approaches:

    def podVersionOld = project.version
    def podVersionNew = "${project.version}"
    print "podVersionOld: " + podVersionOld + "\n"
    print "podVersionNew: " + podVersionNew + "\n"

The output was **unspecified** for both:

    podVersionOld: unspecified
    podVersionNew: unspecified

brunobowden:
@confile - I suspect that you're specifying the project.version somewhere else in your build.gradle file. That it's missing from my setup which is why it encounters this error. Did you do any special setup for this?

confile:
Yes you are right. in my build.gradle I did: 

```
version = '1.0'
```
It was in the default template I used.

brunobowden:
Is this a requirement for Podfile? Do we need to educate all users on how
to do this? Do we need a version number?

Basically I'm wondering if we need to add the complexity or if there's a
way around it.

confile:
Not sure but it is important when project version changes. So I think we should de: 
```
def podVersion =  project.version ?: "1.0" 
```

brunobowden:
I'm wary of this as your system might have different behaviour than mine
depending on whether you've created a project.version setting. If we're not
sure if it's needed, much better that it doesn't depend on external
settings. That way, we get exactly the same behaviour. I'm trying to avoid
the situation where I encounter a bug that you never see. If that happens,
then you could easily introduce a bug for others without realizing it.
Better that it has the same code path.

Unless there's a specific reason, I'm just going to set podVersion to "1.0"
and ignore the project version entirely. Any objections to that?

confile:
My point is that every pod has a version. So we should use at least a default of 1.0.

brunobowden:
Are there occasions when you should change a Pod version?

confile:
Yes when you updated the source code of the pod. Typically a pod is only reinstalled if the version has changed.

PR #84, opened 2015-06-02
Avoid incremental builds with --build-closure by default; Fixes #70.

### LINKED ISSUE #70: Fix --build-closure performance exception
I disagree that we should prioritize speed over correctness in the case below (my projects use --build-closure).  What are folks opinions' on me getting rid of this conditional and always keeping --build-closure if the user asked for it?  @confile  @brunobowden 

If we later find a way to make this work in all cases, we should make the optimization, but I don't think we should unless it is guaranteed correct.

```
...
        // add java classpath base to classpath for incremental translation
        // we have to check if translation has been done before or not
        // if it is only an incremental build we must remove the --build-closure
        // argument to make fewer translations of dependent classes 
        // NOTE: TODO there is one case which fails, when you have translated the code
        // make an incremental change which refers to a not yet translated class from a 
        // source lib. In this case due to not using --build-closure the dependent source 
        // we not be translated, this can be fixed with a clean and fresh build
        def translateFlags = "${project.j2objcConfig.translateFlags}"
        if (translatedFiles > 0) {
            translateFlags = translateFlags.toString().replaceFirst("--build-closure","").trim()
        }
...
```

advayDev1:
UNSAFE_incrementalBuildClosure can be used to speed up incremental
builds with --build-closure IF you know what you are doing (for example
very rarely change your projects dependencies)

### DIFF
--- a/j2objc.gradle
+++ b/j2objc.gradle
@@ -135,7 +135,14 @@ class J2objcPluginExtension {
     String translateFlags = "--no-package-directories"
     // -classpath library additions from ${projectDir}/lib/, e.g.: "json-20140107.jar", "somelib.jar"
     String[] translateClassPaths = []
-    
+
+    // WARNING: Do not use this unless you know what you are doing.
+    // If true, incremental builds will be supported even if --build-closure is included in
+    // translateFlags. This may break the build in unexpected ways if you change the dependencies
+    // (e.g. adding new files or changing translateClassPaths). When you change the dependencies and
+    // the build breaks, you need to do a clean build.
+    boolean UNSAFE_incrementalBuildClosure = false
+
     // Additional libraries that are part of the j2objc release
     // TODO: warn if different versions than testCompile from Java plugin
     // TODO: just import everything in the j2objc/lib/ directory?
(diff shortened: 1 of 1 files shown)


j2objc.gradle
```diff
@@ -135,7 +135,13 @@ class J2objcPluginExtension {
     String translateFlags = "--no-package-directories"
     // -classpath library additions from ${projectDir}/lib/, e.g.: "json-20140107.jar", "somelib.jar"
     String[] translateClassPaths = []
-    
+
+    // WARNING: Do not use this unless you know what you are doing.
+    // If true, incremental builds will be supported even if --build-closure is included in
+    // translateFlags.  This may break in unexpected ways if you change the dependencies (ex. the
+    // translateClassPaths) of your project without doing a 'clean'.
```
brunobowden: Single space after "."

Change last sentence to:

This may break the build in unexpected ways if you change the dependencies (e.g. adding new files or changing translateClassPaths). When you change the dependencies and the build break, you need to do a clean build.

advayDev1:
PTAL

PR #87, opened 2015-06-02
remove j2objcXcode task description

confile:


brunobowden:
@confile - this is redundant after my commit, sorry that I needlessly asked you to do this

PR #91, opened 2015-06-03
Pod task UP-TO-DATE checks and cleanup

brunobowden:
- @OutputFile for the podspec and Podfile
- Rename podspec from <project-name>.podspec to podspec

Cleanup:
- Move checkPodDefExistence to task
- Remove unused j2objcConfig: xcodeProject and xcodeJ2objcGeneratedDir
- lineSeparator => “\n” as task can only be run on a mac
- Other formatting and whitespace reduction


brunobowden:
@confile and @advayDev1, please take a look

brunobowden:
Note that this renames the podspec file from <project-name>.podspec to simply podspec. I'm relying on there being a separate buildDir for each project that's generating a podspec. If that's a bad assumption then let me know.

j2objc.gradle
```diff
@@ -1040,8 +1002,13 @@ class J2objcCopyTask extends DefaultTask {
  */
 class J2objcPodTask extends DefaultTask {
 
+    // Generated ObjC source files
     @InputDirectory File srcDir
 
+    // CocoaPods podspec file that's used by the Podfile
+    @OutputFile File getPodspecFile() { new File(project.buildDir, "podspec") }
```
advayDev1: better to put this in build/outputs/podspec maybe?

j2objc.gradle
```diff
@@ -1074,32 +1039,34 @@ class J2objcPodTask extends DefaultTask {
...
-            "s.xcconfig = { 'HEADER_SEARCH_PATHS' => '${getJ2ObjCHome()}/include', 'LIBRARY_SEARCH_PATHS' => '${getJ2ObjCHome()}/lib' }" + lineSeparator +
-            "end")
-          
+        // Line separator assumed to be "\n" as this task can only be run on a Mac
+        String podspecFileContents =
+            "Pod::Spec.new do |s|\n" +
+            "s.name = '${podName}'\n" +
+            "s.version = '1.0'\n" +
+            "s.summary = 'Generated by the j2objc.gradle plugin.'\n" +
+            "s.source_files = 'j2objc/**/*.{h,m}'\n" +
```
advayDev1: Where do these j2objc/ subpaths come from?
Should they be some function of destDir in the config?
(and then destDir as an additional input to this task)

j2objc.gradle
```diff
@@ -1040,8 +1002,13 @@ class J2objcCopyTask extends DefaultTask {
...
+    // Generated ObjC source files
     @InputDirectory File srcDir
 
+    // CocoaPods podspec file that's used by the Podfile
+    @OutputFile File getPodspecFile() { new File(project.buildDir, "podspec") }
+    @OutputFile File getPodFile() { new File(project.j2objcConfig.xcodeProjectDir, "Podfile") }
+
     // j2objcConfig dependencies for UP-TO-DATE checks
     @Input String getJ2ObjCHome() { return J2objcUtils.j2objcHome(project) }
     @Input String getXcodeProjectDir() { return project.j2objcConfig.xcodeProjectDir }
```
advayDev1: i'm confused by why some of these are in the build directory, some are in the xcode directory, some are in the destination directory.  we should be consistent.  everything intermediate should be in the build directory, and all things you might want to use outside of our plugin should be in a single configurable output directory would be my suggestion

j2objc.gradle
```diff
@@ -1063,8 +1030,6 @@ class J2objcPodTask extends DefaultTask {
             throw new InvalidUserDataException(message)
         }
 
-        def lineSeparator = System.getProperty("line.separator")
-
         // Resource Folder is copied to buildDir where it's accessed by the pod later
         String j2objcResourceDirName = "j2objcResources"
         String j2objcResourceDirPath = "${project.buildDir}/${j2objcResourceDirName}"
```
advayDev1: don't the resources need to be used later on by other systems outside of gradle?  if so, they should be either in build/outputs/SOMETHING or a user configurable directory.

advayDev1:
comments provided

j2objc.gradle
```diff
@@ -1074,32 +1039,34 @@ class J2objcPodTask extends DefaultTask {
...
+        File podspecFile = getPodspecFile()
+        podspecFile.write(podspecFileContents)
+
         // link the podspec in pod file
-        File podFile = new File(getXcodeProjectDir(), "Podfile")
+        File podFile = getPodFile()
         if (!podFile.exists()) {
             // TODO: offer to run the setup commands
             def message =
                     "No podfile exists in the directory: ${xcodeProjectDir}\n" +
```
advayDev1: should this be getXcodeProjectDir()

j2objc.gradle
brunobowden: @confile - I'll leave the directories as is for now... but we'll likely change them at some point in the future during this big overhaul

j2objc.gradle
brunobowden: This is defined on line 1218 and is already an @Input. Not sure if this should be part of j2objcConfig or at least extracted to a variable. Just adding TODO for now.

            tasks.create(name: 'j2objcPod', type: J2objcPodTask,
                dependsOn: 'j2objcTest') {
                description 'Depends on j2objc translation, create a Pod file link it to Xcode project'
                srcDir = file("${buildDir}/j2objc")
            }

j2objc.gradle
brunobowden: I believe the Podfile needs to be in the Xcode directory. The podspec can be temporary just for the moment.

j2objc.gradle
brunobowden: Fixed getXcodeProjectDir(), pushing change now...

j2objc.gradle
advayDev1: i believe, even though it an input, nothing will actually put the source files there since destDir can be arbitrary, but yes TODO sg

advayDev1:
LGTM

PR #104, opened 2015-06-05
Fix XcodeTask not being able to find podspec

brunobowden:
- Ignore local.properties file from Android Studio
- XML_LEGACY_SETTINGS_IMPORTED, which I presume is a one time import


brunobowden:
@confile - please take a look

@advayDev1 - the fix is simple, so you can continue making progress elsewhere

src/main/groovy/com/github/brunobowden/j2objcgradle/tasks/J2objcXcodeTask.groovy
```diff
@@ -41,7 +41,7 @@ class J2objcXcodeTask extends DefaultTask {
 
     // CocoaPods podspec file that's used by the Podfile
     @OutputFile
-    File getPodspecFile() { new File(project.buildDir, "podspec") }
+    File getPodspecFile() { new File(project.buildDir, "${project.name}.podspec") }
```
confile: Where is ${project.name}.podspec defined in?

confile:
@brunobowden Have a look at my question beside that LGTM

brunobowden:
@confile - it's defined on that line. Not anywhere else, though perhaps it
should be.

brunobowden:
Changed PodName to an @Input method:

    @Input
    String getPodName() { "j2objc-${project.name}" }

Issue #106, opened 2015-06-06
Load j2objcHome from properties file or env variables


advayDev1:
Similar to how sdk.dir is loaded from local.properties file.

PR #107, opened 2015-06-06
Load j2objcHome from local.properties or environment variable; Fixes …

advayDev1:
…#106.

### DIFF
--- a/src/main/groovy/com/github/brunobowden/j2objcgradle/tasks/J2objcUtils.groovy
+++ b/src/main/groovy/com/github/brunobowden/j2objcgradle/tasks/J2objcUtils.groovy
@@ -86,17 +86,28 @@ class J2objcUtils {
     // MUST be used only in @Input getJ2ObjCHome() methods to ensure UP-TO-DATE checks are correct
     // @Input getJ2ObjCHome() method can be used freely inside the task action
     static def j2objcHome(Project proj) {
-        def result = proj.j2objcConfig.j2objcHome
+        def localPropertiesFile = new File(proj.rootDir, 'local.properties')
+        String result = null
+        if (localPropertiesFile.exists()) {
+            Properties localProperties = new Properties();
+            localPropertiesFile.withInputStream {
+                localProperties.load it
+            }
+            result = localProperties.getProperty('j2objc.home')
+        }
+        if (result == null) {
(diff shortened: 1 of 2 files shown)


src/main/groovy/com/github/brunobowden/j2objcgradle/tasks/J2objcUtils.groovy
```diff
@@ -86,17 +86,28 @@ class J2objcUtils {
...
     // @Input getJ2ObjCHome() method can be used freely inside the task action
     static def j2objcHome(Project proj) {
-        def result = proj.j2objcConfig.j2objcHome
+        def localPropertiesFile = new File(proj.rootDir, 'local.properties')
+        String result = null
+        if (localPropertiesFile.exists()) {
+            Properties localProperties = new Properties();
+            localPropertiesFile.withInputStream {
+                localProperties.load it
+            }
```
brunobowden: I found this to be a clearer way of loading the properties file:

    Properties localProperties = new Properties();
    localProperties.load(new FileInputStream(localPropertiesFile))

src/main/groovy/com/github/brunobowden/j2objcgradle/tasks/J2objcUtils.groovy
```diff
@@ -86,17 +86,28 @@ class J2objcUtils {
...
+            result = localProperties.getProperty('j2objc.home')
```
brunobowden: You've confirmed that if `j2objc.home` doesn't exist then getProperty returns null?

advayDev1: javaDoc says so.

src/main/groovy/com/github/brunobowden/j2objcgradle/tasks/J2objcUtils.groovy
advayDev1: no, that leaves the File descriptor open.  withInputStream ensures things get closed properly.

brunobowden: I'll fix the prefixes.properties loading to match.

Issue #85, opened 2015-06-02
Flags to configure include directories, linked libraries, etc.


advayDev1:
As we use gradle objc compilation and cocoa pods, we should not have the user tell us multiple times common attributes like include directories and linked libraries by overriding command lines as flags.  Instead provide separate flags for each of these, and then convert those the command-line as appropriate for each tool in the chain.

brunobowden:
prefix.properties is a bad example of this. I wrote the code to parse the
command line and find the file. Much better done as a distinctive set of
controls.

advayDev1:
so i think i got all the right inputfiles in #110 except prefix.properties, can you capture that one here?

PR #118, opened 2015-06-08
added --static-accessor-methods to translateFlags

confile:
--static-accessor-methods should be a default because it is a must have to use generated Enums with Swift. Makes access very easy.

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
@@ -59,7 +59,7 @@ class J2objcPluginExtension {
     // A list of all possible flag can be found here
     // https://github.com/google/j2objc/blob/master/translator/src/main/resources/com/google/devtools/j2objc/J2ObjC
     // .properties
-    String translateFlags = "--no-package-directories"
+    String translateFlags = "--no-package-directories --static-accessor-methods"
     // -classpath library additions from ${projectDir}/lib/, e.g.: "json-20140107.jar", "somelib.jar"
     String[] translateClassPaths = []
 
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
@@ -57,8 +57,7 @@ class J2objcPluginExtension {
     // TRANSLATE
     // Flags copied verbatim to j2objc command
     // A list of all possible flag can be found here
-    // https://github.com/google/j2objc/blob/master/translator/src/main/resources/com/google/devtools/j2objc/J2ObjC
-    // .properties
+    // https://github.com/google/j2objc/blob/master/translator/src/main/resources/com/google/devtools/j2objc/J2ObjC.properties
     String translateFlags = "--no-package-directories --static-accessor-methods"
     // -classpath library additions from ${projectDir}/lib/, e.g.: "json-20140107.jar", "somelib.jar"
     String[] translateClassPaths = []


advayDev1:
Shouldn't you ask j2objc to make this a default on their end?

advayDev1:
Actually speaking of which I think we should get rid of --nopackagedirs
too.

advayDev1:
Actually, I think we should be consistent with j2objc wherever possible, in that we never set any default flags, including removing nopackagedirs already there

brunobowden:
@Confile - there's a few trade offs with this so you might not always want it. See in this discussion:
https://groups.google.com/forum/#!topic/j2objc-discuss/dGSAigkaakI Perhaps we don't need the package-directories as we're now compiling with Clang outside of Xcode. Just put in a TODO for now as I'd rather get it working again and then look at improvements.

You can fix the link to the J2ObjC properties file:
https://github.com/google/j2objc/blob/master/translator/src/main/resources/com/google/devtools/j2objc/J2ObjC.properties

confile:
@advayDev1 Why don't we need the ``nopackagedirs``Objective-C has a flat package structure. I see no reason to skip this flag?

confile:
@brunobowden If a user needs a different config he can change it but so far I think using ``--static-accessor-methods`` is a good and easy start. Lets discuss the ``nopackagedirs``in another thread.

advayDev1:
Good point about separating threads.  Still I see no compelling reason our
defaults should be unequal to j2objc even for static access or methods.
Consider a user migrating from j2objc scripts to the plugin.  The principle
of least surprise would dictate that the defaults in the plugin would be as
close to the binary as possible.

I do however see a good reason to have a DSL method for easy configuration
of swift.  You could have a method in the extension class that when called
sets the swift-specific conventions you desire like this flag and maybe
also the bridging header functionality.  Not everyone uses swift (ex. Me)
either.

advayDev1:
(p.s. i see that you've been working with j2objc here anyway: https://github.com/google/j2objc/issues/520 - given that the functionality was working it seems only 5 days ago, why not let it mature and have j2objc set the default there?)

advayDev1:
i believe per yesterday's meeting we decided this was not the right thing to do?  @brunobowden

brunobowden:
@tomball - IIRC, you're planning to make "--static-accessor-methods" a
default for j2objc?

tomball:
No, it can't be a default since "--static-accessor-methods" adds code that
can make large apps too big for App Store acceptance. We're considering
some sort of "Swift-friendly" flag that sets more specific flags
by default. Currently that's not there because only
--static-accessor-methods is necessary for access to static methods from
Swift. You may want to consider a Swift-specific rule or setting to do
something similar.

advayDev1:
I'm a fan of the swift specific DSL method, see comments above...

Also @confile, this PR claims it is from an unknown repo.

brunobowden:
@tomball - I do all my ios development exclusively in Swift, so I'd be very receptive to that

confile:
@tomball So you do not recommend to use the ``--static-accessor-methods`` flag?

confile:
@advayDev1 I deleted the branch. If we need this I can redo it any time.

confile:
I am doing also everything in Swift.

advayDev1:
SG, by the way @confile - i have no issue with this flag, just not in this variable by default.  i think that having a for example swiftConventions DSL method in J2objcPluginExtensions might be the way to go.  That method would enable this flag, and maybe others you think are Swift best practices.  in general, i think translateFlags should be empty, as it avoids confusion of users who are trying to put their own flags (for example, should they first parse translateFlags to remove only certain flags they don't want?  should they assign the variable or just append to it?).  If we finish #85, this should be much easier.

Since you deleted the branch, closing the PR as well.

confile:
@advayDev1 SG please file an issue for your idea.

tomball:
@confile I don't support setting --static-accessor-methods true as the
default, because it generates extra code. Since the gradle plugin is new
and so projects that use it likely to be newer and more likely to use
Swift, you folks can make it the default.

confile:
@tomball But, if it add extra code your makes apps to big is a bad choice.

tomball:
We translate static methods into C functions for two reasons: they save a
little space and are slightly faster because there's no class method
overhead. However, Swift only imports Objective C methods, so the flag puts
them back at a small cost of space. Performance is still faster than only
class methods, because any calls to those static methods in transpiled code
call the functions directly. If you want Swift access without the method
wrapper overhead, call the C functions directly. Otherwise, use the flag
for code that's more "Swift like". The choice for your app is yours, not
the j2objc team's.

brunobowden:
I think a --swift will be popular. Certainly with @confile and myself but
also likely to a lot of other people as well.

PR #119, opened 2015-06-08
Working build: 0.2.0 version bump + add missing javax_inject library

brunobowden:
These changes allow a working build for my local project. The Xcode
compile and debugging haven’t been tested yet but it looks promising.

- Bump version to 0.2.0-alpha
- Fix missing javax_inject library for native compile + clear
organization
- added j2objcConfig for testMinExpectedUnitTests support and
testMinExpectedUnitTestsDisable
- try catch block for ‘j2objcTest’, warns on known Mockito.spy() crash
- J2objcUtils.matchNumberRegex(…) to extract int from proces.exec(…)

Minor:
- Removed indentation for “j2objConfig {“ to facilitate copy and paste
- Consolidate naming for srcGenDir throughout tasks
- Better command example for creating Podfile

TESTED=yes


src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
```diff
@@ -142,11 +142,18 @@ class J2objcNativeCompilation {
...
-                    linker.args '-lguava', '-ljsr305'
-                    linker.args '-ljre_emul', '-licucore', '-lz', '-lj2objc_main', '-lc++'
-                    linker.args '-framework', 'foundation', '-framework', 'Security'
+
+                    // J2ObjC provided libraries and search path:
+                    linker.args '-lguava', '-lj2objc_main', '-ljavax_inject', '-ljre_emul', '-ljsr305'
                     linker.args "-L$j2objcPath/lib"
 
+                    // J2ObjC iOS library dependencies:
+                    linker.args '-lc++'                    // supports ??
```
brunobowden: @tomball - I'd appreciate your input on cataloging what each of the library dependencies are needed for. The ones I'm not sure about are -lc++ and -framework foundation.

brunobowden:
@advayDev1 and @confile - please take a look. I'll likely get this working tomorrow.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
```diff
@@ -51,9 +52,11 @@ class J2objcPluginExtension {
     // Warning will ask user to configure this within j2objcConfig
     String cycleFinderFlags = null
     // Expected number of cycles, defaults to all those found in JRE
+    // This is an exact number rather than minimum as any change is significant
```
advayDev1: period at end here

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
```diff
@@ -110,11 +113,17 @@ class J2objcPluginExtension {
...
 
+
     // TEST
     // Skip test task if true
     boolean testSkip = false
     // Flags copied verbatim to testrunner command
     String testFlags = ""
+    // Error if it runs less than the expected number of tests
+    // It is set as a minimum so adding a unit test doesn't break the j2objc build
+    int testMinExpectedUnitTests = 1
```
advayDev1: how about just testMinExpectedTests,  there's nothing semantically causing these to be unit tests

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcTestTask.groovy
```diff
@@ -105,37 +112,89 @@ class J2objcTestTask extends DefaultTask {
...
+                    "Copy the command from \"Command:\" onwards, then try varying the command\n" +
+                    "to drop tests and figure out which ones are causing the failures.\n"
+                    "Then disable them as described above.\n"
+            if (exception.getMessage().find("finished with non-zero exit value 139")) {
+                message +=
+                        "\n" +
+                        "\"non-zero exit value 139\" indicates a process crash, most likely\n" +
+                        "caused by a segmentation fault (SIGSEGV) in user space.\n" +
+                        "\n" +
+                        "Mockito.spy() is known to cause this issue, disable tests that use it:\n" +
```
advayDev1: Do you intend to keep the list of known issues up-to-date in the code?
Perhaps instead a link to all crash related issues?
https://github.com/google/j2objc/issues?q=is%3Aissue+crash+is%3Aopen+

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcTestTask.groovy
```diff
@@ -105,37 +112,89 @@ class J2objcTestTask extends DefaultTask {
...
+                    "If there are legitimately fewer tests then adjust as follow:\n" +
+                    "\n" +
+                    "j2objcConfig {\n" +
+                    "    testMinExpectedUnitTests ${testCount}\n" +
+                    "}\n"
+                    "\n" +
+                    "The check can also be disabled (not best practice):\n" +
+                    "\n" +
+                    "j2objcConfig {\n" +
+                    "    testMinExpectedUnitTestsDisable true}\n" +
```
advayDev1: no } at end of string

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcTestTask.groovy
```diff
@@ -105,37 +112,89 @@ class J2objcTestTask extends DefaultTask {
...
+                    "}\n"
+            throw new Exception(message)
+        } else if (testCount != getTestMinExpectedUnitTests()) {
+            def message =
+                    "testMinExpectedUnitTests can be increased:\n" +
```
advayDev1: i don't think this is a best practice and therefore we shouldn't suggest it to users.  users should add tests whenever they add new code, but shouldn't have to (or think they should) keep changing a config to keep up.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcTranslateTask.groovy
```diff
@@ -251,19 +250,20 @@ class J2objcTranslateTask extends DefaultTask {
                 errorOutput output
             }
 
-        } catch (e) {
-            def processOutput = output.toString()
+        } catch (Exception exception) {
```
advayDev1: Why use Exception here?

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcUtils.groovy
```diff
@@ -195,4 +195,25 @@ class J2objcUtils {
     static def filterJ2objcOutputForErrorLines(String processOutput) {
         return processOutput.tokenize('\n').grep(~/^(.*: )?error:.*/).join('\n')
     }
+
+    static int matchNumberRegex(String str, String regex) {
+        def matcher = (str =~ regex)
+        if (!matcher.find()) {
+            // Can't use logger.error for output of str, so put it in the exception
+            throw new Exception(
```
advayDev1: IllegalArgumentException would be most appropriate it seems and below.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
```diff
@@ -142,11 +142,18 @@ class J2objcNativeCompilation {
                     objcCompiler.args '-std=c11'
 
                     linker.args '-ObjC'
-                    linker.args '-lguava', '-ljsr305'
-                    linker.args '-ljre_emul', '-licucore', '-lz', '-lj2objc_main', '-lc++'
-                    linker.args '-framework', 'foundation', '-framework', 'Security'
+
+                    // J2ObjC provided libraries and search path:
+                    linker.args '-lguava', '-lj2objc_main', '-ljavax_inject', '-ljre_emul', '-ljsr305'
```
advayDev1: we should understand what the difference between "link all the libraries under j2objcdist/lib/*.a" and "the standard set of j2objc libraries" is, if anything.  for example, i don't use javax_inject and it just bloats my library, but you may not use guava, etc.  maybe we should just link to the actually required ones, jre_emul, jsr305, and j2objc_main, and leave guava, inject, etc. as options.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcCycleFinderTask.groovy
```diff
@@ -125,15 +126,15 @@ class J2objcCycleFinderTask extends DefaultTask {
...
                     "    \$ (cd jre_emul && make java_sources_manifest)\n" +
                     "3) Configure j2objcConfig in build.gradle so CycleFinder uses j2objc source:\n" +
-                    "    j2objcConfig {\n" +
-                    "        cycleFinder true\n" +
-                    "        cycleFinderFlags (\n" +
-                    "                \"--whitelist \${projectDir}/../../<J2OBJC_REPO>/jre_emul/cycle_whitelist.txt " +
+                    "j2objcConfig {\n" +
+                    "    cycleFinder true\n" +
+                    "    cycleFinderFlags (\n" +
+                    "            \"--whitelist \${projectDir}/../../<J2OBJC_REPO>/jre_emul/cycle_whitelist.txt " +
```
confile: cycle_whitelist.txt I not part of the j2objc [distribution package](https://github.com/google/j2objc/releases/download/0.9.7/j2objc-0.9.7.zip). It is only in the source. Maybe we can access the file directly from the github repo.

confile: May be we can also download the ``sourcefilelist`` automatically.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
tomball: -lc++ links in the C++ runtime, used by the protobuf runtime; since the C++ runtime doesn't have any Objective C classes, it adds nothing to the app unless needed. The Foundation Framework defines the core Objective-C classes for iOS: NSObject, NSString, etc. IIRC, this link flag isn't necessary if the -ObjC flag is used instead.

brunobowden: Added note

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
brunobowden: Done and corrected a few other lines as well

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
brunobowden: Shortened to "testMinExpectedTests"

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcTestTask.groovy
brunobowden: Good idea. Changed link to all crashes.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcTestTask.groovy
brunobowden: Fixed

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
```diff
@@ -110,11 +113,17 @@ class J2objcPluginExtension {
...
+    // Disable above check
+    boolean testMinExpectedUnitTestsDisable = false
```
advayDev1: simpler to not have a bool mask an int to 0, and just have the user set the int to 0 when they desire

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcTestTask.groovy
brunobowden: That's why it's a minimum rather than an exact amount... or they can leave it at the default of testMinExpectedTests = 1, so it will just throw an error if there are no tests at all, which is a warning sign.

For the case of cycleFinder. Every cycle added or removed is a big deal, so that's an exact number as you want to know each time it changes.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcTranslateTask.groovy
brunobowden: Consistency with the other try / catch code in other tasks.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcTestTask.groovy
advayDev1: understood that - all of that makes sense.

my above comment is about your info message when testCount != min tests.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcUtils.groovy
brunobowden: Changed to IllegalArgumentException

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcCycleFinderTask.groovy
brunobowden: @confile - good suggestion but cycle_whitelist.txt is a generated file that isn't in the repo. You also need to generated jre_emul/build_result/java_sources.mf.

@tomball - we should talk when we meet up about whether there's an easier way to do the cycle white listing. Perhaps with some items being included with the j2objc releases. A lower priority issue.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
brunobowden: Agreed. Removed testMinExpectedTestsDisable. It was a legacy from when it was exact matching rather than a minimum. Much better this new way.

advayDev1:
thanks @brunobowden !  at least we now know the plugin works for at least 2 projects.  for now, if you just want to leave a TODO for https://github.com/j2objc-contrib/j2objc-gradle/pull/119#discussion-diff-31896406 that's fine with me.

LGTM otherwise.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
brunobowden: Added TODO for the above

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcTestTask.groovy
brunobowden: This is only a logger.info message that people likely won't see. I'll take it out if it causes problem. The main case to catch is testCount == 0, I think it's important to emphasize how important unit tests are J2ObjC.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
brunobowden: Added a TODO about this. We'll discuss on Wednesday.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcTestTask.groovy
advayDev1: Wouldn't a better solution to be logger.warn if min tests is 0 and not do
this logger.info or at least make it a logger.debug.

brunobowden: Changed as you described Advay

PR #121, opened 2015-06-08
Remove all translate flags defaults,

advayDev1:
including --no-package-directories
to be consistent with j2objc.

moved thread here @confile @brunobowden.  argument remains the same, namely:
I see no compelling reason our defaults should be unequal to j2objc.
Consider a user migrating from j2objc scripts to the plugin.  The principle
of least surprise would dictate that the defaults in the plugin would be as
close to the binary as possible.

In this case, for example, when I first used the plugin I was confused because all my files were now flat.
It can also cause problems if I for example had two classes with different package prefixes in different packages.  A default --no-package-directories would cause a name conflict, while in the default state everything builds fine.

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
@@ -59,7 +59,7 @@ class J2objcPluginExtension {
     // A list of all possible flag can be found here
     // https://github.com/google/j2objc/blob/master/translator/src/main/resources/com/google/devtools/j2objc/J2ObjC
     // .properties
-    String translateFlags = "--no-package-directories"
+    String translateFlags = ""
     // -classpath library additions from ${projectDir}/lib/, e.g.: "json-20140107.jar", "somelib.jar"
     String[] translateClassPaths = []
 


confile:
@advayDev1 No I don't thank that your suggestion is the right path. The gradle plugin should be easy to start with J2objc from zero. This way we should provide a good preconfigured system. If users gain experience they can configure whatever they want.

What is the advantage of not using: ``--no-package-directories``?

confile:
Objective-c has no package structure everything is flat so it does not matter if files with the same name are in different folders there still is a collision.

advayDev1:
Class collision is prevented by different package prefixes.  File collision
is not.

confile:
@advayDev1 Assume the following. In Java your classes are named like the file name it is located in and you have only one class per file. Now, having:
```
com/example/app/View.java
com/example/base/View.java
```
This works in Java but you have to ensure that you have different package prefixes otherwise you get a file collision. 

@advayDev1 What is the reason you does not use ``--no-package-directories``? Is it just to get a better organization of your files?

advayDev1:
Yes, I understand that @confile.  But there's no way you can prevent your projects and all your transitive dependencies (when using --build-closure) from not having the same named file in 2 different directories.  This happens not rarely for me.  You *can* fix this with package prefixes as you said.

Also, I understand that the two of us just have different styles.  But when it is an aesthetic decision like this, with no obvious better answer (let's say), consistency is a very important factor.  Unless there is a really compelling reason to be inconsistent, we should be consistent by default.  Both with the default flag of j2objc, and with how j2objc itself organizes, for example, the jre emul code.  All of it is with package directories (which is why you import "com/java/lang/Integer.h" for example).

advayDev1:
@confile , also, I don't believe the experience of the 3 of us working on this plugin is more extensive than the j2objc project as a whole.  If you believe this default is better (in that, for nearly all projects or at least a majority, it would be better), then why not convince j2objc and their community to make the change for everyone as a bug there.

confile:
okay you are right lets skip this.

brunobowden:
Leave this open for now. It's something we can talk about when we meet Google on Wednesday.

brunobowden:
This was a legacy flag that we needed when we were doing the build inside of Xcode. Now that we're doing the build outside of Xcode we nolonger need it. @advayDev1 has confirmed that debugging works with nested folders - I'll try that out after his merge.

brunobowden:
LGTM

PR #124, opened 2015-06-10
Packs architectures into fat library with 'lipo'

advayDev1:
Also Adds building of ios i386 architecture.


advayDev1:
So I've tested this on simulators for all of:
iPad {2, Air, Retina}, iPhone {4s, 5, 5s, 6 Plus, 6}, Resizable iPad, Resizable iPhone

I suspect we'll need another PR to make things work on device as well.  There's a simulator vs. phone flag in there, but one of you (@brunobowden or @confile ) who have a device should make that PR.

brunobowden:
@advayDev1 - LGTM

PR #131, opened 2015-06-11
Workaround for plugins {} + Usage Instructions

### LINKED ISSUE #130: fix plugins {} syntax - java plugin ordering
The plugins syntax for loading the j2objc plugin prevents you from loading the java plugin beforehand. That means that the j2objc plugin needs to operate assuming that the java plugin is loaded afterwards.

    // plugins prevent applying java plugin beforehand
    plugins {
      id "com.github.j2objccontrib.j2objcgradle" version "0.2.0-alpha"
    }
    apply plugin: 'java'

brunobowden:
- Fixes #130
- build clean up to remove redundant config

@advayDev1, @confile - please take a look


build.gradle
```diff
@@ -14,28 +14,37 @@
...
-
 group = 'com.github.j2objccontrib.j2objcgradle'
-version = '0.2.0-alpha'
+version = '0.2.1-alpha'
 
-dependencies {
-    testCompile group: 'junit', name: 'junit', version: '4.12'
-}
+pluginBundle {
+    website = 'http://www.github.com/j2objc-contrib/j2objc-gradle/'
```
advayDev1: https please

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy
```diff
@@ -167,7 +172,8 @@ class J2objcPlugin implements Plugin<Project> {
                 libDir = file("${buildDir}/binaries/${project.name}-j2objcStaticLibrary")
                 packedLibDir = file("${buildDir}/packedBinaries/${project.name}-j2objcStaticLibrary")
             }
-            lateDependsOn(project, 'assemble', 'j2objcAssemble')
+            // TODO: figure out how to make this work
+            // lateDependsOn(project, 'assemble', 'j2objcAssemble')
```
advayDev1: this is critical, otherwise these steps aren't part of the build.
why did it just stop working now?  can you give some more detail.

advayDev1: ah i see #130 - please see my suggestion there to undo this

brunobowden:
@advayDev1 - please take a look, lateDependsOn is now restored

advayDev1:
a few things:
1) please take a look at #137 for versioning suggestions; i don't think you should do it in one cl.  also while your PR says 0.2.2, the file says 0.2.1
2) do we have to get rid of the shortname 'j2objc' for the plugin?  if so, change README.md as well (beacuse you renamed the META-INF file).  or can we keep the shortname META-INF file as well?

brunobowden:
1) The commit message and the file contents match. The challenge is that the title and initial comment don't get updated. In the future, I will do the version update and milestone tag in a separate CL.
2) I believe that this may need an approval step for this. Secondly, what we're doing now follows the convention for how pretty much all other plugins are loaded. I will also talk with them about this further.

I will work on the README changes.

advayDev1:
hmm. for me the PR and your commit says 0.2.2, and build.gradle says 0.2.1-alpha...

advayDev1:
LGTM

brunobowden:
I've removed mention of plugin version except one line to make it consistently 0.2.1. I'll do the version change in a separate commit.

Merging now...

brunobowden:
Did a bunch of final testing but I think that's looking good

PR #135, opened 2015-06-11
added flag to skip tests

confile:


advayDev1:
the correct way to do this is now to just run `j2objcAssemble` (which does not run the test).  if you run `build` tests are supposed to be run.  you can skip them by doing ex: `./gradlew build -x j2objcText`

brunobowden:
Is there a way of having cycleFinder default disabled and then enabled by
config if needed? Or do we need to keep that flag?

advayDev1:
when creating the task in the plugin you could do `enabled = false`, but...

i'd argue the right thing to do is leave the cycleFinder out of the dependency of the translation step by default.  similar to how we removed tests from translation.

advayDev1:
@confile, Also if you want to permanently skip j2objc tests for a particular build.gradle, put this:
```
j2objcTest {
    enabled false
}
```

You can do this with any named task (i.e. you can configure it after plugin application in your own build file).

Since this is already supported the standard gradle way, closing.  Please reopen as needed.

confile:
@advayDev1 Why did you close this pull request? We should be consistent for cycle finder we have a skip flag as well.

advayDev1:
CycleFinder's skip flag is going away too, about to file the PR.

The right way to do this is as above, and I'm adding the functionality to do that.

advayDev1:
See #145

confile:
ok

confile:
@advayDev1 You should add: 
```
j2objcTest {
    enabled false
}
```
to the docs.

advayDev1:
Can you make the PR?

confile:
@advayDev1 yes. Since you are an expert with git. I made a fork of the master. How can I update my master fork that it contains the latest commits?

brunobowden:
You should do all the development on a branch and not use master. That's
both safer and easier for you.

Then follow the instructions in CONTRIBUTING.md to rebase your master and
branches on the upstream changes.

advayDev1:
First
git fetch upstream/master
Then
git checkout master
git merge upstream/master

Note that you really should create a new branch for every PR, otherwise the
instructions in contributing.MD will not work.

confile:
@advayDev1 Thanks!

confile:
@brunobowden I create a branch for each pull request, but I forked your master first and created branches from this fork.

Issue #101, opened 2015-06-05
Publish this as a gradle community plugin


advayDev1:
After #55 is done, need to decide how/where/when to publish and name our plugin

brunobowden:
This is now published but there is a serious issue I'll file separately:
https://plugins.gradle.org/plugin/com.github.j2objccontrib.j2objcgradle

Issue #97, opened 2015-06-04
Style Guidelines for CONTRIBUTING.md


brunobowden:
Start developing style guidelines for the project. Ideally based of some existing style guidelines, which are extended for this particularly use case.

- return statements
- project.j2objcConfig access and @Input and @Output connections
- project.with

advayDev1:
Here's the standard Groovy style guide:
http://www.groovy-lang.org/style-guide.html

Since we are a 100% groovy project, I think we should stick as closely as we can to them.  A mixed Java/Groovy style I think will be even more confusing.

brunobowden:
Looks good, we should follow this and then add extra stuff specific to Gradle. I'll work on that.

advayDev1:
Ah good point.

advayDev1:
Suggest we say that our style is Groovy style guide, and where it does not specify and does not conflict, back-off to Google Java style guide:
https://google-styleguide.googlecode.com/svn/trunk/javaguide.html

Also, we need to make a decision on top-level method calls in statements.  We are very inconsistent between `java.call(arg)` style and `groovy.call arg` style.

brunobowden:
I personally prefer parentheses as I find it more familiar. It sounds like the Groovy Style Guide isn't clear on this but I suspect it would make it easier for other engineers contributing.

What's your thoughts on this @advayDev1?

advayDev1:
I'm torn.  I'm huge on consistency so that style things become more mechanical and less subject to reviewer dispute.  If we're going to split into parens and no-parnes, I'd suggest the following rules:

1.  If the statement is solely a DSL/property setting method (ex. 'translateArgs in j2objcConfig'), do not use parentheses on the top-level method.
1.  If the statement is a logging method of the form logger.LEVEL or log.LEVEL, do not use parentheses on the top-level method.
1.  Otherwise, always use parentheses.

Thoughts?

PR #137, opened 2015-06-11
Recommendations for proper versioning of the plugin

advayDev1:
Want to make sure we don't have inconsistencies between the published
versions and the commits in git.

Issue #97 and #101

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -87,3 +87,18 @@ git push -f
 
 For guidance on doing the merge itself, see
 https://git-scm.com/book/en/v2/Git-Branching-Basic-Branching-and-Merging
+
+## Publishing and versioning
+These instructions are only for plugin maintainers; if you are just working
+on issues or pull requests, you can ignore this section.
+
+We'd like to maintain proper versioning history for the plugin.  This requires
+a little bit of coordination between in-code versions, GitHub tags, and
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -87,3 +87,18 @@ git push -f
...
+
+We'd like to maintain proper versioning history for the plugin.  This requires
+a little bit of coordination between in-code versions, GitHub tags, and
+published versions on plugins.gradle.org.  The steps are:
+
+1.  As a separate commit, bump the version number in `build.gradle`. Use
+https://semver.org to guide which slot in the version number should be bumped.
+2.  Tag that separate commit as for example `v0.2.3-alpha` and push that
+tag to the j2objc-contrib/j2objc-gradle repository.
+3.  Do a clean build (`./gradlew clean build`), and then publish the new version
```
brunobowden: Here's the full command that cleans, builds and publishes:

./gradlew clean publishPlugins

advayDev1: Actually it has to be
clean build publishPlugins
because publishPlugins does not run the tests.

brunobowden: I went and checked and found the same thing, glad you caught that as well

advayDev1:
@brunobowden ptal

brunobowden:
Looks good

PR #140, opened 2015-06-11
Clarify versioning instructions.

advayDev1:
It is important to tag the master commit that merged the
version PR, NOT the version bump commit itself.


brunobowden:
looks good

PR #148, opened 2015-06-11
Readme for disabling tasks

confile:


README.md
```diff
@@ -52,6 +52,17 @@ The preferred folder structure is:
         ├───Project
         └───ProjectTests
 
+## Disable tasks
+
+You can disable tasks performed by the plugin using the following configuration block in your ``build.gradle``
+
+```
+j2objcTASKNAME {
```
brunobowden: I think it's better to use a working example, e.g. `j2objcTest`

Then the comment becomes:

See the list of defined tasks in [J2objcPlugin.groovy#L70](https://github.com/j2objc-contrib/j2objc-gradle/blob/master/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy#L70)

README.md
```diff
@@ -52,6 +52,17 @@ The preferred folder structure is:
...
+## Disable tasks
+
+You can disable tasks performed by the plugin using the following configuration block in your ``build.gradle``
+
+```
+j2objcTest {
+    enabled = false
+}
+```
+where ``j2objcTest`` can be any task defined [here](https://github.com/j2objc-contrib/j2objc-gradle/blob/master/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy#L70).
```
brunobowden: Capitalize the `w` at the start of the sentence. Also, did you see what I said in my previous comment about replacing `here` with `J2objcPlugin.groovy#L70`? That'll make the source more obvious.

If you disagree with a comment, then feel free to push back and suggest something else. When you respond to some of the comment but not all of it, then I have to spend additional time to figure out what has been done and what hasn't. Best if I don't need to make the same comment twice.

Also, please do a `git rebase -i master` to squash all the commits then `git push -f` to update your GitHub repo.

confile: okay. Thanks for help.

confile:
@brunobowden Done!

PR #149, opened 2015-06-11
Proposed issue/PR workflow.

advayDev1:


CONTRIBUTING.md
```diff
@@ -6,6 +6,20 @@
 4. Push to the branch (`git push origin my-new-feature`)
 5. Create a new Pull Request (once your PR has been reviewed and is ready to submit see 'Preparing your pull request for submission' below).
 
+## Team workflow
+
+1.  An issue is created describing the change required.
+Feature requests should be labeled 'enhancement' while bugs should be labeled 'bug'.
```
brunobowden: I would make the issue necessary for major changes as an indication of what you're planning to avoid the committers disagreeing with you. For minor changes, e.g. a typo, I feel that creating an issue is an unnecessary burden.

brunobowden:
Looks good except make it clear that the issue it optional

CONTRIBUTING.md
advayDev1: good point done.  also made clear that PRs are never optional even if trivial.

Issue #154, opened 2015-06-12
Android plugin projects fail to build when in the same multi-project build


advayDev1:
Quite possibly my own issue, but filing if either of you have seen similar. @confile @brunobowden 
This started occurring only when I switched to the new .jar which downloads its dependencies from plugins.gradle.org.  I'm wondering if the dependent jar versions retrieved from plugins.gradle.org conflict the Android plugin's dependencies...

The error is strange:
```

FAILURE: Build failed with an exception.

* Where:
Build file '/Users/X/android/app/build.gradle' line: 18

* What went wrong:
A problem occurred evaluating project ':app'.
> Failed to apply plugin [id 'com.android.application']
   > Could not create plugin of type 'AppPlugin'.

* Try:
Run with --info or --debug option to get more log output.

* Exception is:
org.gradle.api.GradleScriptException: A problem occurred evaluating project ':app'.
	at org.gradle.groovy.scripts.internal.DefaultScriptRunnerFactory$ScriptRunnerImpl.run(DefaultScriptRunnerFactory.java:76)
	at org.gradle.configuration.DefaultScriptPluginFactory$ScriptPluginImpl$1.run(DefaultScriptPluginFactory.java:148)
	at org.gradle.configuration.DefaultScriptPluginFactory$ScriptPluginImpl.apply(DefaultScriptPluginFactory.java:156)
	at org.gradle.configuration.project.BuildScriptProcessor.execute(BuildScriptProcessor.java:39)
	at org.gradle.configuration.project.BuildScriptProcessor.execute(BuildScriptProcessor.java:26)
	at org.gradle.configuration.project.ConfigureActionsProjectEvaluator.evaluate(ConfigureActionsProjectEvaluator.java:34)
	at org.gradle.configuration.project.LifecycleProjectEvaluator.evaluate(LifecycleProjectEvaluator.java:55)
	at org.gradle.api.internal.project.AbstractProject.evaluate(AbstractProject.java:487)
	at org.gradle.api.internal.project.AbstractProject.evaluate(AbstractProject.java:85)
	at org.gradle.execution.TaskPathProjectEvaluator.configureHierarchy(TaskPathProjectEvaluator.java:47)
	at org.gradle.configuration.DefaultBuildConfigurer.configure(DefaultBuildConfigurer.java:35)
	at org.gradle.initialization.DefaultGradleLauncher.doBuildStages(DefaultGradleLauncher.java:129)
[...]

PR #155, opened 2015-06-12
Suggestion to not use the daemon when testing plugin changes

advayDev1:
To avoid others hitting issues like #154

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,10 @@ apply plugin: 'com.github.j2objccontrib.j2objcgradle'
 // (regular j2objcConfig here...)
 ```
 
+Note that when rapidly developing and testing changes to the plugin by building your own project,
+avoid using the Gradle daemon as issues sometimes arise with the daemon using an old version
+of the plugin jar.  You can stop an existing daemon with `./gradlew --stop` and avoid the daemon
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -76,6 +76,10 @@ apply plugin: 'com.github.j2objccontrib.j2objcgradle'
 // (regular j2objcConfig here...)
 ```
 
+Note that when rapidly developing and testing changes to the plugin by building your own project,
+avoid using the Gradle daemon as issues sometimes arise with the daemon using an old version
+of the plugin jar.  You can stop an existing daemon with `./gradlew --stop` and avoid the daemon
+on a particular build with the `--no-daemon` flag to gradlew.
```
brunobowden: See if you can link to an existing Gradle issue so someone reading can check if it's still an issue.

Also for us, it would be good to get a notification when it's fixed.

brunobowden:
Looks good

README.md
advayDev1: I cannot find such an issue - I'm guessing our usecase of rapidly switching between building tool and using that jar is not common.  If I get better repro steps I will file a bug myself.

Issue #51, opened 2015-05-09
Multiple java project translation and compilation


advayDev1:
I have two projects, say 2 java gradle projects libx and appx.  They live under one root gradle project, and therefore appx has a 'compile' dependency on libx.  Both libx and appx need to be converted to objc.

Ideally, libx's output directory for j2objcCopy would become an additional include directory for appx's j2objcCompile.  In addition, .a files from compiling libx, would be available for linking into appx's j2objcTest phase.  libx's jar file would be on the classpath for appx's translation.  All the tasks should have inter-project dependencies set correctly (I'm guessing I'm going to need to add a new flavor of dependency like 'j2objc').

This should be all done automatically, without --build-closure (a crutch that is getting less useful as my project graph gets more complicated), and in a hermetic environment (no copying files to j2objc's main /include and /lib directories).   Why doesn't this work out of the box and what needs fixing?
- no way to add additional include directories for j2objcCompile
- j2objc* tasks between projects do not have dependency edges (i.e. appx:j2objc* depends only on libx:compile not on libx:j2objcTranslate)
- no way to specify classpath libraries that are not either in $J2OBJC_HOME/lib or in $PROJECT/lib.  (gradle doesn't use $PROJECT/lib to save away a peer, dependent project's jar output)
- j2objc compile tasks don't create/save away .a objects anywhere permanently
- no way to specify objc .a objects to link that are not in $J2OBJC_HOME/lib
- input directories for appx's j2objc tasks need to be set so as to include the output directories of libx's j2objc tasks.  (this ensures gradle runs the right tasks on file update).

Thoughts?  I'm working on figuring this out on my own in my private codebase, if I succeed I'll try to issue a pull request as long as things haven't diverged too far by then.

advayDev1:
(I'm aware the shortcut way of doing this is a single j2objc project that just accumulates the src/main/java directories of all possible source directories of all possible projects, but that prevents me from building/updating my many independent libraries separately).

confile:
I have a similar setup. Just include lib in your main project and translate the main project. This issue is not related to this plugin.

advayDev1:
Can you expound on 'include lib in your main project'?   You mean combine the sources of two different projects?

confile:
I included the lib source in the main project.

advayDev1:
[...]

PR #157, opened 2015-06-12
Gradle magically creates tasks on findByName

### LINKED ISSUE #156: tasks.findByName magically *creates* tasks, preventing future mutation (!!)
after hours of debugging the entirely unhelpful message:
```
Caused by: java.lang.IllegalStateException: Cannot add Mutate rule 
'org.gradle.language.base.plugins.ComponentModelBasePlugin$Rules#closeSourcesForBinaries(org.gradle.platform.base.BinaryContainer, org.gradle.language.base.ProjectSourceSet)'
for model element 'binaries' when element is in state GraphClosed.
```

the issue is the use of findByName('assemble').  This cause Gradle to *create* the assembly task, preventing modification of say the binaries later.

Pulpit:
This is why I hate things that have too much magic.  Only a line by line binary search caused to me find this, since the stack trace is thrown much later when the project's tasks have to be resolved, and nowhere near either the mutation or the findByName.  SHEESH!

@brunobowden I'm fixing this to allow native dependencies, hopefully you can rollback your workaround for plugin syntax once I merge this in.

advayDev1:
Gradle magically creates tasks on findByName
which prevents you from modifying their inputs
later on.

Instead use tasks.all which seems not to have this issue.

Fixes #156

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy
@@ -186,14 +186,19 @@ class J2objcPlugin implements Plugin<Project> {
     // The before task must already exist.
     private def lateDependsOn(Project proj, String afterTaskName, String beforeTaskName) {
         assert null != proj.tasks.findByName(beforeTaskName)
-        def afterTask = proj.tasks.findByName(afterTaskName)
-        if (afterTask != null) {
-            afterTask.dependsOn beforeTaskName
-        } else {
-            proj.tasks.whenTaskAdded { task ->
-                if (task.name == afterTaskName) {
-                    task.dependsOn beforeTaskName
-                }
+        // You can't just call tasks.findByName on afterTaskName - for certain tasks like 'assemble' for
+        // reasons unknown, the Java plugin creates - right there! - the task; this prevents
+        // later code from modifying binaries, sourceSets, etc.  If you see an error
+        // mentioning 'state GraphClosed' saying you can't mutate some object, see if you are magically
+        // causing Gradle to make the task by using findByName!  Issue #156
+
+        // tasks.all cleanly calls this closure on any existing elements and for all elements
+        // added in the future.
+        // TODO: Find a better way to have afterTask depend on beforeTask, without
+        // materializing afterTask early.
+        proj.tasks.all { task ->
+            if (task.name == afterTaskName) {
+                task.dependsOn beforeTaskName
             }
         }
     }


brunobowden:
So you can reorder the 'j2objc' plugin to be applied before the 'java' plugin and it still works?

Looks good assuming the above is true. I guess this'll take a new version spin to test properly with the `plugin {}` syntax.

advayDev1:
I haven't tested that; what I did test was that a series of other issues regarding StateClosed were fixed.  (I'm trying to do #51 and it needs modifying the native binary config.)

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy
```diff
@@ -186,14 +186,17 @@ class J2objcPlugin implements Plugin<Project> {
     // The before task must already exist.
     private def lateDependsOn(Project proj, String afterTaskName, String beforeTaskName) {
         assert null != proj.tasks.findByName(beforeTaskName)
-        def afterTask = proj.tasks.findByName(afterTaskName)
-        if (afterTask != null) {
-            afterTask.dependsOn beforeTaskName
```
brunobowden: @bigguy suggested to try this...
```
private def lateDependsOn(Project proj, String afterTaskName, String beforeTaskName) {
    assert null != proj.tasks.findByName(beforeTaskName)
    if (afterTask != null) {
        afterTask.dependsOn {
            // evaluated after all tasks are created
            proj.tasks.findByName(beforeTaskName)
        }
    }
}
```

advayDev1: I'm OK with either.  They are equally confusing to me but this new
suggestion puts the depends on upfront so I'll do that.

advayDev1: Sorry no that doesn't work.  It is after task which is the one we do not
want to materialize.  That soln requires after task to be found inline.

advayDev1:
due to my response above, is this LG to submit?

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy
```diff
@@ -186,14 +186,17 @@ class J2objcPlugin implements Plugin<Project> {
...
+        // reasons unknown, the Java plugin creates - right there! - the task; this prevents
+        // later code from modifying binaries, sourceSets, etc.  If you see an error
+        // mentioning 'state GraphClosed' saying you can't mutate some object, see if you are magically
+        // causing Gradle to make the task by using findByName!  Issue #156
+
+        // tasks.all cleanly calls this closure on any existing elements and for all elements
+        // added in the future.
+        proj.tasks.all { task ->
+            if (task.name == afterTaskName) {
+                task.dependsOn beforeTaskName
```
brunobowden: LGTM

Post something on StackOverflow to see if there's a better way of doing this and add a TODO. I'd prefer not to iterate through all the tasks by name, it feels like there should be a better way. If Gradle isn't responsive to the SO question, I can give them a nudge over email.

advayDev1:
will do, merging.

confile:
LGTM

Issue #112, opened 2015-06-07
Split translation of main & test sourceSets


advayDev1:
It is more correct to translate & compile main, then translate & compile test as depending on main, in 2 separate phases.  This ensures that the production library has *only* the files needed by main.

At the moment, if tests rely on libraryB, the production library will contain libraryB for no reason.

advayDev1:
Not for summit milestone.

advayDev1:
i will not have a chance to do this any time soon, but it is pretty important.

advayDev1:
with auto-config, your production libraries could be linking with test libraries.

advayDev1:
This is currently my top correctness concern (there are other usability ones though); it also requires a large refactor of the plugin.

Many projects' test files don't end with the word Test, or their *Test classes have support classes that aren't similarly.  All such files are erroneously ending up in the production static library, and also all such tests aren't being run.

Example: https://github.com/JodaOrg/joda-convert/tree/master/src/test/java/org/joda/convert

brunobowden:
I agree on that. It's a big assumption we're making there that's nolonger relevant.

Issue #41, opened 2015-04-27
Auto Resolve Dependencies


confile:
When using:
 
```
translateClassPaths "json-20140107.jar", "javax.inject-1.jar",
translateSourcepaths "some-sources.jav"
```
the dependencies should be resolved automatically.

confile:
@advayDev1 I suppose you know more about this?

advayDev1:
Can you be more specific about what you mean by dependencies?  Are these transitive dependencies of the instant jars on other jars?

confile:
@advayDev1 I mean transitive dependencies.

confile:
I mean that they should be resolved like in: 

```
dependencies {
  compile ...
}
```

advayDev1:
are you asking for:
1) compile dependencies named in the dependencies DSL block should be automatically converted into translate* flags?
or 
2) such dependencies, whether in the dependencies DSL block or mentioned in translate* flags, should have their transitive dependencies included automatically?

confile:
No to make it simple. I want that our plugin auto resolves transitive dependencies. That's it.

advayDev1:
I don't know if we should do that.  I know in my case it would actually be harmful, and I can imagine it would be for many people.  Consider that your projectA uses libB.  libB relies on Guava.  With this, we would use the public Guava jar as referenced in libB's pom file (for example), but that would be wrong, we should instead use the guava in the j2objc dist library.  This could happen for any library.  There is no way to map say a Maven pom file transitive dependency to the name in the j2objc lib directory (afaik).

I assume you want this to make --build-closure easier to manage, as it would be for me as well.  But as soon as I started doing something that wasn't simple (for example, in my case my projectA relies on libB which relies on libC which has some JNI components that I then had to create a libCModified version myself which stubs or replaces with OCNI those JNI components), it did not work to do it automatically.

advayDev1:
So, the right way to do this is use:
project.sourceSets.main.compileClasspath.getAsPath()
project.sourceSets.test.compileClasspath.getAsPath()

they can just be added to translateClasspath.

Also, #112 does not need to be done to do this, only nice-to-have.

Need to solve #180 because otherwise the junit.jar for example we reference in the translateClasspath may be of a different version than the junit.jar in the j2objc dist.

PR #158, opened 2015-06-12
Multi-project dependencies

### LINKED ISSUE #51: Multiple java project translation and compilation

### LINKED ISSUE #153: translateClassPaths should be agonostic to lib/ directory
For some reason translateClassPaths assumes libraries must be in the lib/ directory.  This is inconsistent with our other flags, and makes it hard to for example link in jars from dependent projects.  Fixing this along with #51.

advayDev1:
Prereq: #157 

- Users must call finalConfigure() at the end of their j2objcConfig.
- dependsOnJ2objcLib with a project will automatically wire up all dependencies to another j2objc project
- extra* flags allow you to customize the objective-c builds.
- remove lib/ prefix from all translateClassPaths
- disable shared library tasks by default; 50% faster.
- fix logger.warning -> logger.warn
- support using a subset of all iOS platforms

Ugly but it works.  The inability to reconfigure a previously
created native component is an issue others are hitting too:
https://discuss.gradle.org/t/problem-with-model-block-when-switching-from-2-2-1-to-2-4/9937


advayDev1:
All of the functionality in this PR has been tested by my internal set of projects.  They have multiple layers of j2objc project dependencies, and some need custom compiler/ld flags as well.  One also needs to overlay some partially migrated objective-c source.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
```diff
@@ -24,6 +30,12 @@ class J2objcNativeCompilation {
                 if ((task.name =~ /^compile.*Objc$/).matches()) {
                     task.dependsOn 'j2objcTranslate'
                 }
+
+                // Building the shared libraries is for now not useful but also
+                // not possible to configure below, so we disable them dynamically.
+                if ((task.name =~ /^.*SharedLibrary.*$/).matches()) {
+                    task.enabled = false
```
brunobowden: Perfect, much needed change

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
```diff
@@ -24,6 +30,12 @@ class J2objcNativeCompilation {
                 if ((task.name =~ /^compile.*Objc$/).matches()) {
                     task.dependsOn 'j2objcTranslate'
                 }
+
+                // Building the shared libraries is for now not useful but also
```
brunobowden: Reword:

Only static libraries are needed, so disable shared libraries dynamically

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
```diff
@@ -156,6 +168,7 @@ class J2objcNativeCompilation {
                             objc {
                                 source {
                                     srcDirs "${srcGenDir}", "${buildDir}/j2objcHackToForceCompilation"
+                                    srcDirs j2objcConfig.extraObjcSrcDirs
```
brunobowden: How does srcGenDir get compiled? Add comment explaining this.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
```diff
@@ -174,6 +187,12 @@ class J2objcNativeCompilation {
                         targetPlatform 'ios_i386'
                         targetPlatform 'ios_x86_64'
                         targetPlatform 'x86_64'
+
+                        binaries.all {
```
brunobowden: Add comment:

    // Resolve dependencies from all java project using the compiled static libs

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
```diff
@@ -189,6 +208,9 @@ class J2objcNativeCompilation {
                         }
                         binaries.all {
```
brunobowden: Repeat same comment as above

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy
```diff
@@ -81,12 +81,16 @@ class J2objcPlugin implements Plugin<Project> {
...
-            extensions.create('j2objcConfig', J2objcPluginExtension)
+            extensions.create('j2objcConfig', J2objcPluginExtension, project)
             afterEvaluate { evaluatedProject ->
                 // Validate minimally required parameters.
                 // j2objcHome() will throw the appropriate exception internally.
                 assert J2objcUtils.j2objcHome(evaluatedProject)
-                evaluatedProject.j2objcConfig.configureDefaults(evaluatedProject)
+
+                if (!evaluatedProject.j2objcConfig.isNativeBuildConfigured()) {
+                    def message = "You must call configureNativeBuild() in j2objcConfig { }."
```
brunobowden: Present as code that they can copy and paste:

    j2objcConfig {
        configureNativeBuild()
    }

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
```diff
@@ -16,16 +16,22 @@
...
 import org.gradle.api.Project
 import org.gradle.api.tasks.util.PatternSet
 import org.gradle.util.ConfigureUtil
-
 /**
  * Further configuration uses the following fields, setting them in j2objcConfig within build.gradle
  */
 class J2objcPluginExtension {
 
+    final protected Project extendedProject
```
brunobowden: Why call it `extendedProject` and not just `project`?

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
```diff
@@ -163,6 +169,56 @@ class J2objcPluginExtension {
...
+    protected J2objcNativeCompilation nativeCompilation
+    /**
+     * Uses the generated headers and compiled j2objc libraries of the given project when
+     * compiling this project.
+     * <p/>
+     * Generally every cross-project 'compile' dependency should have a corresponding
+     * call to dependsOnJ2objcLib.
+     * <p/>
+     * It is safe to use this in conjunction with --build-closure, however do not duplicate
+     * this project's Java or Objective C source and/or libraries in any of the translation
```
brunobowden: I'm not sure how you would "duplicate this project's Java or ObjC source / libs" in the translation or configuration flags? Perhaps there's a more general principle that code shouldn't be copied around. Always have a reference to the code in the original location. The one exception being the export headers.

Alternatively is there any good way of automatically identifying misconfiguration like this? I understand skipping that if it's just too difficult.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
```diff
@@ -163,6 +169,56 @@ class J2objcPluginExtension {
...
+     * or compilation configuration flags; this method sets those up correctly to not
+     * re-compile beforeProject's source.
+     */
+    // TODO: Do this automatically based on project dependencies.
+    def dependsOnJ2objcLib(Project beforeProject) {
+        // We need to have j2objcConfig on the beforeProject configured first.
+        extendedProject.evaluationDependsOn beforeProject.path
```
brunobowden: Can this be pushed inside the `with` block?

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
```diff
@@ -163,6 +169,56 @@ class J2objcPluginExtension {
...
+                throw new InvalidUserDataException(message)
+            }
+
+            // Build the java/objc libraries and the objc headers of
+            // the other project first.
+            j2objcPreBuild.dependsOn beforeProject.j2objcAssemble
+            // Since we assert the presence of the J2objcPlugin above,
+            // we are guaranteed that the java plugin, which creates the jar task,
+            // is also present.
+            j2objcPreBuild.dependsOn beforeProject.jar
```
brunobowden: Love the direction you're heading in... this plugin is getting really sophisticated. That said, I think it's become complex enough that we need a design doc about how it operates. Once the design settles down for the 1.0 release, I think we should write out the logic so that other contributors coming in can learn how it works.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
```diff
@@ -163,6 +169,56 @@ class J2objcPluginExtension {
...
+
+            logger.debug "$extendedProject:j2objcTranslate must use ${beforeProject.jar.archivePath}"
+            j2objcConfig {
+                translateClassPaths += beforeProject.jar.archivePath
```
brunobowden: IMPORTANT

Can this be inferred automatically from the project dependencies? For example:

    dependencies {
        compile 'my-dependency'
        compile group: 'com.google.guava', name: 'guava', version: '14.0.1'
    }

It would be great if it then automatically added the my-dependency.jav.archivePath as a dependency within the j2objc build. Furthemore, you can automatically add a j2objc dependency like guava... AND also check for differences in library versions.

These dependencies could also be test dependences as well, so they get their own special configuration:

    dependencies {
        testCompile 'junit:junit:4.10'
        testCompile "org.mockito:mockito-core:1.9.5"
    }

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
```diff
@@ -210,7 +266,22 @@ class J2objcPluginExtension {
...
+     */
+    String[] extraObjcSrcDirs = []
+    /**
+     * Additional arguments to pass to the objective-c compiler.
+     */
+    String[] extraObjcCompilerArgs = []
+    /**
+     * Additional arguments to pass to the native linker.
+     */
+    String[] extraLinkerArgs = []
```
brunobowden: I'll do another PR to change "Flags" to "Args" so we remain consistent.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
```diff
@@ -222,17 +293,28 @@ class J2objcPluginExtension {
...
+    protected boolean nativeBuildConfigured = false
+    /**
+     * Configures the native build using.  Must be called at the very
+     * end of your j2objcConfig block.
+     */
+    // TODO: When Gradle makes it possible to modify a native build config
+    // after initial creation, we can remove this, and have methods on this object
+    // mutate the existing native model { } block.  See:
+    // https://discuss.gradle.org/t/problem-with-model-block-when-switching-from-2-2-1-to-2-4/9937
+    def configureNativeBuild() {
```
brunobowden: Should this be a more generic name like `configure()`? In case it's purpose is ever extended in the future to do something in addition to the Native Build Configuration. For an internal method, I'd keep it as is... but since this is a public method, it should be good to avoid changes.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
```diff
@@ -222,17 +293,28 @@ class J2objcPluginExtension {
     String xcodeTarget = null
 
     // Configures defaults whose values are dependent on the project.
-    @PackageScope
-    def configureDefaults(Project project) {
+    protected def configureDefaults() {
```
brunobowden: Since this function is now small enough, I'd just inline it in the constructor.

brunobowden:
@advayDev1, thanks for a lot of positive changes. My apologies is took me a while to get to this - currently I'm on the plane to Boston. I'll be responsive later today and tomorrow to finish this up.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
advayDev1: ah good point.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
advayDev1: If you mean how does it get generated, added comment for that.  Otherwise I'm not sure what you mean 'how it gets compiled'...?

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
advayDev1: done

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
advayDev1: done

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
advayDev1: done

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
advayDev1: Sorry, this comment was clearly poorly worded.  Reworded.

The idea is if you use dependsOnJ2objcLib, you should NOT put the other project's jar or source in translateClassPath or translateSourcepaths respectively.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy
advayDev1: done

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
advayDev1: i thought it was clearer... looking at it now it is just confusingly verbose, fixed.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
advayDev1: yep

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
advayDev1: agreed, let's discuss it later.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
brunobowden: Sorry, I meant how it was generated

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
advayDev1: all of what you said is true and is my understanding of the feature req in #41 (can you comment there on what you'd like to see).

however, it will take me some testing & research to get right and i suspect is a bunch of work (i have an internal plugin that does something similar).  let's do a separate PR.  the instant PR is immediately useful with multiple project builds even if it requires a few more lines.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
brunobowden: dependsOnJ2objcLib is fine for now... but I think it's much better if it reuses the dependency chain from the java plugin. We should discuss later how much we can do with that.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
brunobowden: Maybe make it `finalConfigure()` so that it's clearer it's meant to go at the end of j2objcConfig

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
advayDev1: also testCompile will require getting #112 done.  can you take a crack at that?  after this and #160 are in, I will be absent from the plugin work for a while.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
advayDev1: agreed, see my reply to you below re: reusing the dependencies.

advayDev1:
@brunobowden  - all done.  also added a fix for #177 while I was at it.  Everything retested, including error messages, which sg.

advayDev1:
@brunobowden, also I figured out I think how to automate all the dependencies, however that will still be its own PR, due to the fact that I will put it behind an optional 'experimental' flag as I expect it to have some issues we will only find after using it for a while

advayDev1:
@brunobowden - I think both you and I are blocked on this.  Is this good to go?  Sorry if there has been a miscommunication.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
```diff
@@ -1,17 +1,26 @@
...
-
 /**
  *
  */
 class J2objcNativeCompilation {
-    def apply(Project project, File srcGenDir) {
+
+    static final ALL_SUPPORTED_ARCHS = ['ios_arm64', 'ios_armv7', 'ios_armv7s', 'ios_i386', 'ios_x86_64']
+
+    private final Project project
```
brunobowden: add a separator line to indicate that the variable declaration is distinct from the constructor

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
```diff
@@ -129,6 +144,10 @@ class J2objcNativeCompilation {
                     }
                 }
                 platforms {
+                    x86_64 {
+                        architecture 'x86_64'
+                    }
+                    // This list must match ALL_SUPPORTED_ARCHS.
```
brunobowden: Did you mean:

    // The rest of this list must match ALL_SUPPORTED_ARCHS

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
```diff
@@ -168,12 +186,22 @@ class J2objcNativeCompilation {
...
-                        targetPlatform 'ios_x86_64'
+                        j2objcConfig.supportedArchs.each { arch ->
+                            if (!(arch in ALL_SUPPORTED_ARCHS)) {
+                                def message = "Requested architecture $arch must be one of $ALL_SUPPORTED_ARCHS"
+                                throw new InvalidUserDataException(message)
+                            }
+                            targetPlatform arch
+                        }
+                        // Always need x86_64 for unit-testing on Mac.
                         targetPlatform 'x86_64'
```
brunobowden: So this task is always created... but only has a dependency with the unit test? I was wondering if I limit the architecture to ios_i386.... does it build only that library or also x86_64?

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy
```diff
@@ -81,12 +81,17 @@ class J2objcPlugin implements Plugin<Project> {
                 throw new InvalidUserDataException(message)
             }
 
-            extensions.create('j2objcConfig', J2objcPluginExtension)
+            extensions.create('j2objcConfig', J2objcPluginExtension, project)
```
brunobowden: Don't know why I didn't think of doing that before...

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy
```diff
@@ -81,12 +81,17 @@ class J2objcPlugin implements Plugin<Project> {
...
             afterEvaluate { evaluatedProject ->
                 // Validate minimally required parameters.
                 // j2objcHome() will throw the appropriate exception internally.
                 assert J2objcUtils.j2objcHome(evaluatedProject)
-                evaluatedProject.j2objcConfig.configureDefaults(evaluatedProject)
+
+                if (!evaluatedProject.j2objcConfig.isFinalConfigured()) {
+                    def message = "You must call finalConfigure() in j2objcConfig, ex:\n" +
+                                  "    j2objcConfig { finalConfigure() }"
```
brunobowden: Write out j2objcConfig so it can be copied and pasted. The `...` are useful for indicating where the regular settings go.

    "You must call finalConfigure() in j2objcConfig, ex:\n" +
    "j2objcConfig {\n" +
    "    ..." +
    "    finalConfigure()\n" +
    "}"

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy
```diff
@@ -142,10 +144,14 @@ class J2objcPlugin implements Plugin<Project> {
                 description 'Runs all tests in the generated Objective-C code'
                 testBinaryFile = file("${buildDir}/binaries/testJ2objcExecutable/debug/testJ2objc")
             }
+
+
+
```
brunobowden: I would only have one newline, since this lateDepends is more associated with the previous j2objcTest task than anything else.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
```diff
@@ -16,16 +16,27 @@
...
 import org.gradle.api.Project
 import org.gradle.api.tasks.util.PatternSet
 import org.gradle.util.ConfigureUtil
-
 /**
  * Further configuration uses the following fields, setting them in j2objcConfig within build.gradle
  */
 class J2objcPluginExtension {
 
+    final protected Project project
```
brunobowden: one line gap between declaration and constructor

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
```diff
@@ -16,16 +16,27 @@
...
+    J2objcPluginExtension(Project project) {
+        assert project != null
+        this.project = project
+
+        nativeCompilation = new J2objcNativeCompilation(project)
```
brunobowden: It's a little surprising to see this in simple a configuration object. At minimum it needs a comment on why it's here and why not elsewhere. Is there no other choice?

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
```diff
@@ -155,13 +166,76 @@ class J2objcPluginExtension {
...
+    /**
+     * Uses the generated headers and compiled j2objc libraries of the given project when
+     * compiling this project.
+     * <p/>
+     * Generally every cross-project 'compile' dependency should have a corresponding
+     * call to dependsOnJ2objcLib.
+     * <p/>
+     * It is safe to use this in conjunction with --build-closure.
+     * <p/>
+     * Do not also include beforeProject's java source or jar in the
```
brunobowden: Replace `beforeProject` as that's no longer being used

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcPackLibrariesTask.groovy
```diff
@@ -30,13 +30,9 @@ class J2objcPackLibrariesTask extends DefaultTask {
...
     def getInputLibraries() {
         def staticLibraryPath = "${project.buildDir}/binaries/${project.name}-j2objcStaticLibrary"
-        return project.files([
-                "$staticLibraryPath/ios_arm64$buildType/lib${project.name}-j2objc.a",
-                "$staticLibraryPath/ios_armv7$buildType/lib${project.name}-j2objc.a",
-                "$staticLibraryPath/ios_armv7s$buildType/lib${project.name}-j2objc.a",
-                "$staticLibraryPath/ios_i386$buildType/lib${project.name}-j2objc.a",
-                "$staticLibraryPath/ios_x86_64$buildType/lib${project.name}-j2objc.a",
-        ])
+        return project.files(supportedArchs.collect { arch ->
```
brunobowden: I prefer to write it out as `getSupportedArchs().collect` to make it more clear.

brunobowden:
LGTM - please merge in once you've resolved my recent comments. Been busy at a conference, so it was a little difficult to get to. Ping me on #160 when you feel that's ready for my final review.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
advayDev1: done

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
advayDev1: done

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
advayDev1: depends on the task you run.  yes the task is always created.  remember that while j2objcTest may be disabled, we will still build all architecture flavors for j2objcBuildAllLibraries.  it is not possible to know at this point whether j2objcTest will be enabled/disabled (until after evaluation)

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy
advayDev1: done

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy
advayDev1: done

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
advayDev1: done.  you might want to add this to our formatting XML

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
advayDev1: i think this makes more logical sense than putting the J2objcNC object anywhere else.  J2objcNC acts similar to an extension but we don't want to sets of settings blocks, and J2objcPE basically controls J2objcNC. explained in a comment.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcPackLibrariesTask.groovy
advayDev1: i thought as of #97 we decided we are going with the Groovy style, which means Groovy idioms.  done for now, but I really do think we should stick with only Groovy idioms to avoid an in-between style that requires human thought/discussion about things like this.

advayDev1:
all done, retesting, and will then merge in, and then fixup #160

PR #168, opened 2015-06-12
lateDependsOn => plugins.withType(JavaPlugin)

brunobowden:
- plugins.withType(JavaPlugin) to delay adding dependencies
- “buildAllObjcLibraries” => “j2objcBuildAllObjcLibraries”

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy
@@ -27,6 +27,9 @@ import org.gradle.api.InvalidUserDataException
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.logging.LogLevel
+import org.gradle.api.plugins.BasePlugin
+import org.gradle.api.plugins.JavaPlugin
+
 /*
  * Usage:
  * 1) Download a j2objc release and unzip it to a directory:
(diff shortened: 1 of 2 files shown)


advayDev1:
hmm won't this reintroduce the issue i'm trying to fix with #157 ?

brunobowden:
@confile - I think this is a better way of deferring the dependsOn for the JavaPlugin. From testing, it nolonger depends on the 'java' and 'j2objc' apply ordering. When the 'java' plugin isn't specified, then it correctly fails but it's not such a great error message:

```
Brunos-Air:taptap brunobowden$ ./gradlew clean base:j2objcXcode
Failed to notify ProjectEvaluationListener.afterEvaluate(), but primary configuration failure takes precedence.
org.gradle.api.InvalidUserDataException: j2objc plugin didn't find the 'java' plugin in the 'base' project.
This is a requirement for using j2objc. Please see usage information at:

https://github.com/j2objc-contrib/j2objc-gradle/#usage
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:57)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	...
```

advayDev1:
@brunobowden  - it solves only the plugin ordering issue, but not the issue in #157.  as soon as the java plugin is applied (right after the j2objc plugin), then tasks.getByName will be called, assemble task will be created and we won't be able to add native binaries.

brunobowden:
It defers evaluation until the JavaPlugin is loaded... so the tasks exist
by that point. I tested it by reordering the plugins to load the 'j2objc'
plugin first and then the 'java' plugin and it works fine.

advayDev1:
yes agreed, but we still can't add native binaries after assemble is created.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy
```diff
@@ -27,6 +27,9 @@ import org.gradle.api.InvalidUserDataException
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.logging.LogLevel
+import org.gradle.api.plugins.BasePlugin
```
advayDev1: unused import

brunobowden:
Deferring to @advayDev1 fix but we should still look in further to see if there's a better way to do this.

advayDev1:
Sorry I did not mean you should discard the entire PR.  The clean up of JavaPlugin vs. ours ordering wise still works after my PR #157 (which is merged), and I believe is still a valuable cleanup.  Just revert the part of the PR that touches lateDependsOn.

brunobowden:
Rename to `j2objcBuildAllObjcLibraries` was done in commit 7f53fc7.

Issue #128, opened 2015-06-11
Pods-PROJECT-j2objc-base erroneously added by j2objcXcode


brunobowden:
The "Pods-XCODE_PROJECT" library seems to be created under "Frameworks" and added to the linker flags. This causes the build to fail... when it's removed then the build works fine.

@confile - I'd appreciate your insights on this

![screen shot 2015-06-11 at 3 20 50 am](https://cloud.githubusercontent.com/assets/679339/8104949/9fef4dde-0fe9-11e5-986e-595105a22775.png)

confile:
Not sure maybe it is cauesd by this line: 
https://github.com/j2objc-contrib/j2objc-gradle/blob/master/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcXcodeTask.groovy#L110

What is your ``'${project.name}-j2objc``? I guess it is ``base-j2objc`` from your screenshot. IT is really strange let me test this I give you feedback.

brunobowden:
The `${project.name}-j2objc` is for the library `base-j2objc` which you can see in the xcconfig above and is correct. The problem is the `Pods-PROJECT-j2objc-base` which shouldn't be there.

confile:
I keep looking at this as soon as I get the new plugin running which is currently not the case.

advayDev1:
@confile, the main thing preventing you from building your plugin is this issue and #129.  if you have any potential fixes, this is the issue to look at.

PR #160, opened 2015-06-12
BUGFIX: Fix incremental translations

advayDev1:
Prereqs: #157 #158

Currently if one of your class libs or other dependencies of j2objcTranslate changes BUT none of the main source files does, it will try to re-translate 0 source files. The correct thing to do when a non-source file changes is to retranslate everything.

Fixes #159

Issue #129, opened 2015-06-11
Xcode task copies header files when it should leave them in place


brunobowden:
The "${project.buildDir}j2objcOutputs/src/main/objc" is what Xcode should link to the sources files rather than copying the header files elsewhere. The xcconfig currently is configured as follow: 

HEADER_SEARCH_PATHS = $(inherited) "${PODS_ROOT}/Headers/Public" "${PODS_ROOT}/Headers/Public/j2objc-base"...

Instead it should have a single reference to "${project.buildDir}j2objcOutputs/src/main/objc"

@confile - I'd appreciate your help here

confile:
@brunobowden Should be fixed with this: https://github.com/j2objc-contrib/j2objc-gradle/pull/192

advayDev1:
Current Xcode podspec tasks no longer copy headers.

PR #175, opened 2015-06-14
First Draft of J2Objc Gradle Plugin sample app

confile:
Happy to get comments from you @brunobowden @advayDev1.

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy
@@ -186,14 +186,19 @@ class J2objcPlugin implements Plugin<Project> {
     // The before task must already exist.
     private def lateDependsOn(Project proj, String afterTaskName, String beforeTaskName) {
         assert null != proj.tasks.findByName(beforeTaskName)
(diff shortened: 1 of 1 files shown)


advayDev1:
Awesome.  I'll take a look later today, this is a large PR =)

However, I think we need to fix #128 and #129 before we get this provably working, correct?

confile:
I have a lot of things to improve. I am also not sure about a global build.gradle. Looking forward to your thoughts.

brunobowden:
@confile - let's get the PR's from @advayDev1 merged in first and then make sure this works also. Once we get this in, it will be good to have this working as a continuous build. I can set that up as well.

confile:
@brunobowden Not really know what this meant. Should I drop this request and you create it?

brunobowden:
Here's the ordering on how events should happen:
1) @advayDev1 merge in PR #158 and #160 
2) You rebase your PR to include those changes
3) Fix your your sample app and then contact us again to take another look
4) Review process and fixes
5) Submit PR #175 

@confile - I hope that makes sense

brunobowden:
So, you can keep the PR open for now.

confile:
@brunobowden Yes, perfect. I renamed my master to samples this closes the current pull request. I will create a new one with my sample branch.

Issue #164, opened 2015-06-12
Investigate other default xcode compiler args


advayDev1:
Here is an example compile cmd line from a debug build of a new Xcode 6.1 objc iOS project:
```
[...]

PR #185, opened 2015-06-17
Warn when using -use-arc without -fobjc-arc.

advayDev1:
Issue #164 

Hopefully this would have helped you @confile


src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy
```diff
@@ -95,6 +95,16 @@ class J2objcPlugin implements Plugin<Project> {
                                   "}"
                     throw new InvalidUserDataException(message)
                 }
+
+                if (!('-fobjc-arc' in evaluatedProject.j2objcConfig.extraObjcCompilerArgs) &&
+                    evaluatedProject.j2objcConfig.translateFlags.contains('-use-arc')) {
+                    logger.warn "${evaluatedProject.name}: When translating with -use-arc, it is recommended " +
+                                "to use the '-fobjc-arc' Objective C compiler flag, ex:\n"
```
brunobowden: Extend message to briefly explain what `-fobjc-arc` does

brunobowden:
Look good - see final comment

PR #186, opened 2015-06-17
FAQ.md - improve wording & fix nuances

advayDev1:
Note including build/* in your srcDir may not work as you intend; you need to have task dependencies between the things that generated those build/* files and Java compilation (which is very task-specific, and not something with general advice).  I replaced it with an example that does not rely on such assumptions.

In the future, please include references to authorities - we don't want to become the documentation source-of-truth for much of this info, just a way to solve the easiest instances, and then point people in the right directions for details.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -1,4 +1,4 @@
-# User FAQ
+# FAQ
 (for contributors, see [CONTRIBUTING.md](CONTRIBUTING.md))
 
 ### What version of Gradle do I need?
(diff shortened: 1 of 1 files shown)


advayDev1:
@confile and @brunobowden , ptal

FAQ.md
```diff
@@ -1,19 +1,23 @@
-# FAQ
+# User FAQ
```
brunobowden: I'd just keep it as `FAQ` rather than `User FAQ`. The former is ten times more common than the latter in Google search results.

brunobowden:
Looks good - see single comment

FAQ.md
confile: Yes only FAQ not User FAQ

advayDev1: yeah makes sense

PR #188, opened 2015-06-17
README.md - include finalConfigure() instructions

advayDev1:


confile:
LGTM. 

Isn't there a more elegant solution than forcing the user to call a function in the extension block?

README.md
```diff
@@ -37,6 +37,11 @@ Tracked issue: https://github.com/j2objc-contrib/j2objc-gradle/issues/130**
         xcodeProjectDir "${projectDir}/../ios"   // Xcode workspace directory
         xcodeTarget "MyTarget"                   // Xcode application target name
         // More: https://github.com/j2objc-contrib/j2objc-gradle/blob/master/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy#L25
+        
+        // Must always be the last statement in the j2objcConfig block.
+        // You must include this call even if you do not want to configure
```
brunobowden: I'd combine the last two lines of comments in to a single line:

       // You must include this call even if you do not configure any other settings

brunobowden:
LGTM - see single comment

README.md
advayDev1: added, slightly different wording to ensure users keep it at the end.

advayDev1:
@confile, not at the moment, see #158 comment:
"The inability to reconfigure a previously
created native component is an issue others are hitting too:
https://discuss.gradle.org/t/problem-with-model-block-when-switching-from-2-2-1-to-2-4/9937"
:(

PR #190, opened 2015-06-17
Added docs for extraObjcCompilerArgs

confile:


README.md
```diff
@@ -36,6 +36,10 @@ Tracked issue: https://github.com/j2objc-contrib/j2objc-gradle/issues/130**
     j2objcConfig {
         xcodeProjectDir "${projectDir}/../ios"   // Xcode workspace directory
         xcodeTarget "MyTarget"                   // Xcode application target name
+        
+        // make ARC avalable for the objective-c compiler
+        extraObjcCompilerArgs += '-fobjc-arc'
```
brunobowden: Having this in the default example is basically suggesting making ARC the default - do we want to do this?.

I'm wondering if we should comment this out, so people don't enable unless they clearly know that they need it.

brunobowden:
@advayDev1 - interested in your thoughts on this

advayDev1:
my view is consistency.  since the default J2objc does not have -use-arc, i don't see why the default compile would have -fobjc-arc.

confile:
Done

confile:
Swift uses ARC per default so we should consider doing this as well.

brunobowden:
The J2ObjC team recommeend NOT using ARC. If you do a search for
"j2objc-discuss" and ARC, you'll find the following from late last year:

Keith Stanger:
"My opinion on this may be different than Tom's, but I always advise users
not to use ARC with translated code."

https://groups.google.com/forum/#!topic/j2objc-discuss/sGJHQgao788

confile:
Okay lets keep it. It is easy to set the flag. Lets focus on the other issues.

README.md
```diff
@@ -36,6 +36,10 @@ Tracked issue: https://github.com/j2objc-contrib/j2objc-gradle/issues/130**
     j2objcConfig {
         xcodeProjectDir "${projectDir}/../ios"   // Xcode workspace directory
         xcodeTarget "MyTarget"                   // Xcode application target name
+        
+        // make ARC avalable for the objective-c compiler
+        // extraObjcCompilerArgs += '-fobjc-arc'
```
brunobowden: I'd remove this from the README.md and have it solely in the FAQ.

brunobowden:
@tomball, @kstanger - looking at the J2ObjC docs, it appears that you now recommend using ARC for new projects (this is different to what Keith suggested on j2objc-discuss in the past). Accordingly, we're thinking of suggesting ARC for the j2objc-gradle plugin.

Will this become a default for the j2objc? This is important to us as the plugin has the convention of defaulting to the j2objc command line flag defaults.

Current documentation:
http://j2objc.org/docs/Memory-Management.html#automatic-reference-counting-arc

@confile - let's hold off until we get Tom or Keith's insights.

confile:
@brunobowden Agreed!

kstanger:
Note that j2objc's -use-arc and clang's -fobjc-arc go hand-in-hand. If you use one without the other compilation will fail.

My opinion remains that translated code is best without ARC.

brunobowden:
Understood Keth. My concern is that should have consistency between the
documentation, j2objc command line, j2objc-discuss and the gralde plugin.
It's quite alright to present alternatives but we should collectively agree
on the default.

advayDev1:
The issue @kstanger mentioned is already fired as a warning by implementation #185 but perhaps it should become an error.

README.md
```diff
@@ -36,6 +36,8 @@ Tracked issue: https://github.com/j2objc-contrib/j2objc-gradle/issues/130**
     j2objcConfig {
         xcodeProjectDir "${projectDir}/../ios"   // Xcode workspace directory
         xcodeTarget "MyTarget"                   // Xcode application target name
```
brunobowden: Match other documentation:

    xcodeTarget "IosApp"                   // Xcode application target name

README.md
```diff
@@ -36,6 +36,8 @@ Tracked issue: https://github.com/j2objc-contrib/j2objc-gradle/issues/130**
...
+
```
brunobowden: Only one blank line rather than two

README.md
```diff
@@ -36,6 +36,8 @@ Tracked issue: https://github.com/j2objc-contrib/j2objc-gradle/issues/130**
     j2objcConfig {
         xcodeProjectDir "${projectDir}/../ios"   // Xcode workspace directory
         xcodeTarget "MyTarget"                   // Xcode application target name
+    
+
         // More: https://github.com/j2objc-contrib/j2objc-gradle/blob/master/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy#L25
```
brunobowden: Amend doc link to:

```
        // Documentation on further configuration:
        // https://github.com/j2objc-contrib/j2objc-gradle/blob/master/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy#L25
```

FAQ.md
```diff
@@ -18,3 +18,18 @@ sourceSets {
...
+### How do I enable ARC for my objective-c classes?
+
+Add the following to your configuration block.
+
+```
+j2objcConfig {
+   extraObjcCompilerArgs += '-fobjc-arc'
+}
+```
+
```
advayDev1: Add a reference to https://developer.apple.com/library/mac/releasenotes/ObjectiveC/RN-TransitioningToARC/Introduction/Introduction.html#//apple_ref/doc/uid/TP40011226-CH1-SW15

(and keep the blank line too)

advayDev1:
LGTM from me, after doing the above commits (+squash/rebase)

FAQ.md
confile: done

README.md
confile: What is your point here?

README.md
confile: done

README.md
confile: don't get it?

confile:
@advayDev1 done

FAQ.md
```diff
@@ -22,3 +22,18 @@ sourceSets {
...
 ```
+
+
+### How do I enable ARC for my objective-c classes?
+
+Add the following to your configuration block. [See](https://developer.apple.com/library/mac/releasenotes/ObjectiveC/RN-TransitioningToARC/Introduction/Introduction.html#//apple_ref/doc/uid/TP40011226-CH1-SW15).
+
+```
+j2objcConfig {
+   extraObjcCompilerArgs += '-fobjc-arc'
```
brunobowden: Include the translateFlags line as well

README.md
brunobowden: To match the documentation in README.md

confile: Still don'T get what I should change!??!

brunobowden: This is a comment line 39. Change that to new code that I gave. Note how "MyTarget" becomes "IosApp" in the comment, repeated here:

    xcodeTarget "IosApp"                   // Xcode application target name

confile: done

confile:
done

FAQ.md
```diff
@@ -22,3 +22,19 @@ sourceSets {
...
+   translateFlags '-use-arc'
```
brunobowden: Swap two lines to match execution order... putting the translate before the compile args. This also matches the code error.

README.md
```diff
@@ -36,7 +36,8 @@ Tracked issue: https://github.com/j2objc-contrib/j2objc-gradle/issues/130**
     // Plugin settings:
     j2objcConfig {
         xcodeProjectDir "${projectDir}/../ios"   // Xcode workspace directory
-        xcodeTarget "MyTarget"                   // Xcode application target name
+        xcodeTarget "IosApp"                   // Xcode application target name
```
brunobowden: NIT: indent the comment two spaces so it matches the line above.

FAQ.md
confile: done

FAQ.md
confile: done

confile:
done

PR #192, opened 2015-06-17
Fixed xcode task does not compile in Xcode fixes #128

### LINKED ISSUE #128: Pods-PROJECT-j2objc-base erroneously added by j2objcXcode

confile:


confile:
TESTED=Yes

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcXcodeTask.groovy
```diff
@@ -102,11 +102,11 @@ class J2objcXcodeTask extends DefaultTask {
                 "s.name = '${getPodName()}'\n" +
                 "s.version = '1.0'\n" +
                 "s.summary = 'Generated by the j2objc-gradle plugin.'\n" +
-                "s.source_files = '${srcGenDirRelativeToBuildDir}**/*.{h,m}'\n" +
+                "s.source_files = '${srcGenDirRelativeToBuildDir}**/*.{h}'\n" +
```
brunobowden: Any particular reason to remove the `.m` files? If you only import the header files, is that you can't step through the source during debugging.

confile: Yes, I looked at diffent samples using static libs in pods they all does not use m in sources. The problem is when you use .m in sources everything gets imported twice.

confile:
@brunobowden As you can see here it is impossible to browse source code when using a static lib: 

https://github.com/CocoaPods/CocoaPods/issues/3700#issuecomment-113104921

brunobowden:
@advayDev1 - please comment on this

I had Xcode stepping through source code of the statically linked library myself. I wasn't able to set breakpoints (which would've been ideal) but if I hit a breakpoint in my Swift code, it would step in to the static lib.

confile:
@brunobowden Does the current Xcode task compiles with you in Xcode?

advayDev1:
I have an xcode project where I had to get rid of all the referenced folders in the xcode project, but of course kept the user header search path the same.  With such a project, I am still able to debug fully (breakpoints, symbols, stacks, step-into).  I think Ctrl-Click was also working for me.

I think we should be fine getting rid of all the files from the xcode project, and thus would assume but don't know for certain, get rid of all files in the pod as well.

confile:
@brunobowden @advayDev1 As I pointed out before this is removing the ``.m`` files from the pod is the only way to get the pod working. I posed the link from a discussion I had with the pod creators.

advayDev1:
@confile - that's not i interpreted the link you gave.  the alternative i'm suggesting is including no source files, .m or .h.  do pods support that?

confile:
@advayDev1 Yes you can delete this line: 

```
s.source_files = 'j2objcSrcGen/**/*.{h}'
```

But then you do not have an overview of the headerfiles in Xcode. 

@advayDev1 Your last comment was unclear. 

1. Did you use pods?
2. How do you browse the code of a static lib? With ``cmd + click`` you can only see the headers.

advayDev1:
No i never use pods which is why it is very difficult for me to help on these bugs/PRs :)  It is pretty much guess and check.

You need to find a setting/directive in pods that lets you add /referenced/ files as opposed to /source/ files.  You can put headers and source there, and everything will work.

confile:
@advayDev1 Please see this discussion. There is no way to do this with pod and static libs.
https://github.com/CocoaPods/CocoaPods/issues/3700#issuecomment-113104921

advayDev1:
@confile, I explained the issue in detail here: https://github.com/CocoaPods/CocoaPods/issues/3700#issuecomment-113302219

confile:
@advayDev1 I think we are done here. Arn't we?

advayDev1:
yes we're done.  summary of thread on CocoaPods:
- this setup (headers only) is fine to get auto-complete
- cocoapods does not and will not support 'referenced files', needed to allow easy breakpoints in the source itself (you can still open the file and set a breakpoint; debugging works regardless of the presence of the file in or referenced by the xcode project)

please do confirm that this will not however /copy/ the headers from j2objcSrcGen to another place.

if my understanding above is correct, LGTM from me.

confile:
@advayDev1 It does not copy the header it links just the file. So do I get an LGTM on this pull request?


It is said that I have to say it but I thing the old j2objc plugin had an advantage using pods while taking full source code instead of static libs. We lost the chance to browse code. But on the other hand we gain compile time since the static lib do not have to be compiled after Xcode clean I guess.

advayDev1:
i'd like you to get @brunobowden 's LG as well if he has any concerns with the above summary of the state of things.

yes there are tradeoffs.  i think i've been always thinking about things in terms of scaling to large projects, and often those needs are in conflict with earlier-stage or smaller projects.  one of the things i find important for large projects is streamlined/automateable processes, and compiling things properly in gradle for all architectures and such i think is an important part of that.

confile:
@brunobowden please take a look!

advayDev1:
Also we should remember that we could manipulate the Xcode project file with our own java code to support browsing of source files (by adding 'referenced files').  I don't think we have to give up on it just b/c cocoapods doesn't want to support this.  This is one of the issues using an intermediary between Gradle and Xcode, we are the mercy of that intermediary.

confile:
@advayDev1 Please open a new Issue where we could discuss this. God point by the way.

brunobowden:
LGTM - I would prefer a better solution but it's fine for now

confile:
I would also prefer a better solution.

Issue #98, opened 2015-06-04
Unit Testing for Plugin


brunobowden:
This is going to be needed for more external contributors.

@advayDev1 - please set up initial example test, then reassign to me and I'll build out the tests for all the other tasks

advayDev1:
I will not get a chance to do this

brunobowden:
There isn't as much of a rush on this. I hope to work on this later next
week.

brunobowden:
Marking as fixed as the unit testing framework is in place and covers a consider amount of functionality. I'll add something to the style guidelines about requiring tests for future changes.

PR #198, opened 2015-06-20
J2objcUtils: type checking + unit tests

brunobowden:
Changes isolated to J2objcUtils, J2objcUtilsTest and test resource
- First unit tests for the plugin
- TODO’s for methods not unit tested
- Added prefixes.properties as test resource
- Replace many instances of “def” with explicit type
- Remove unused fileFilter(…) method
- Introduces unit tests - issue #98

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcUtils.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcUtils.groovy
@@ -22,6 +22,7 @@ import org.gradle.api.Project
 import org.gradle.api.file.FileCollection
 import org.gradle.api.file.SourceDirectorySet
 import org.gradle.api.tasks.SourceSet
+import java.util.regex.Matcher
 
 /**
  * Internal utilities supporting plugin implementation.
@@ -33,8 +34,8 @@ class J2objcUtils {
     // TODO: ideally bundle j2objc binaries with plugin jar and load at runtime with
     // TODO: ClassLoader.getResourceAsStream(), extract, chmod and then execute
 
-    static isWindows() {
-        return System.getProperty("os.name").toLowerCase().contains("windows")
+    static boolean isWindows() {
+        return System.getProperty('os.name').toLowerCase().contains('windows')
     }
 
     // Retrieves the configured source directories from the Java plugin SourceSets.
@@ -47,8 +48,8 @@ class J2objcUtils {
         return sourceSet.getProperty(fileType)
     }
 
-    static def sourcepathJava(Project proj) {
-        def javaRoots = []
+    static String sourcepathJava(Project proj) {
+        String[] javaRoots = []
         srcDirs(proj, 'main', 'java').srcDirs.each {
             javaRoots += it.path
         }
@@ -60,8 +61,8 @@ class J2objcUtils {
 
     // MUST be used only in @Input getJ2ObjCHome() methods to ensure UP-TO-DATE checks are correct
     // @Input getJ2ObjCHome() method can be used freely inside the task action
-    static def j2objcHome(Project proj) {
-        def localPropertiesFile = new File(proj.rootDir, 'local.properties')
+    static String j2objcHome(Project proj) {
+        File localPropertiesFile = new File(proj.rootDir, 'local.properties')
         String result = null
         if (localPropertiesFile.exists()) {
             Properties localProperties = new Properties();
@@ -74,7 +75,7 @@ class J2objcUtils {
             result = System.getenv('J2OBJC_HOME')
         }
         if (result == null) {
-            def message =
+            String message =
                     "j2objc home not set, this should be configured either:\n" +
                     "1) in a 'local.properties' file in the project root directory as:\n" +
                     "   j2objc.home=/PATH/TO/J2OBJC/DISTRIBUTION\n" +
@@ -87,38 +88,30 @@ class J2objcUtils {
             throw new InvalidUserDataException(message)
         }
         if (!proj.file(result).exists()) {
-            def message = "j2objc directory not found, expected location: ${result}"
+            String message = "j2objc directory not found, expected location: ${result}"
             throw new InvalidUserDataException(message)
         }
         return result
     }
 
-    // Filters a FileCollection by path:
-    // must match includeRegex and NOT match excludeRegex, regex ignored if null
-    static def fileFilter(FileCollection files, String includeRegex, String excludeRegex) {
-        return files.filter { file ->
-            return file.path.matches(includeRegex)
-        }.filter { file ->
-            return !file.path.matches(excludeRegex)
-        }
-    }
-
-    // Reads properties file and flags from translateFlags (last flag takes precedence)
+    // Reads properties file and arguments from translateArgs (last argument takes precedence)
     //   --prefixes dir/prefixes.properties --prefix com.ex.dir=Short --prefix com.ex.dir2=Short2
-    // TODO: separate this out to a distinct flag that's added to translateFlags
+    // TODO: separate this out to a distinct argument that's added to translateArgs
     // TODO: @InputFile conversion for this
-    static def prefixProperties(Project proj, String translateFlags) {
+    static Properties prefixProperties(Project proj, List<String> translateArgs) {
         Properties props = new Properties()
-        def matcher = (translateFlags =~ /--prefix(|es)\s+(\S+)/)
-        def start = 0
+        String joinedTranslateArgs = translateArgs.join(' ')
+        Matcher matcher = (joinedTranslateArgs =~ /--prefix(|es)\s+(\S+)/)
+        int start = 0
         while (matcher.find(start)) {
             start = matcher.end()
-            def newProps = new Properties()
-            def argValue = matcher.group(2)
+            Properties newProps = new Properties()
+            String argValue = matcher.group(2)
             if (matcher.group(1) == "es") {
                 // --prefixes prefixes.properties
                 // trailing space confuses FileInputStream
-                def prefixesPath = argValue.trim()
+                String prefixesPath = argValue.trim()
+                log.debug "Loading prefixesPath: $prefixesPath"
                 newProps.load(new FileInputStream(proj.file(prefixesPath).path))
             } else {
                 // --prefix com.example.dir=CED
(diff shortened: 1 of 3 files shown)


brunobowden:
@advayDev1, @confile - please take a look

This is the first beachhead for adding unit testing to the plugin.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcUtilsTest.groovy
```diff
@@ -0,0 +1,164 @@
...
+    // TODO: testJ2objcHome()
+
+    @Test
+    public void testPrefixProperties_FileOnly() {
+        // TODO: fix to use this.getClass().getResource(...) once Android Studio issue fixed
+        // https://code.google.com/p/android/issues/detail?id=75991
+        File prefixesProp = new File(
+                'src/test/resources/com/github/j2objccontrib/j2objcgradle/prefixes.properties')
+
+        List<String> translateArgs = new ArrayList<String>()
```
confile: List<String> translateArgs = new ArrayList<>()

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcUtils.groovy
```diff
@@ -74,7 +75,7 @@ class J2objcUtils {
             result = System.getenv('J2OBJC_HOME')
         }
         if (result == null) {
-            def message =
+            String message =
```
advayDev1: do we really need to remove 'def' in these simple cases? seems like standard groovy.

I understand many of the other cases, it makes the code much clearer.  but using 'def' in these cases is purely shorthand and I think makes the code easier to understand (because it is easier to skim).  often times people wouldn't even use variables for such strings and have them directly in the exception constructor.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcUtils.groovy
```diff
@@ -126,19 +119,24 @@ class J2objcUtils {
             }
             props.putAll(newProps)
         }
-//        for (key in props.keys()) {
-//            log.debug key + ": " + props.getProperty(key)
-//        }
+        for (key in props.keys()) {
+            log.debug key + ": " + props.getProperty(key)
```
advayDev1: There's no context for this debug message.  Perhaps add a log.debug before and after the for loop that indicates what this is, end where it ends.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcUtils.groovy
```diff
@@ -155,22 +153,22 @@ class J2objcUtils {
...
             }
         }
         return files
     }
 
-    static def absolutePathOrEmpty(Project proj, List<String> relativePaths) {
+    static String absolutePathOrEmpty(Project proj, List<String> relativePaths) {
         if (relativePaths.size() > 0) {
-            def tmpPaths = ""
+            String tmpPaths = ''
```
advayDev1: why not just join use something similar to:
relativePaths.collect { proj.file(it).path }.join(':')

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcUtilsTest.groovy
```diff
@@ -0,0 +1,164 @@
...
+import org.gradle.api.file.ConfigurableFileCollection
+import org.gradle.api.file.FileCollection
+import org.gradle.api.file.SourceDirectorySet
+import org.gradle.testfixtures.ProjectBuilder
+import org.junit.After
+import org.junit.Before
+import org.junit.Test;
+
+/**
+ * J2objcPluginExtension tests.
```
advayDev1: These are tests for J2objcUtils not PE

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcUtilsTest.groovy
```diff
@@ -0,0 +1,164 @@
...
+ */
+public class J2objcUtilsTest {
+
+    final File testDir = new File('build/tmp/tests')
+    private Project proj
+
+    @Before
+    void setUp() {
+        proj = ProjectBuilder.builder().withProjectDir(testDir).build()
```
advayDev1: hmm... this makes a strong assumption that the working directory is the one gradle will use.

why not just let ProjectBuilder use its own default tmp directory?

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcUtilsTest.groovy
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    @After
+    void tearDown() {
+        if (testDir.exists()) {
```
advayDev1: if you use ProjectBuilder's default directory, this manual cleanup should not be needed.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcUtilsTest.groovy
```diff
@@ -0,0 +1,164 @@
...
+        assert count == 15
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testMatchNumberRegex_NoMatch() {
+        int count = J2objcUtils.matchNumberRegex("AA CYCLES FOUND", /(\d+) CYCLES FOUND/)
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testMatchNumberRegex_notNumber() {
```
advayDev1: use consistent naming scheme.  either the suffix after the underscore should be initial uppercase or not (compare NoMatch)

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcUtils.groovy
brunobowden: The style guide lists it as a choice but asks you to explicit about choice:
http://www.groovy-lang.org/style-guide.html#_def_and_type

Though def is more common in Groovy, I think it's better for correctness and compile time checking to be explicit on types. It's perhaps easier to be fast and loose with def but I think for a plugin like this, it should steer more towards correctness than dev speed.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcUtils.groovy
brunobowden: Improved debug messages and linked to documentation:

        log.debug 'Package Prefixes: http://j2objc.org/docs/Package-Prefixes.html'
        for (key in props.keys()) {
            log.debug "Package Prefix Property: $key : ${props.getProperty((String) key)}"
        }

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcUtils.groovy
advayDev1: i think what the style guide is saying is don't do `def String message`.  it isn't saying you have to use only def or only String throughout.

I'm fine with this if we are consistent - i.e. we say in our style guide, never use def.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcUtils.groovy
advayDev1: use stringPropertyNames() instead of keys() for strong-typed generics.

also wouldn't the stance on strong-typing then require it to read:
```
for (String key in props.stringPropertyNames()) {
```

the auto-type here is essentially the same as a `def`

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcUtils.groovy
brunobowden: A definite improvement but with unit tests, I see that something else is needed. Perhaps a single ':' by itself it ok since it just creates an empty location in the path... but I'll add it as a TODO for now.

        if (relativePaths.size() == 0) {
            return ''
        } else {
            return ':' + relativePaths.collect({ proj.file(it).path }).join(':')
        }

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcUtilsTest.groovy
brunobowden: fixed

advayDev1:
LGTM with above comments considered

advayDev1:
also please add the issue number for unit-tests to the commit message so they are linke

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcUtilsTest.groovy
brunobowden: fixed

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcUtilsTest.groovy
brunobowden: Fixed the single lowercase test name

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcUtilsTest.groovy
brunobowden: fixed

brunobowden:
Add reference to unit testing issue #98 in commit message

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcUtils.groovy
```diff
@@ -87,38 +88,30 @@ class J2objcUtils {
...
-    }
-
-    // Reads properties file and flags from translateFlags (last flag takes precedence)
+    // Reads properties file and arguments from translateArgs (last argument takes precedence)
     //   --prefixes dir/prefixes.properties --prefix com.ex.dir=Short --prefix com.ex.dir2=Short2
-    // TODO: separate this out to a distinct flag that's added to translateFlags
+    // TODO: separate this out to a distinct argument that's added to translateArgs
     // TODO: @InputFile conversion for this
-    static def prefixProperties(Project proj, String translateFlags) {
+    static Properties prefixProperties(Project proj, List<String> translateArgs) {
```
confile: @brunobowden The import of Properties is missing.

confile: import java.util.Properties

brunobowden: This is a little peculiar to me. When I add that line, it shows in the Android Studio IDE as `unused import`. I'm not quite sure what Gradle / Groovy magic is going on here. Is it an automatic import or something else?

So, clearly something is wrong here but I'm not sure how to fix it in a way that works in the IDE:

screen shot 2015-06-21 at 9 44 20 am

confile: I use Eclipse but it does not show anything. Better keep an Unused import than a missing import somewhere else.

advayDev1: http://www.groovy-lang.org/differences.html#_default_imports

No import should be added

PR #201, opened 2015-06-20
Fix `plugins {...}` syntax by allowing reordering of plugin loading

### LINKED ISSUE #130: fix plugins {} syntax - java plugin ordering

brunobowden:
- Creates new `throwIfNoJavaPlugin(…)` method
- Move the above check from `project.with` to `project.afterEvaluate`
- Add another check in `srcDirs(…)` method as a double check

This allows the `j2objc` to be loaded before the `java` plugin… but
still produces meaningful errors if the `java` plugin is missing. This
allows the use of the `plugins` syntax and should fix #130.

This will need to be verified after a new plugin is published. Likely
version 0.3.0.

    plugins {
        id "com.github.j2objccontrib.j2objcgradle" version "0.3.0-alpha"
    }

    apply 'java'

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,9 +38,10 @@ Tracked issue: https://github.com/j2objc-contrib/j2objc-gradle/issues/130**
         xcodeProjectDir "${projectDir}/../ios"   // Xcode workspace directory
         xcodeTarget "IosApp"                     // Xcode application target name
     
-        // More: https://github.com/j2objc-contrib/j2objc-gradle/blob/master/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy#L25
+        // Help information on other settings:
+        // https://github.com/j2objc-contrib/j2objc-gradle/blob/master/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy#L25
         
-        // You must include this call (at the end of the block) even if you do not configure any other settings.
+        // You must include this call (at the end of j2objcConfig) regardless of settings
         finalConfigure()
     }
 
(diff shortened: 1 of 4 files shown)


brunobowden:
@advayDev1, @confile - This is an important change that *should* allow use of the `plugins {...}` syntax. I've tested a number of error scenarios, e.g. the plugin not loaded and it gives an intelligent error message.

Once this change is merged in, I'll spin a new version and verify that it fixes #130. After that, I'll work through some other bugs, including any issues that @confile is seeing. Then do another version and finally promote that in a few places.

confile:
@brunobowden Sound great. I just have a look at it. Sorry for raising so many error but I still have a lot of problems with the new plugin.

confile:
LGTM

brunobowden:
@confile - making you happy is a good path to making many other people
happy as well

confile:
@brunobowden Thanks!

advayDev1:
LGTM.  There's some odd smart-quotes thing going in your comment.  Hopefully that isn't in the code too:
    “0.3.0-alpha"

brunobowden:
Fixed weird smart quotes - probably from using the GitHub client, though I still find it useful for visual diffs. Added unit test for throw without Java plugin and TODO for test for no throw when Java plugin present.

brunobowden:
Fixes #130 

... I'll try being confident ;-) Verification later tonight.

PR #206, opened 2015-06-21
def => explicit types fixes #205

### LINKED ISSUE #205: def => explicit types

    // Before
    def count = 4
    def method() {
        return "string"
    }

    // After
    int count = 4
    String method() {
        return "string"
    }

    declared returning def).

confile:


src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcPackLibrariesTask.groovy
```diff
@@ -29,9 +31,10 @@ import org.gradle.api.tasks.TaskAction
 class J2objcPackLibrariesTask extends DefaultTask {
 
     // Generated ObjC binaries
+    // TODO Bruno unused in this task may be remove it
```
brunobowden: Note the @InputFiles annotation. So if any of the input files change, it means that the task needs to be run again. So it's intentional designed this way. So you don't need the comment.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcTranslateTask.groovy
```diff
@@ -132,12 +134,12 @@ class J2objcTranslateTask extends DefaultTask {
                     logger.debug "New or Updated non-source file: " + change.file
                 }
             }
-            def removedFileNames = []
+            List<String> removedFileNames = new ArrayList<>()
```
brunobowden: This is a more important change, effectively `Object[]` => `List<String>`. So you need to carefully check that the associated methods still exist. Does `+=`, `size()` and `contains()` all exist for `List<String>`, or do they change?

Much safe for now to change to `String[]` instead.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcXcodeTask.groovy
```diff
@@ -177,7 +180,7 @@ class J2objcXcodeTask extends DefaultTask {
 
     // checks if a pod is still integrated into a pod file
     // return: if podspec exists and the target line for insertion
-    static def checkPodDefExistence(File podFile, String xcodeTarget, String podName) {
+    static List<Object> checkPodDefExistence(File podFile, String xcodeTarget, String podName) {
```
brunobowden: Changing from an array to list, is a bigger change. I think if you instead change it to `Object[]`, then you'll be able to minimize the other changes... which means minimizes the risk. Always something important for a big change like this.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcXcodeTask.groovy
```diff
@@ -132,9 +133,11 @@ class J2objcXcodeTask extends DefaultTask {
             logger.debug "Pod exists at path: ${getXcodeProjectDir()}"
             // check if this podspec has been included before
             // (a, b) = f() syntax unpacks the values in the return tuple
-            def (boolean podIntegrationExists, String targetPodLine) =
+            List<Object> checkPodDefResult =
             checkPodDefExistence(podFile, xcodeTarget, getPodName())
-
+            boolean podIntegrationExists = checkPodDefResult[0]
+            String targetPodLine = checkPodDefResult[0]
```
brunobowden: See the bug here? Index should be `1` instead of `0`. This is the challenge with making a wide change.

If you use the type `Object[]`, are you able to reduce these changes?

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcXcodeTask.groovy
```diff
@@ -177,7 +180,7 @@ class J2objcXcodeTask extends DefaultTask {
 
     // checks if a pod is still integrated into a pod file
     // return: if podspec exists and the target line for insertion
-    static def checkPodDefExistence(File podFile, String xcodeTarget, String podName) {
```
brunobowden: Since this is a static function, this is relatively straight forward to test. Also the unit test makes it much faster to try out type changes and verify that it works correctly. Have a look at J2objcPluginExtensionTest.groovy, copy that file and modify it, then you can create another test for this class, maybe only this method to start with. You can do that in a separate commit. There is a prefixes.properties file in `test/resources` that you can put example Podfile's.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcPackLibrariesTask.groovy
confile: okay deleted

brunobowden:
Revert a few of the `List<...>` changes back to closer to their original type of `Object[]` or `String[]`.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcTranslateTask.groovy
confile: @brunobowden Isn't String[] an array of Strings with fixed length? So you want: 

String[] removedFileNames = []

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcXcodeTask.groovy
confile: Not sure why Object[] does a change. 

The fix is: String targetPodLine = checkPodDefResult[0]

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcXcodeTask.groovy
confile: fixed

confile:
Done

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcTranslateTask.groovy
advayDev1: This is not the case.  'def' literally means Object, and [] is new ArrayList().  So the old line was

    Object removedFileNames = new ArrayList();

The new line is correct.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcXcodeTask.groovy
advayDev1: I'm confused by these comments regarding changing arrays to lists.  def is Object, not Object[].
http://juixe.com/techknow/index.php/2006/06/11/common-groovy-errors/

advayDev1:
IMPORTANT note re: Lists, def, and arrays.
def is Object, not Object[].
Also [BLAH] is an ArrayList initializer:
http://juixe.com/techknow/index.php/2006/06/11/common-groovy-errors/

They were never arrays, unless we did .toArray or 'as String[]' or something like that.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcXcodeTask.groovy
advayDev1: see comment below

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcXcodeTask.groovy
advayDev1: See below re: arrays/lists.

advayDev1:
Also @confile, you should consider making sure all your commits are under your github.com-registered email and/or that the user email you use for your local git configuration is also registered under your Github profile.  That will cause all of your commits to appear from 'you'.  Right now some of your commits appear as linked to @confile, and the others are unknown to Github:

screen shot 2015-06-21 at 3 42 59 pm

confile:
@advayDev1 Okay I check the config. Also I will revert my last changes.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy
```diff
@@ -28,11 +28,13 @@ import org.gradle.api.InvalidUserDataException
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.logging.LogLevel
+import java.io.File
```
advayDev1: don't add this import: http://www.groovy-lang.org/structure.html#_default_imports

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcCycleFinderTask.groovy
```diff
@@ -19,11 +19,13 @@ package com.github.j2objccontrib.j2objcgradle.tasks
 import org.gradle.api.DefaultTask
 import org.gradle.api.InvalidUserDataException
 import org.gradle.api.file.FileCollection
+import org.gradle.api.file.SourceDirectorySet
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.InputFiles
 import org.gradle.api.tasks.Optional
 import org.gradle.api.tasks.OutputFile
 import org.gradle.api.tasks.TaskAction
+import java.io.ByteArrayOutputStream
```
advayDev1: don't add this import: http://www.groovy-lang.org/structure.html#_default_imports

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcXcodeTask.groovy
```diff
@@ -23,6 +23,7 @@ import org.gradle.api.tasks.InputDirectory
 import org.gradle.api.tasks.OutputFile
 import org.gradle.api.tasks.TaskAction
 import org.gradle.api.tasks.incremental.IncrementalTaskInputs
+import java.io.ByteArrayOutputStream
```
advayDev1: don't add this import: http://www.groovy-lang.org/structure.html#_default_imports

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcTranslateTask.groovy
```diff
@@ -18,12 +18,14 @@ package com.github.j2objccontrib.j2objcgradle.tasks
...
 import org.gradle.api.DefaultTask
 import org.gradle.api.file.FileCollection
+import org.gradle.api.file.SourceDirectorySet
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.InputFiles
 import org.gradle.api.tasks.Optional
 import org.gradle.api.tasks.OutputDirectory
 import org.gradle.api.tasks.TaskAction
 import org.gradle.api.tasks.incremental.IncrementalTaskInputs
+import java.io.ByteArrayOutputStream
```
advayDev1: don't add this import: http://www.groovy-lang.org/structure.html#_default_imports

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcTestTask.groovy
```diff
@@ -172,25 +176,25 @@ class J2objcTestTask extends DefaultTask {
     // Generate list of tests from the source java files
     // e.g. src/test/java/com/example/dir/ClassTest.java => "com.example.dir.ClassTest"
     // depends on --prefixes dir/prefixes.properties in translateArgs
-    def static testNames(Project proj, FileCollection srcFiles, List<String> translateArgs) {
-        def prefixesProperties = J2objcUtils.prefixProperties(proj, translateArgs)
+    FileCollection static testNames(Project proj, FileCollection srcFiles, List<String> translateArgs) {
```
advayDev1: static should come before the type

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcTestTask.groovy
```diff
@@ -19,11 +19,15 @@ package com.github.j2objccontrib.j2objcgradle.tasks
 import org.gradle.api.DefaultTask
 import org.gradle.api.Project
 import org.gradle.api.file.FileCollection
+import org.gradle.api.file.SourceDirectorySet
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.InputFile
 import org.gradle.api.tasks.InputFiles
 import org.gradle.api.tasks.OutputFile
 import org.gradle.api.tasks.TaskAction
+import java.io.ByteArrayOutputStream
```
advayDev1: don't add any of these 3 imports: http://www.groovy-lang.org/structure.html#_default_imports

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcPackLibrariesTask.groovy
```diff
@@ -21,6 +21,8 @@ import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.InputFiles
 import org.gradle.api.tasks.OutputDirectory
 import org.gradle.api.tasks.TaskAction
+import org.gradle.api.file.ConfigurableFileCollection
+import java.io.ByteArrayOutputStream
```
advayDev1: don't add this import: http://www.groovy-lang.org/structure.html#_default_imports

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcPackLibrariesTask.groovy
```diff
@@ -30,8 +32,8 @@ class J2objcPackLibrariesTask extends DefaultTask {
 
     // Generated ObjC binaries
     @InputFiles
-    def getInputLibraries() {
-        def staticLibraryPath = "${project.buildDir}/binaries/${project.name}-j2objcStaticLibrary"
+    ConfigurableFileCollection getInputLibraries() {
+        String staticLibraryPath = "${project.buildDir}/binaries/${project.name}-j2objcStaticLibrary"
         return project.files(getSupportedArchs().collect { arch ->
```
advayDev1: arch needs type specified

confile: What do  you mean by that?

advayDev1:
btw, this doesn't solve the many implicit defs all over our code (examples include closure parameter types, implied delegates, the implicit 'it' parameter, for loop variables, etc.).

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy
confile: done

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcXcodeTask.groovy
confile: fixed

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcXcodeTask.groovy
confile: fixed

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy
confile: fixed

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcCycleFinderTask.groovy
confile: fixed

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcTestTask.groovy
confile: fixed

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcPackLibrariesTask.groovy
advayDev1: the parameter of the closure 'arch' needs a type.  I don't know what the type is from a glance, it should look like:
`return project.files(getSupportedArchs().collect { SomeType arch ->`

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcTranslateTask.groovy
confile: fixed

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcPackLibrariesTask.groovy
confile: fixed

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcTestTask.groovy
confile: fixed

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcTranslateTask.groovy
confile: fixed

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcXcodeTask.groovy
confile: fixed

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcPackLibrariesTask.groovy
confile: its String now fixed

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcXcodeTask.groovy
```diff
@@ -132,9 +132,11 @@ class J2objcXcodeTask extends DefaultTask {
             logger.debug "Pod exists at path: ${getXcodeProjectDir()}"
             // check if this podspec has been included before
             // (a, b) = f() syntax unpacks the values in the return tuple
-            def (boolean podIntegrationExists, String targetPodLine) =
+            List<Object> checkPodDefResult =
```
advayDev1: this through L138 is awkward.
see http://docs.groovy-lang.org/latest/html/documentation/#_multiple_assignment
prefer:
```
def (boolean podIntegrationExists, String targetPodLine) = checkPodDefExistence(...)
```

advayDev1:
one last comment re: multi-assignment, then LGTM from me.
please wait for LG from @brunobowden

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcXcodeTask.groovy
confile: fixed

confile:
@advayDev1 I have to wait I still have no merge rights.

confile:
done

advayDev1:
@confile, i would be happy to merge the PR if i was confident bruno had no further comments, but he and i often look at different types of feedback in the code and he clearly was reviewing this before i was.  hence me not merging it myself.

confile:
ok

brunobowden:
Thanks @confile - I'm about to send out a PR for a better fix for the multiple assignment use of def, along with appropriate unit tests.

confile:
@brunobowden your welcome. There are missing types I was not sure about. 

See missing types here: https://github.com/j2objc-contrib/j2objc-gradle/issues/212

PR #208, opened 2015-06-21
New Version: 0.3.0-alpha

brunobowden:
- Increment the version number, 0.2 => 0.3 given substantial changess
- Expanded tags: 'android', 'java', 'j2objc', 'ios', 'iphone', 'objective-c', 'xcode'
- More extensive plugin description

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,7 +16,6 @@
 
 // Based on https://plugins.gradle.org/docs/publish-plugin
 
-// First, apply the publishing plugin
 plugins {
     id "com.gradle.plugin-publish" version "0.9.0"
 }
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -33,18 +32,24 @@ task wrapper(type: Wrapper) {
...
+version = '0.3.0-alpha'
 
 pluginBundle {
     website = 'https://www.github.com/j2objc-contrib/j2objc-gradle/'
     vcsUrl = 'https://github.com/j2objc-contrib/j2objc-gradle'
-    description = 'J2ObjC Gradle Build Plugin'
-    tags = ['j2objc']
+    description = 'Gradle Plugin for J2ObjC, which is an open-source tool from Google that translates ' +
+                  'Java source code to Objective-C for the iOS (iPhone/iPad) platform. The plugin is ' +
+                  'not affiliated with Google but was developed by former Google Engineers and others. ' +
```
advayDev1: haha.  do we really need 'former Google Engineers'?  I think the code stands on its own; anyone can find our LinkedIn pages.

brunobowden:
@advayDev1, @confile - I won't be committing this until we've worked through some other fixes and cleanup, but still wanted to send it through review.

@tomball - please see the changes to the description. I've repurposed the description you have on the j2objc.org website. Making it clear that this is for using a Google project but the plugin is not officially associated with Google.

build.gradle
```diff
@@ -33,18 +32,24 @@ task wrapper(type: Wrapper) {
...
-version = '0.2.2-alpha'
+version = '0.3.0-alpha'
 
 pluginBundle {
     website = 'https://www.github.com/j2objc-contrib/j2objc-gradle/'
     vcsUrl = 'https://github.com/j2objc-contrib/j2objc-gradle'
-    description = 'J2ObjC Gradle Build Plugin'
-    tags = ['j2objc']
+    description = 'Gradle Plugin for J2ObjC, which is an open-source tool from Google that translates ' +
+                  'Java source code to Objective-C for the iOS (iPhone/iPad) platform. The plugin is ' +
```
advayDev1: can we do some \n's between the first sentence and 'The plugin is'

advayDev1:
@brunobowden  - LGTM with above comments.
Let me know when we're about to release 0.3 so I can do some more testing on the precise commit, with multi-project builds and such.  I think I'm the only one using some of the new features we've added (but not using pods).

build.gradle
brunobowden: People can do a lot of things but my point is that all those things are so much work that people typically don't do it. Make it easy for someone and they do it more often and more willingly. For a random GitHub project, it takes time to look at the code and commits, then figure out the history of the authors. As you said, this takes the same information they could already find and makes pertinent part of it available to them.

advayDev1: fair enough

confile: former Google Engineers sounds always impressive :-)

advayDev1:
All pending bugfixes for 0.3 should be PR/merged into `release_v0.3.0-alpha` not `master`

advayDev1:
this PR would need to be reposted against the release_v0.3.0-alpha branch

brunobowden:
Reposted as #254... but I won't merge until we're definitive on the release.

Issue #200, opened 2015-06-20
Error: "ld can't map file, errno=22 file" -bundle_loader /../Debug-iphoneSimulator/MyApp.app/MyApp


confile:
``gradle j2objcXcode`` task works without problems but when I compile my xcode project I get the following error: 

![majestella_xcworkspace](https://cloud.githubusercontent.com/assets/64265/8268202/a67cb890-177c-11e5-964e-64e3cbca9a58.png)

Here is the image as text:

```
Ld /Users/mg/Library/Developer/Xcode/DerivedData/Majestella-grakkgsgamubapejewabyyvcxynu/Build/Products/Debug-iphonesimulator/MajestellaTests.xctest/MajestellaTests normal x86_64
    cd /Users/mg/Documents/Android/Android-Studio/Majestella/ios
    export IPHONEOS_DEPLOYMENT_TARGET=8.4
    export PATH="/Applications/Xcode-beta.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/usr/bin:/Applications/Xcode-beta.app/Contents/Developer/usr/bin:/usr/bin:/bin:/usr/sbin:/sbin"
[...]

PR #214, opened 2015-06-22
create sample project...

confile:
... based on this structure: https://github.com/j2objc-contrib/j2objc-gradle


Sample/.gitignore
```diff
@@ -0,0 +1,7 @@
+.gradle
+/local.properties
+/.idea/workspace.xml
+/.idea/libraries
+.DS_Store
+/build
+/captures
```
brunobowden: I have a more extensive .gitignore based on a lot of experience. I would use this in preference.

Looks like it's ok to have a nested .gitignore files:
http://stackoverflow.com/questions/3305869/are-multiple-gitignores-frowned-on

# Based on:
# http://stackoverflow.com/questions/16736856/what-should-be-in-my-gitignore-for-an-android-studio-project#18454251
# http://stackoverflow.com/questions/16640566/which-files-shouldnt-be-checked-in-into-version-control-in-android-studio/18454251#18454251


# TEMP FILES

# Ignore all temporary files
*~

# *.lock - this is used and abused by many editors for many different things.
*.lock


# GRADLE

# Temporarily ignored as these files are specially set for server side builds
local.properties


# ANDROID

# Android Studio generated files
bin/
gen/
build/
target/
google_generated/
.gradle/

# Release Binaries
android/android-release.apk
android/manifest-merger-release-report.txt


# IDE's

# Intellij IDEA (see https://intellij-support.jetbrains.com/entries/23393067)
# Note: Android Studio is a modified version of IntelliJ
*.iws
.idea/workspace.xml
.idea/tasks.xml
.idea/datasources.xml
.idea/dataSources.ids
.idea/libraries/

# Eclipse project files
# Android Studio or Intellij is recommend for working with this project
.classpath
.project
proguard/
.metadata/


# XCODE

# User settings
**/xcuserdata/

# Temporary files that should never be committed
*~.nib
DerivedData/

# Private settings (window sizes, bookmarks, breakpoints, custom executables, smart groups)
*.pbxuser
*.mode1v3
*.mode2v3
*.perspectivev3
# NB: also, whitelist the default ones, some projects need to use these
!default.pbxuser
!default.mode1v3
!default.mode2v3
!default.perspectivev3


# COCOAPODS

# Temp files generated by 'j2objcXcode' task
Xcode/Pods/


# OS X CLUTTER

# c.f. http://www.westwind.com/reference/os-x/invisibles.html
.DS_Store
.Trashes
*.swp


# Windows CLUTTER

# Mac OS X is recommended for all builds
Thumbs.db

brunobowden: Apologies as that should've have had ```code``` quotes.

Sample/ios/Podfile
```diff
@@ -0,0 +1,20 @@
+
+pod 'j2objc-shared', :path => '/Users/mg/Documents/Android/Android-Studio/Sample/shared/build'
```
brunobowden: Can you investigate doing relative paths here? There shouldn't be any mention of `/User/mg` in the committed code. So naturally this will break when someone moves around their source code... or forks the code to their desktop. With a relative path, then this'll work fine.

Sample/.gitignore
confile: fixed

Sample/ios/Podfile.lock
```diff
@@ -0,0 +1,14 @@
+PODS:
```
brunobowden: I think that the lock shouldn't be committed. It's something that CocoaPods should manage directly. The `.gitignore` that I provided will ignore this.

Sample/ios/Pods/Headers/Private/j2objc-shared/SharedClass.h
```diff
@@ -0,0 +1 @@
+../../../../../shared/build/j2objcSrcGen/SharedClass.h
```
brunobowden: I think that the `ios/Pods/...` directory shouldn't be ignored but just regenerated by the local CocoaPods utility as needed.

Sample/ios/Podfile
confile: @brunobowden Currently I cannot build in xcode because of this: https://github.com/j2objc-contrib/j2objc-gradle/issues/200

Sample/ios/Pods/Headers/Private/j2objc-shared/SharedClass.h
confile: What do you mean by that?

Sample/android/proguard-rules.pro
```diff
@@ -0,0 +1,17 @@
+# Add project specific ProGuard rules here.
+# By default, the flags in this file are appended to flags specified
+# in /Users/mg/Library/Android/sdk/tools/proguard/proguard-android.txt
```
brunobowden: Replace `/Users/mg/` with `/Users/<USERNAME>/`.

Sample/ios/Pods/Pods.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,548 @@
...
+			sourceTree = "<group>";
+		};
+		CC0202B5FE29C677AFBA8773 /* j2objc-shared */ = {
+			isa = PBXGroup;
+			children = (
+				993BFD7A3EF137863909474B /* Resources */,
+				832404709C371C8889063DD5 /* j2objcSrcGen */,
+			);
+			name = "j2objc-shared";
+			path = "/Users/mg/Documents/Android/Android-Studio/Sample/shared/build";
```
brunobowden: Note the reference to `/Users/mg/Documents/...` - we'll have to see how Xcode handle that.

brunobowden: Actually this should all be gone when you ignore the `ios/Pods` directory

Sample/android/proguard-rules.pro
confile: fixed

Sample/ios/Pods/Headers/Private/j2objc-shared/SharedClass.h
brunobowden: Corrected to "ios/Pods/... should be ignored". When you copy and paste the .gitignore example file in my first comment, it will ignore the `ios/Pods/...` directory and not commit any of the code contained within it. As generated files it should be regenerated on your local machine rather than be part of the repository.

Since these files contains references to your `/Users/mg/` directory, they will be different for every user... so at every check-in, the user will be prompted to commit files that have changed when the change is only relevant for them.

brunobowden:
NOTE: you can include a link to the specific part of the document. Note the "#folder-structure" in the url:

https://github.com/j2objc-contrib/j2objc-gradle#folder-structure

You can get this url, by hovering over the section title and then click on the link to the left:

screen shot 2015-06-22 at 12 55 02 pm

Sample/ios/Podfile
confile: I have a better idea. We can override the pod line every time the xcode task is performed. So the path is always correct.

brunobowden: I both are good steps:
1) It should be a relative path so there's no mention of your home directory
2) It should check the path and then write the file ONLY if the path is changed

Sample/.gitignore
```diff
@@ -0,0 +1,93 @@
...
+# c.f. http://www.westwind.com/reference/os-x/invisibles.html
+.DS_Store
+.Trashes
+*.swp
+
+
+# Windows CLUTTER
+
+# Mac OS X is recommended for all builds
+Thumbs.db
```
brunobowden: Add trailing newline

advayDev1:
Hi @confile  - should this PR be closed out?  Thanks.

PR #215, opened 2015-06-22
Add Style Guidelines to fix #97

### LINKED ISSUE #97: Style Guidelines for CONTRIBUTING.md

brunobowden:


brunobowden:
@advayDev1 and @confile - I'm interested in your feedback on this. The decision around parentheses is probably the biggest one for us to decide. I have conflicting opinions on this but want your input.

confile:
I strongly support parentheses for functions.

CONTRIBUTING.md
```diff
@@ -23,11 +23,47 @@ code review from anyone is welcome.  The PR author should keep the PR in a state
...
+
+* Function calls should use parentheses:
+```
+logging.captureStandardOutput(LogLevel.INFO)  // approved
+logging.captureStandardOutput LogLevel.INFO   // avoid
+```
+* Configure calls do not use parentheses:
+```
+project.exec {
+    executable 'j2objc'              // approved
```
confile: I think we should be consistent here and use parentheses here too.

confile:
LGTM beside my comment.

CONTRIBUTING.md
advayDev1: this is very unlike most groovy/gradle you find.
almost always configuration/DSL methods don't use parens.

CONTRIBUTING.md
```diff
@@ -23,11 +23,47 @@ code review from anyone is welcome.  The PR author should keep the PR in a state
...
+    args "-sourcepath", sourcepath   // approved
+
+    executable('j2objc')             // avoid
+    args("-sourcepath", sourcepath)  // avoid
+}
+```
+* Explicit Types instead of 'def':
+```
+String message = "a message"  // approved
```
advayDev1: please use 'single quotes' as this is a string not a GString

advayDev1:
LGTM - I'm ok with the call on parens (sorry I didn't see this, I had made my comment on the bug instead).

CONTRIBUTING.md
confile: @advayDev1 good point. So leave it.

brunobowden:
Fixed single quote issue. Now merging.

PR #211, opened 2015-06-22
PodFile updating reworked

brunobowden:
- Start of J2objcXcodeTaskTest, limited to static method so far
- Small cleanup in checkPodDefExistence
- Moved prefixes.properties to task package
- Reworked multiple assignment to not use `def`

PR #217, opened 2015-06-22
Override pod old pod definition if pod definition already exists…

confile:
…otherwise create a new pod definition. 

tested=yes

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcXcodeTask.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcXcodeTask.groovy
@@ -135,15 +135,15 @@ class J2objcXcodeTask extends DefaultTask {
             def (boolean podIntegrationExists, String targetPodLine) =
             checkPodDefExistence(podFile, xcodeTarget, getPodName())
 
+            String text = podFile.getText()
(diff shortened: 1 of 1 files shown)


confile:
@brunobowden This fixes your idea with relative paths of pod references. But instead of relative paths it overrides the pod reference on each build. Same effect.

brunobowden:
This is very much not the same thing:

*Current Situation*
1) I download and build the sample project.
2) It will modify the PodFile to point to a new location in my home
directory.
3) When I look in Git, it will show PodFile as a modified file that it
wants me to commit
4) Every single damn time, I have to commit a change to project, the
PodFile will always show up
5) If someone accidentally checks in the file, then I see a change in a
file completely unrelated to the fix that they were actually making

The central point is that your home directory should never ever appear in a
shared codebase. If that's happening, then something is going wrong.

confile:
Still having to much problems with the new plugin. As soon as I get it to work I take a look at it.

confile:
Should I close this pull request?

brunobowden:
@confile - please close this. Didn't mean you to waste work on this but I've just published a much more extensive reworking of the PodFile updating along with adding unit tests. See PR #211.

brunobowden:
@confile - thanks for looking in to this. Closing as it's superseded by #211.

Issue #203, opened 2015-06-20
Warning throwing wrong:  using ARC with J2ObjC, the project is missing one of the two flags required


confile:
My ``build.gradle``: 

```
// The 'java' plugin must be applied before the 'j2objc' plugin
apply plugin: 'java'
apply plugin: 'eclipse'
apply plugin: 'com.github.j2objccontrib.j2objcgradle'


sourceCompatibility = 1.7
version = '1.0'

buildscript {
  repositories {
      maven {
          url "https://plugins.gradle.org/m2/"
      }
  }
  dependencies {
      // Current Version: https://plugins.gradle.org/plugin/com.github.j2objccontrib.j2objcgradle
   //   classpath "gradle.plugin.com.github.j2objccontrib.j2objcgradle:j2objc-gradle:0.2.2-alpha"

	classpath files('/Users/mg/Downloads/j2objc-gradle/build/libs/j2objc-gradle-0.2.2-alpha.jar')
  }
}

repositories {
    mavenCentral()
    flatDir {
      dirs 'lib'
    }
}

dependencies {
}



// Plugin settings:
j2objcConfig {
    xcodeProjectDir "${projectDir}/../ios"   // Xcode workspace directory
    xcodeTarget "ios"                        // Xcode application target name
   
    translateFlags "--no-package-directories -use-arc"
	extraObjcCompilerArgs += '-fobjc-arc'
	
	finalConfigure()
}
```

Doing ``gradle clean`` I get: 

```
gradle clean
shared: using ARC with J2ObjC, the project is missing one of the two flags required:

j2objcConfig {
    // other settings
    translateFlags '-use-arc'
    extraObjcCompilerArgs = ['-fobjc-arc']
}
-fobjc-arc enables Automatic Reference Counting functionality in the compiler:
https://developer.apple.com/library/mac/releasenotes/ObjectiveC/RN-TransitioningToARC/Introduction/Introduction.html#//apple_ref/doc/uid/TP40011226-CH1-SW15
:clean

BUILD SUCCESSFUL
```

This is strange the correct args are in place.

brunobowden:
Not sure what's happening here. We'll debug this after all my changes are
merged in. The new syntax will push you to list each argument separately,
like this:

    translateArgs '--no-package-directories', '-use-arc'

advayDev1:
the bug here is putting multiple args in one string:
translateFlags "--no-package-directories -use-arc"

do
[...]

PR #218, opened 2015-06-23
FAQ: Address misconception re: multiple arguments

advayDev1:


advayDev1:
I address -prefixes FILE because some people might think FILE is an "argument to the -prefixes argument"

FAQ.md
```diff
@@ -9,6 +9,28 @@ You need at least [Gradle version 2.4](https://discuss.gradle.org/t/gradle-2-4-r
...
+
+    // ALSO CORRECT
+    translateArgs '-use-arc'
+    translateArgs '-prefixes'
+    translateArgs 'file.prefixes'
+
+    // WRONG
+    translateArgs '-use-arc -prefixes file.prefixes'
+    
+    // ALSO WRONG
```
confile: // WRONG

FAQ.md
```diff
@@ -9,6 +9,28 @@ You need at least [Gradle version 2.4](https://discuss.gradle.org/t/gradle-2-4-r
...
 
+### How do I properly pass multiple arguments to j2objc?
+
+Make sure your arguments are separate strings, not a single space-concatenated string.
+```
+j2objcConfig {
+    // CORRECT
+    translateArgs '-use-arc', '-prefixes', 'file.prefixes'
+
+    // ALSO CORRECT
```
confile: // CORRECT

confile:
see my comments then LGTM

brunobowden:
LGTM

This should really be an error message as well. I'll work on that message when I do the wider changes for all the flags => args changes.

FAQ.md
brunobowden: I agree simplifying it to just CORRECT or WRONG

advayDev1:
@brunobowden 
As I pointed out here: https://github.com/j2objc-contrib/j2objc-gradle/issues/203#issuecomment-114311174
this cannot be an error message

PR #220, opened 2015-06-23
SourceDirectorySet => FileCollection

brunobowden:
- Fixes #219
- Replace SourceDirectorySet with FileColleciton
- J2objcTestTaskTest unit tests for getTestNames()
- Minor comments

PR #221, opened 2015-06-23
Explicit Types for `->` iterators, fixes #205

### LINKED ISSUE #205: def => explicit types

brunobowden:
Depends on PR #220
- Add explicit types for `->` style iteration, fixes #205
- prefixProperties => packagePrefixes
- srcFiles => getSrcFiles() to make method call explicit
- Other similar cases as above
- ArrayList for {objcCompiler|linker}.withArguments iterators
- Configure: args => List<String> args

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
@@ -20,6 +20,7 @@ import com.github.j2objccontrib.j2objcgradle.tasks.J2objcUtils
 import groovy.transform.PackageScope
 import org.gradle.api.InvalidUserDataException
 import org.gradle.api.Project
+import org.gradle.api.Task
 import org.gradle.nativeplatform.NativeExecutableSpec
 import org.gradle.nativeplatform.NativeLibraryBinary
 import org.gradle.nativeplatform.NativeLibrarySpec
@@ -42,7 +43,7 @@ class J2objcNativeCompilation {
     def apply(File srcGenDir) {
         project.with {
             // Wire up dependencies with tasks created dynamically by native plugin(s).
-            tasks.whenTaskAdded { task ->
+            tasks.whenTaskAdded { Task task ->
                 // The Objective-C native plugin will add tasks of the form 'compile...Objc' for each
                 // combination of buildType, platform, and component.  Note that components having only
                 // one buildType or platform will NOT have the given buildType/platform in the task name, so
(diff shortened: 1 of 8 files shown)


brunobowden:
@advayDev1, @confile - PTAL

This fixes up I think the last remaining issues to fix #205. For the `J2objcNativeCompilation` class, it wasn't clear what where the types of the objcCompiler and linker members. By using `pringln "args type: ${args.getClass()}", I was able to determine that they were both ArrayList. Though I wasn't able to confirm that independently.

This is the repeated section of code you should pay most attention to during the review:

    target('ios_arm64') {
        def iosClangArgs = [
                '-arch',
                'arm64']
        iosClangArgs += iphoneClangArgs
        objcCompiler.withArguments { ArrayList args ->
            iosClangArgs.each { args << it }
        }
        linker.withArguments { ArrayList args ->
            iosClangArgs.each { args << it }
        }
    }

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
```diff
@@ -99,10 +100,10 @@ class J2objcNativeCompilation {
                                     '-arch',
                                     'arm64']
                             iosClangArgs += iphoneClangArgs
-                            objcCompiler.withArguments { args ->
+                            objcCompiler.withArguments { ArrayList args ->
```
confile: Why not List?

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
```diff
@@ -99,10 +100,10 @@ class J2objcNativeCompilation {
...
                                 iosClangArgs.each { args << it }
                             }
-                            linker.withArguments { args ->
+                            linker.withArguments { ArrayList args ->
```
confile: List?

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
```diff
@@ -111,10 +112,10 @@ class J2objcNativeCompilation {
                                     '-arch',
                                     'armv7']
                             iosClangArgs += iphoneClangArgs
-                            objcCompiler.withArguments { args ->
+                            objcCompiler.withArguments { ArrayList args ->
```
confile: List?

confile:
LGTM beside my comments

advayDev1:
Per https://docs.gradle.org/current/javadoc/org/gradle/nativeplatform/toolchain/CommandLineToolConfiguration.html#withArguments(org.gradle.api.Action)

The args should be List<String>.  ArrayList works but doesn't specify the element type but is also too specific for the container type.

advayDev1:
(btw, this branch is in upstream not your origin)

brunobowden:
Rebase and now merging...

Issue #31, opened 2015-04-14
Cycle Finder Task java_sources.mf not found anymore


confile:
To use cycle finder I do:

```
cycleFinderFlags "--whitelist /Applications/J2Objc/j2objc-sorce/jre_emul/cycle_whitelist.txt --sourcefilelist /Applications/J2Objc/j2objc-sorce/jre_emul/build_result/java_sources.mf"
```

In the latest source I could not find 

```
j2objc-sorce/jre_emul/build_result/java_sources.mf
```
anymore. So I use an old version.

@brunobowden How do you handle this?

brunobowden:
@tomball - was cyclefinder whitelist removed?

It used to be in the source tree:
j2objc-sorce/jre_emul/build_result/java_sources.mf

tomball:
The whitelist is still there:
https://github.com/google/j2objc/blob/master/jre_emul/cycle_whitelist.txt

Everything in jre_emul/build_result is a build artifact, including the
directory (AFAIK, it was never in the source tree). The
build_result/java_sources.mf is created when building jre_emul, so run
'make dist' or, if you just want the minimum build, run 'make
java_sources_manifest'.

confile:
@advayDev1 Do you have a good solution on this?

advayDev1:
never used cycle finder yet, sorry.

brunobowden:
When we do a plugin, I'll copy the file in manually for each version of
j2objc.

confile:
@brunobowden That is great.

confile:
@advayDev1 @brunobowden as Bruno suggested we can copy the two files into the plugin. Do you see a way of specifying the correct path to the files. It is easy when it is not in the plugin but in the plugin it is hard. 

Assume the files are in ``src/main/resources/file1.txt`` and ``src/main/resources/file2.txt``. Is it possible to get the paths to the files somehow?

brunobowden:
You don't just need those two files... you also need the complete J2ObjC source and to do the local build. This is enough of a narrow case that I don't think there's a good way to solve it. There's many more other issues that are more pressing.

MyCADDev:
Hi 
In j2objc-1.1 the jre_emul java_sources.mf is missing.

make dist is not generating this file

Could someone help?

thanks

brunobowden:
I believe this requires a special build step to create this file:

[...]

PR #228, opened 2015-06-24
args(String... args) syntax and unit tests

brunobowden:
- Reworked for all plugin config
- Separate out common functionality to method
- Unit tests

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -3,11 +3,13 @@
 
 ### What version of Gradle do I need?
 
-You need at least [Gradle version 2.4](https://discuss.gradle.org/t/gradle-2-4-released/9471), due to support for native compilation features.
+You need at least [Gradle version 2.4](https://discuss.gradle.org/t/gradle-2-4-released/9471),
+due to support for native compilation features.
 
 ### How do I solve the Eclipse error message ``Obtaining Gradle model...``?
 
-You have to create the [Gradle wrapper](https://docs.gradle.org/current/userguide/gradle_wrapper.html) first. Go to your project folder and do ``gradle wrapper``. Refresh your Eclipse project.
+You have to first create the [Gradle wrapper](https://docs.gradle.org/current/userguide/gradle_wrapper.html).
+Go to your project folder and do ``gradle wrapper``. Refresh your Eclipse project.
 
 ### How do I properly pass multiple arguments to j2objc?
 
@@ -33,7 +35,10 @@ j2objcConfig {
 
 ### How do I include Java files from additional source directories?
 
-In order to include source files from sources different than ``src/main/java`` you have to [modify the Java plugin's sourceSet(s)](https://docs.gradle.org/current/userguide/java_plugin.html#N11FD1). For example, if you want to include files from ``src-gen/base`` both into your JAR and (translated) into your Objective C libraries, then add to your ``build.gradle``:
+In order to include source files from sources different than ``src/main/java`` you have to
+[modify the Java plugin's sourceSet(s)](https://docs.gradle.org/current/userguide/java_plugin.html#N11FD1).
+For example, if you want to include files from ``src-gen/base`` both into your JAR and (translated) into
+your Objective C libraries, then add to your ``build.gradle``:
 
 ```
 sourceSets {
@@ -46,7 +51,7 @@ sourceSets {
 ```
 
 
-### How do I enable ARC for my objective-c classes?
+### How do I enable ARC for my Objective-C classes?
 
 Add the following to your configuration block. [See](https://developer.apple.com/library/mac/releasenotes/ObjectiveC/RN-TransitioningToARC/Introduction/Introduction.html#//apple_ref/doc/uid/TP40011226-CH1-SW15).
 
@@ -59,4 +64,4 @@ j2objcConfig {
 
 ### Error: implicit declaration of function 'JreRelease' is invalid in C99 [-Werror,-Wimplicit-function-declaration] JreRelease(this$0_)
 
-See: [How do I enable ARC for my objective-c classes?](#How do I enable ARC for my objective-c classes?)
+See: [How do I enable ARC for my Objective-C classes?](#how-do-i-enable-arc-for-my-objective-c-classes?)
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
@@ -16,6 +16,7 @@
 
 package com.github.j2objccontrib.j2objcgradle
 
+import com.google.common.annotations.VisibleForTesting
 import org.gradle.api.InvalidUserDataException
 import org.gradle.api.Project
 import org.gradle.api.tasks.util.PatternSet
@@ -33,7 +34,7 @@ class J2objcPluginExtension {
         this.project = project
 
         // The J2objcNativeCompilation effectively provides further extensions
-        // to the Project, by configuring the 'objective-c' native build plugin.
+        // to the Project, by configuring the 'Objective-C' native build plugin.
         // We don't want to expose the instance to clients of the 'j2objc' plugin,
         // but we also need to configure this object via methods on J2objcPluginExtension.
         nativeCompilation = new J2objcNativeCompilation(project)
@@ -67,22 +68,46 @@ class J2objcPluginExtension {
     String destLibDir = null
 
     /**
-     * Only generated source files, e.g. from dagger annotations.
+     * Generated source files directories, e.g. from dagger annotations.
      * <p/>
-     * The script will ignore changes in this directory so they must
+     * The plugin will ignore changes in this directory so they must
      * be limited to files generated solely from files within your
      * main and/or test sourceSets.
      */
-    String[] generatedSourceDirs = []
-    // TODO: void generatedSourceDirs(String... generatedSourceDirs) {
+    List<String> generatedSourceDirs = new ArrayList<>()
+    /**
+     * Add generated source files directories, e.g. from dagger annotations.
+     * <p/>
+     * The plugin will ignore changes in this directory so they must
+     * be limited to files generated solely from files within your
+     * main and/or test sourceSets.
+     *
+     * @param generatedSourceDirs adds generated source directories for j2objc translate
+     */
+    void generatedSourceDirs(String... args) {
+        appendArgs(this.generatedSourceDirs, 'generatedSourceDirs', args)
+    }
 
 
     // CYCLEFINDER
     /**
-     * Add command line arguments to cycle_finder command.
+     * Command line arguments for j2objc cycle_finder.
+     * <p/>
+     * A list of all possible arguments can be found here:
+     * http://j2objc.org/docs/cycle_finder.html
      */
-    final List<String> cycleFinderArgs = new ArrayList<>()
-    // TODO: void cycleFinderArgs(String... cycleFinderArgs) {
+    List<String> cycleFinderArgs = new ArrayList<>()
+    /**
+     * Add command line arguments for j2objc cycle_finder.
+     * <p/>
+     * A list of all possible arguments can be found here:
+     * http://j2objc.org/docs/cycle_finder.html
+     *
+     * @param cycleFinderArgs add args for 'cycle_finder' tool
+     */
+    void cycleFinderArgs(String... args) {
+        appendArgs(this.cycleFinderArgs, 'cycleFinderArgs', args)
+    }
     /**
      * Expected number of cycles, defaults to all those found in JRE.
      * <p/>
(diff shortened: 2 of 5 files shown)


brunobowden:
@advayDev1, @confile - this probably could still do with some improvements, interested to see what you think

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
```diff
@@ -358,4 +378,26 @@ class J2objcPluginExtension {
...
     }
+
+    // Provides a subset of "args" interface from project.exec as implemented by ExecHandleBuilder:
+    // https://github.com/gradle/gradle/blob/master/subprojects/core/src/main/groovy/org/gradle/process/internal/ExecHandleBuilder.java
+    // Allows the following:
+    // j2objcConfig {
+    //     translateArgs '--no-package-directories', '--prefixes', 'prefixes.properties'
+    // }
+    @PackageScope
+    static void configureArgs(List<String> listArgs, String nameArgs, String... newArgs) {
```
advayDev1: call it appendArgs.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
```diff
@@ -358,4 +378,26 @@ class J2objcPluginExtension {
...
         return finalConfigured
     }
+
+    // Provides a subset of "args" interface from project.exec as implemented by ExecHandleBuilder:
+    // https://github.com/gradle/gradle/blob/master/subprojects/core/src/main/groovy/org/gradle/process/internal/ExecHandleBuilder.java
+    // Allows the following:
+    // j2objcConfig {
+    //     translateArgs '--no-package-directories', '--prefixes', 'prefixes.properties'
+    // }
+    @PackageScope
```
advayDev1: private not package, right?

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
```diff
@@ -358,4 +378,26 @@ class J2objcPluginExtension {
...
+    static void configureArgs(List<String> listArgs, String nameArgs, String... newArgs) {
+        if (newArgs == null) {
+            throw new IllegalArgumentException("args == null!");
+        }
+        for (String arg in newArgs) {
+            if (arg.contains(' ')) {
```
advayDev1: i don't think this is true.  it is indeed fine to have spaces, say in file paths

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
advayDev1: or make it @VisibleForTesting

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
```diff
@@ -69,20 +70,31 @@ class J2objcPluginExtension {
...
      * main and/or test sourceSets.
+     *
+     * @param generatedSourceDirs generated source directories to translate
      */
-    String[] generatedSourceDirs = []
-    // TODO: void generatedSourceDirs(String... generatedSourceDirs) {
+    void generatedSourceDirs(String... generatedSourceDirs) {
+        configureArgs(this.generatedSourceDirs, 'generatedSourceDirs', generatedSourceDirs)
+    }
+    List<String> generatedSourceDirs = new ArrayList<>()
```
advayDev1: General comment for all of these pairs:
- You need to keep the GroovyDoc on the property/field itself.
- You should add a copy of that GroovyDoc, but where the comment says "adds..."

Users may want to add or replace all of them.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
brunobowden: I've had Project.Exec {} commands fail when an argument has had spaces. Need to investigate this further... for the most part they shouldn't need spaces.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
brunobowden: fixed

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
brunobowden: @VisibleForTesting

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
brunobowden: Duped GroovyDoc and modified the version for the property.

brunobowden:
Last issue: working out the correct type for outOfDate as the IDE currently shows it in error... it's ok when there's no explicit type:

```
inputs.outOfDate { InputFileDetails change ->
    // We must filter by srcFiles, since all possible input files are @InputFiles to this task.
    if (originalSrcFiles.contains(change.file)) {
        logger.debug "New or Updated file: " + change.file
        srcFilesChanged += project.files(change.file)
    } else {
        nonSourceFileChanged = true
        logger.debug "New or Updated non-source file: " + change.file
    }
}
```

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
```diff
@@ -67,22 +68,46 @@ class J2objcPluginExtension {
     String destLibDir = null
 
     /**
-     * Only generated source files, e.g. from dagger annotations.
+     * Generated source files directories, e.g. from dagger annotations.
```
confile: This is very unclear event to me. 

1. I think this is only important when you use ``translateArgs "--build-closure"``. Otherwise you do not need dependent sources. May be I am wrong here not sure.
2. I to get my dagger APT generated files from a different gradle project included I have to use: ``generatedSourceDirs "build/source/apt"`` as extend the sourceSets: 

```
sourceSets {
  apt{
    java{
      srcDir 'build/source/apt'
    }
  }
  main {
    java {
      srcDir 'build/source/apt'
    }
  }
}
``` 

Probably I am doing something wrong here but this is a good opportunity to make this clear.

3. Please describe how to include the source of a subproject you want to include in the source translation. The way I do it is: 

```
sourceSets {
  main {
    java {
      srcDir 'build/source/base'
    }
  }
}
```
Then using: ``generatedSourceDirs "build/source/base"`` and the following task: 

```
task copyBaseSources(type: Copy) {
  description "Copy all source files from base into the current project."
  exclude '**/*processor*'  // Annotation Processors should not be copied and translated
  from project(":base").projectDir.toString() + "/src/main/java"
  into project.projectDir.toString() + "/build/source/base"
}

tasks.compileJava.dependsOn += copyBaseSources
```

Happy to get your comments.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
```diff
@@ -67,22 +68,46 @@ class J2objcPluginExtension {
...
     /**
-     * Add command line arguments to cycle_finder command.
+     * Command line arguments for j2objc cycle_finder.
+     * <p/>
+     * A list of all possible arguments can be found here:
+     * http://j2objc.org/docs/cycle_finder.html
      */
-    final List<String> cycleFinderArgs = new ArrayList<>()
-    // TODO: void cycleFinderArgs(String... cycleFinderArgs) {
+    List<String> cycleFinderArgs = new ArrayList<>()
```
confile: There should be default args for whitelist sourcefilelist as in this issue: https://github.com/j2objc-contrib/j2objc-gradle/issues/31 

I tried to include then in the plugin and load it from there but still struggling how to get the correct path to the resources.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcTranslateTask.groovy
```diff
@@ -123,7 +124,9 @@ class J2objcTranslateTask extends DefaultTask {
         } else {
             boolean nonSourceFileChanged = false
             srcFilesChanged = project.files()
-            inputs.outOfDate { InputFileDetails change ->
+            inputs.outOfDate { change ->
```
confile: fix the type of ``change`` I think it is: 
```
org.gradle.api.tasks.incremental.InputFileDetails
```

brunobowden: I've tried that and the IDE doesn't like it, not sure why. Still working on this.

screen shot 2015-06-24 at 10 03 44 am

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
advayDev1: You should definitely not be copying another project's source into this project in order to get j2objc to work.
Have you tried setting it up using the functionality in #158 ?

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcTranslateTask.groovy
advayDev1: maybe because it should be String nameWithoutExt = change.file.name etc. etc. ...

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
brunobowden: Multi-project builds as @advayDev1 described is the proper way to do this.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
brunobowden: There isn't a good way to do this yet. At the moment, it requires downloading the J2ObjC source, doing a build and then setting up a reference. That's beyond the sophistication of most people and certainly not a default. For now, setting the expectedCycles to 40 is the best option.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcTranslateTask.groovy
brunobowden: I'll fix in another commit

brunobowden: Actually, just fixed it here

FAQ.md
```diff
@@ -59,4 +59,4 @@ j2objcConfig {
 
 ### Error: implicit declaration of function 'JreRelease' is invalid in C99 [-Werror,-Wimplicit-function-declaration] JreRelease(this$0_)
 
-See: [How do I enable ARC for my objective-c classes?](#How do I enable ARC for my objective-c classes?)
+See: [How do I enable ARC for my Objective-C classes?](#How do I enable ARC for my Objective-C classes?)
```
advayDev1: the hyperlink is actually
#how-do-i-enable-arc-for-my-objective-c-classes

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcTranslateTask.groovy
brunobowden: fixed

FAQ.md
brunobowden: fixed

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
advayDev1: So for now we're fine just failing on paths with spaces in them?

advayDev1:
LGTM.  Replied to one of your resolutions above.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
confile: @brunobowden @advayDev1 I will try it. Thanks for advise.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
confile: @brunobowden We could create the two files and put them into the plugin as default config. Making a note which version we have compiled with.

confile:
Beside my last comment LGTM

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
confile: @advayDev1 I have project B which includes project A. Why must project A use the j2objc plugin when using dependsOnJ2objcLib to include it?

advayDev1: @confile Because project A must generate j2objc libraries as well.  I'd suggest taking this conversation into its own issue if you'd like help debugging multi project builts

PR #227, opened 2015-06-24
xcodeProjectDir relative path, fixes #226

brunobowden:
- (String) cast to clarify ambiguous method call
- Unit tests for the fix (valid and invalid args)
- @Optional for possibly null j2objcConfig members
- Single quote for string
- Allow relative path for xcodeProjectDir
- Reworked error messages to handle relative paths

PR #231, opened 2015-06-24
xcodeProjectDir relative path + errors, fixes #226

### LINKED ISSUE #226: unset xcodeProjectDir throws exception: Ambiguous method overloading
For the following line, when `xcodeProjectDir == null`, it could match either File(File, String) or File(String, String), so it throws an exception as it's ambiguous. Even though xcodeProjectDir is explicitly type to String.

```
    @OutputFile
    File getPodFile() { new File(project.j2objcConfig.xcodeProjectDir, "Podfile") }
```

Here's the exception:
```
groovy.lang.GroovyRuntimeException: Ambiguous method overloading for method java.io.File#<init>.
Cannot resolve which method to invoke for [null, class java.lang.String] due to overlapping prototypes between:
	[class java.lang.String, class java.lang.String]
	[class java.io.File, class java.lang.String]
	at com.github.j2objccontrib.j2objcgradle.tasks.J2objcXcodeTask.getPodFile(J2objcXcodeTask.groovy:53)
```

brunobowden:
- (String) cast to clarify ambiguous method call
- Unit tests for the fix (valid and invalid args)
- @Optional for possibly null j2objcConfig members
- Single quote for string
- Allow relative path for xcodeProjectDir
- Reworked error messages to handle relative paths

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcXcodeTask.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcXcodeTask.groovy
@@ -16,10 +16,12 @@
 
 package com.github.j2objccontrib.j2objcgradle.tasks
 
+import com.google.common.annotations.VisibleForTesting
 import org.gradle.api.DefaultTask
 import org.gradle.api.InvalidUserDataException
 import org.gradle.api.tasks.Input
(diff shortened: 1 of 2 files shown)


brunobowden:
Switch from upstream branch (and deleted that branch). Merging based on LGMT with PR #227.

PR #232, opened 2015-06-25
def => explicit types, fixes #205

### LINKED ISSUE #205: def => explicit types

brunobowden:
- Figured out tricky explicit type for IncrementalTaskInputs
- Explicit types for J2objcNativeCompilation
- Explicit types for iterators

TESTED=yes

Verified:
a) `grep -R "def " .` => only shows Style Guidelines
b) Explicit types for all “->” statements
c) Explicit types for all implicit "it" iterators

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
@@ -40,7 +40,7 @@ class J2objcNativeCompilation {
     }
 
     @PackageScope
-    def apply(File srcGenDir) {
+    void apply(File srcGenDir) {
         project.with {
             // Wire up dependencies with tasks created dynamically by native plugin(s).
             tasks.whenTaskAdded { Task task ->
@@ -71,18 +71,16 @@ class J2objcNativeCompilation {
             file("${buildDir}/j2objcHackToForceCompilation/Empty.m").createNewFile()
             file("${buildDir}/j2objcHackToForceCompilation/EmptyTest.m").createNewFile()
 
-            def simulatorClangArgs = [
+            String[] simulatorClangArgs = [
                     '-isysroot',
-                    '/Applications/Xcode' +
-                            '.app/Contents/Developer/Platforms/iPhoneSimulator' +
-                            '.platform/Developer/SDKs/iPhoneSimulator.sdk',
+                    '/Applications/Xcode.app/Contents/Developer/Platforms/' +
+                            'iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator.sdk',
                     '-mios-simulator-version-min=8.3',
             ]
-            def iphoneClangArgs = [
+            String[] iphoneClangArgs = [
                     '-isysroot',
-                    '/Applications/Xcode' +
-                            '.app/Contents/Developer/Platforms/iPhoneOS' +
-                            '.platform/Developer/SDKs/iPhoneOS.sdk',
+                    '/Applications/Xcode.app/Contents/Developer/Platforms/' +
+                            'iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk',
                     '-miphoneos-version-min=8.3',
             ]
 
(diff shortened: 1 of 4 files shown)


brunobowden:
@advayDev1, @confile - I think these are the last changes needed for adding explicit types. The one thing I haven't been able to track down are these iterators for the Native Build:

screen shot 2015-06-24 at 6 39 09 pm

brunobowden:
@advayDev1, @confile - this should be the final commit for making all types explicit.

I've also created a "0.3 Release" Milestone. Please add relevant items to that.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcTranslateTask.groovy
```diff
@@ -122,30 +124,34 @@ class J2objcTranslateTask extends DefaultTask {
...
-            // TODO: figure out the explicit type for change
-            inputs.outOfDate { change ->
-                // We must filter by srcFiles, since all possible input files are @InputFiles to this task.
-                if (originalSrcFiles.contains(change.file)) {
-                    logger.debug "New or Updated file: " + change.file
-                    srcFilesChanged += project.files(change.file)
-                } else {
-                    nonSourceFileChanged = true
-                    logger.debug "New or Updated non-source file: " + change.file
+            inputs.outOfDate(new Action<InputFileDetails>() {
```
advayDev1: hmm... why is this needed?  (the wrapper with new Action)?

confile:
I cannot add anything I am not a contributor. 


Michael

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
```diff
@@ -71,14 +71,14 @@ class J2objcNativeCompilation {
             file("${buildDir}/j2objcHackToForceCompilation/Empty.m").createNewFile()
             file("${buildDir}/j2objcHackToForceCompilation/EmptyTest.m").createNewFile()
 
-            def simulatorClangArgs = [
+            String[] simulatorClangArgs = [
```
confile: isn't that List<String>

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
```diff
@@ -96,63 +96,83 @@ class J2objcNativeCompilation {
                     // https://docs.gradle.org/current/userguide/nativeBinaries.html#withArguments
                     clang(Clang) {
                         target('ios_arm64') {
-                            def iosClangArgs = [
+                            String[] iosClangArgs = [
```
confile: List<String>  ??

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
```diff
@@ -96,63 +96,83 @@ class J2objcNativeCompilation {
...
                             linker.withArguments { List<String> args ->
-                                iosClangArgs.each { args << it }
+                                iosClangArgs.each { String arg ->
+                                    args << arg
+                                }
                             }
                         }
                         target('ios_armv7') {
-                            def iosClangArgs = [
+                            String[] iosClangArgs = [
```
confile: List<String> ??

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
```diff
@@ -96,63 +96,83 @@ class J2objcNativeCompilation {
...
                             linker.withArguments { List<String> args ->
-                                iosClangArgs.each { args << it }
+                                iosClangArgs.each { String arg ->
+                                    args << arg
+                                }
                             }
                         }
                         target('ios_armv7s') {
-                            def iosClangArgs = [
+                            String[] iosClangArgs = [
```
confile: List<String> ??

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
```diff
@@ -96,63 +96,83 @@ class J2objcNativeCompilation {
...
                             linker.withArguments { List<String> args ->
-                                iosClangArgs.each { args << it }
+                                iosClangArgs.each { String arg ->
+                                    args << arg
+                                }
                             }
                         }
                         target('ios_i386') {
-                            def iosClangArgs = [
+                            String[] iosClangArgs = [
```
confile: List<String> ??

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
```diff
@@ -96,63 +96,83 @@ class J2objcNativeCompilation {
...
                             linker.withArguments { List<String> args ->
-                                iosClangArgs.each { args << it }
+                                iosClangArgs.each { String arg ->
+                                    args << arg
+                                }
                             }
                         }
                         target('ios_x86_64') {
-                            def iosClangArgs = [
+                            String[] iosClangArgs = [
```
confile: List<String> ??

brunobowden: @confile - if there's 2 or 3 repeats, then feel free to mark it. If there's
more, just note it once and then mark (repeated below). I'll make sure I
modify all of them.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcTranslateTask.groovy
brunobowden: I found it in a Gradle unit test:
https://github.com/gradle/gradle/blob/478a891a864c8749c194d42c52973cb1c8dfa8a7/subprojects/core/src/test/groovy/org/gradle/api/internal/changedetection/changes/DefaultTaskArtifactStateRepositoryTest.groovy#L558

brunobowden: I can't seem to make all the type checking work without this. Since it's in an official Gradle unit test, I'm inclined to believe it's correct.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
brunobowden: @confile - another thoughts is to say `Fix XXX, repeat 7 times below`. That succinctly explain it and I'll find the other references.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcTranslateTask.groovy
advayDev1: weird, we don't do that for other Action params that we pass as closures, but ok.

advayDev1:
LGTM with above comments

brunobowden:
The `String[]'s` are ok? I don't think that they're meant to be `List's`.

advayDev1:
@brunobowden - yes, totally fine:
http://docs.groovy-lang.org/docs/latest/html/documentation/#_array_initializers

advayDev1:
@brunobowden - sorry, i should have been clearer before.  it was LGTM with (my) above comments.  I was not saying you need to do @confile 's.

brunobowden:
I understood it that way

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcNativeCompilation.groovy
confile: @brunobowden Okay I will do that. But what about the fix?

brunobowden: Hi @confile, @advayDev1 confirmed that doing the `String[]` initialization was ok:

http://docs.groovy-lang.org/docs/latest/html/documentation/#_array_initializers

See previous comment from him:

screen shot 2015-06-26 at 7 00 16 am

confile: ok fine

confile:
LGTM

PR #233, opened 2015-06-25
CycleFinder Fix + improvements

brunobowden:
- Fix cycle_finder!!!
- Was erroneously calling j2objc instead of cycle_finder
- Comment that “40 expected cycles” is for version 0.9.6.1
- Move j2objcCycleFinder instructions to README
- Improved error for expected cycle mismatch
- Strong assert on zero cycles

Other Improvements:
- Remove redundant SourceDirectorySet imports
- “${var}” => “$var” in a few places

TESTED for both regular and difficult modes of Cycle Finder

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,14 +120,39 @@ disable the `j2objcTest` task, do the following:
 This task is disabled by default as it requires greater sophistication to use. It runs the
 `cycle_finder` tool in the J2ObjC release to detect memory cycles in your application.
 Objects that are part of a memory cycle on iOS won't be deleted as it doesn't support
-garbage collection. The tool will give you configuration feedback when you use it.
-To enable it:
+garbage collection.
+
+The basic setup will implicitly check for 40 memory cycles - this is the expected number
+of erroneous matches with `jre_emul` library for j2objc version 0.9.6.1. This may cause
+issues if this number changes with new versions of j2objc libraries.
 
     j2objcCycleFinder {
         enabled = true
     }
 
-For more details: http://j2objc.org/docs/Cycle-Finder-Tool.html
+##### Cycle Finder Advanced Setup
+
+This uses a specially annotated version of the `jre_emul` source that marks all the
+erroneously matched cycles such that they can be ignored. It requires downloading
+and building the J2ObjC source:
+
+1. Download the j2objc source to a directory (hereafter J2OJBC_REPO):<br>
+    https://github.com/google/j2objc
+2. Within the J2OJBC_REPO, run:<br>
+    `(cd jre_emul && make java_sources_manifest)`
+3. Configure j2objcConfig in build.gradle so CycleFinder uses the annotated J2ObjC source
+and whitelist. Note how this give expected cycles of zero.
+```
+j2objcConfig {
+    cycleFinderArgs '--whitelist', 'J2OBJC_REPO/jre_emul/cycle_whitelist.txt'
+    cycleFinderArgs '--sourcefilelist', 'J2OBJC_REPO/jre_emul/build_result/java_sources.mf'
+    cycleFinderExpectedCycles 0
+}
+```
+
+For more details:
+- http://j2objc.org/docs/Cycle-Finder-Tool.html
+- https://groups.google.com/forum/#!msg/j2objc-discuss/2fozbf6-liM/R83v7ffX5NMJ
 
 
 ### Plugin Development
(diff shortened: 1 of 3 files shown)


brunobowden:
@advayDev1, @confile - rather embarrassingly, Cycle Finder has been broken for a while. It was mistakenly using `j2objc` instead of `cycle_finder` as the executable. Since the tools by design use the same command line flags, nothing was apparent.

README.md
```diff
@@ -120,14 +120,38 @@ disable the `j2objcTest` task, do the following:
 This task is disabled by default as it requires greater sophistication to use. It runs the
 `cycle_finder` tool in the J2ObjC release to detect memory cycles in your application.
 Objects that are part of a memory cycle on iOS won't be deleted as it doesn't support
-garbage collection. The tool will give you configuration feedback when you use it.
-To enable it:
+garbage collection.
+
+The basic setup will implicitly check for 40 memory cycles - this is the expected number
+of erroneous matches with `jre_emul` library.
```
advayDev1: can you add a reference for the 40 memory cycles, for example a link to the j2objc project at a certain commit SHA?

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcCycleFinderTask.groovy
```diff
@@ -173,22 +145,25 @@ class J2objcCycleFinderTask extends DefaultTask {
                 errorOutput output;
                 standardOutput output;
             }
+            // No exception thrown means that cycle_finder found zero cycles
```
advayDev1: nice

advayDev1:
LGTM with minor comment

confile:
This is strange. I used cycle finder 4 days ago and it worked fine.

README.md
```diff
@@ -120,14 +120,38 @@ disable the `j2objcTest` task, do the following:
...
+##### Cycle Finder Advanced Setup
+
+This uses a specially annotated version of the `jre_emul` source that marks all the
+erroneously matched cycles such that they can be ignored. It requires downloading
+and building the J2ObjC source:
+
+1. Download the j2objc source to a directory (hereafter J2OJBC_REPO):<br>
+    https://github.com/google/j2objc
+2. Within the J2OJBC_REPO, run:<br>
+    `(cd jre_emul && make java_sources_manifest)`
```
confile: replace ``make -j4`` makes compile faster

README.md
```diff
@@ -120,14 +120,38 @@ disable the `j2objcTest` task, do the following:
...
+3. Configure j2objcConfig in build.gradle so CycleFinder uses the annotated J2ObjC source
+and whitelist. Note how this give expected cycles of zero.
+```
+j2objcConfig {
+    cycleFinderArgs '--whitelist', 'J2OBJC_REPO/jre_emul/cycle_whitelist.txt'
```
confile: As you pointed out here: https://github.com/j2objc-contrib/j2objc-gradle/issues/31 we should include the files in the plugin. I doubt that anyone beside us can use the plugin. There is to much to configure. Even I are to unable to configure it correctly (the new version :-)). We should reduce the complexity a lot.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcCycleFinderTask.groovy
confile: The task should output something like "no cycles found" rather than nothing.

brunobowden:
Because j2objc and cycle_finder use almost exactly the same arguments, using one command rather than the other should work. So instead of doing the cycle_finder checks, it was just doing the translation successfully and not reporting any errors. Then it would repeat the translation again in the translation task.

README.md
brunobowden: The simple setup is easy. The advanced setup is too complex and too low a priority to make automatic.

README.md
brunobowden: This only takes 2 seconds on my machine after `make clean`, so I think it's better to make it clearer rather than faster. Fair point for a big build.

README.md
brunobowden: This applies to version 0.9.6.1 - I still need to test for 0.9.7. This will cause problems if it changes between library versions but they seem pretty stable at this point. Added comment:

```
The basic setup will implicitly check for 40 memory cycles - this is the expected number
of erroneous matches with `jre_emul` library for j2objc version 0.9.6.1. This may cause
issues if this number changes with new versions of j2objc libraries.
```

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/J2objcCycleFinderTask.groovy
brunobowden: Added debug line:

    logger.debug "CycleFinder found 0 cycles!"

Issue #229, opened 2015-06-24
translateClassPaths and translateSourcepaths should have same syntax


confile:
Current state: 

```
  translateClassPaths "lib/json-20140107.jar", 
  translateSourcepaths "${projectDir}/lib/dagger-2.0.1-sources.jar"
```

preferred syntax: 

```
  translateClassPaths "lib/json-20140107.jar", 
  translateSourcepaths "lib/dagger-2.0.1-sources.jar"
```

brunobowden:
Good point. I think we should allow both paths but encourage relative paths, so you can do something like this:

     translateXxxxPaths '../relativeToProject/lib/some.jar'
     translateXxxxPaths "$projectDir}/dir/lib/some.jar"

@advayDev1 - your comments?

advayDev1:
we should just be consistent, if the rule is "all paths configured in j2objc are relative to project.file(X)" sgtm

brunobowden:
I think we should support an absolute path when it might be needed... but perhaps not document that and make everything else relative to the projectDir as you said.

advayDev1:
project.file(X) supports absolute as long as you put the root / in there.
as in, if we document that all paths are resolved per project.file(), all is good

brunobowden:
Fixed by #234

PR #234, opened 2015-06-25
translateSourcepaths to relative paths + refactor

brunobowden:
J2objConfig changes:
- translateSourcepaths to List<String>
- Refactor translateSourcepaths, translateClasspaths & generatedSourceDirs
- SourcePath => sourcepath, ClassPath => classpath
- Fully name parameters

Task changes:
- Remove @Optional from List<String> @Inputs (always defined)
- Reorder @Inputs for consistency
- Remove advanced CycleFinder setup instructions

Rename:
- Utils.srcDirs => srcSet, indicates a set of files
- Remove Utils.sourcepathJava and inline calls to getSrcDirs()
  (it was confusing returning sets of files and directories)

Testing:
- Added testJ2objcHome
- Added testSrcSet
- Added testThrowIfNoJavaPlugin_JavaPlugin

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
@@ -84,8 +84,8 @@ class J2objcConfig {
      *
      * @param generatedSourceDirs adds generated source directories for j2objc translate
      */
-    void generatedSourceDirs(String... args) {
-        appendArgs(this.generatedSourceDirs, 'generatedSourceDirs', args)
+    void generatedSourceDirs(String... generatedSourceDirs) {
+        appendArgs(this.generatedSourceDirs, 'generatedSourceDirs', generatedSourceDirs)
     }
 
 
@@ -105,8 +105,8 @@ class J2objcConfig {
      *
      * @param cycleFinderArgs add args for 'cycle_finder' tool
      */
-    void cycleFinderArgs(String... args) {
-        appendArgs(this.cycleFinderArgs, 'cycleFinderArgs', args)
+    void cycleFinderArgs(String... cycleFinderArgs) {
+        appendArgs(this.cycleFinderArgs, 'cycleFinderArgs', cycleFinderArgs)
     }
     /**
      * Expected number of cycles, defaults to all those found in JRE.
@@ -133,23 +133,36 @@ class J2objcConfig {
      *
      * @param translateArgs add args for the 'j2objc' tool
      */
-    void translateArgs(String... args) {
-        appendArgs(this.translateArgs, 'translateArgs', args)
+    void translateArgs(String... translateArgs) {
+        appendArgs(this.translateArgs, 'translateArgs', translateArgs)
     }
 
     /**
-     *  Libraries from ${projectDir}/lib/, e.g.: "json-20140107.jar", "somelib.jar".
+     *  Local jars for translation e.g.: "lib/json-20140107.jar", "lib/somelib.jar".
      *  This will be added to j2objc as a '-classpath' argument.
      */
-    List<String> translateClassPaths = new ArrayList<>()
+    List<String> translateClasspaths = new ArrayList<>()
     /**
-     *  Add libraries from ${projectDir}/lib/, e.g.: "json-20140107.jar", "somelib.jar".
+     *  Local jars for translation e.g.: "lib/json-20140107.jar", "lib/somelib.jar".
      *  This will be added to j2objc as a '-classpath' argument.
      *
-     *  @param add libraries for -classpath argument
+     *  @param translateClasspaths add libraries for -classpath argument
      */
-    void translateClassPaths(String... args) {
-        appendArgs(this.translateClassPaths, 'translateClassPaths', args)
+    void translateClasspaths(String... translateClasspaths) {
+        appendArgs(this.translateClasspaths, 'translateClasspaths', translateClasspaths)
+    }
+
+    /**
+     * Source jars for translation e.g.: "lib/json-20140107-sources.jar"
+     */
+    List<String> translateSourcepaths = new ArrayList<>()
+    /**
+     * Source jars for translation e.g.: "lib/json-20140107-sources.jar"
+     *
+     *  @param translateSourcepaths args add source jar for translation
+     */
+    void translateSourcepaths(String... translateSourcepaths) {
+        appendArgs(this.translateSourcepaths, 'translateSourcepaths', translateSourcepaths)
     }
 
 
@@ -157,7 +170,7 @@ class J2objcConfig {
     // WARNING: Do not use this unless you know what you are doing.
     // If true, incremental builds will be supported even if --build-closure is included in
     // translateArgs. This may break the build in unexpected ways if you change the dependencies
-    // (e.g. adding new files or changing translateClassPaths). When you change the dependencies and
+    // (e.g. adding new files or changing translateClasspaths). When you change the dependencies and
     // the build breaks, you need to do a clean build.
     boolean UNSAFE_incrementalBuildClosure = false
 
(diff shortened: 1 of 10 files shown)


brunobowden:
@advayDev1, @confile - this commit currently breaks the plugin. Nonetheless, I'm sharing an early version, so I get your preliminary feedback. Ultimately, I would like to share as much between the CycleFinder and Translation tasks as possible.

I'm also going to look at mocking `project.exec {...}`. I'm not sure how to do that yet, so let me know if you have any good ideas for that.

advayDev1:
i take it this is intentionally being delayed past 0.3?

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPluginExtension.groovy
```diff
@@ -273,7 +281,7 @@ class J2objcPluginExtension {
 
             logger.debug "$project:j2objcTranslate must use ${beforeProject.jar.archivePath}"
             j2objcConfig {
-                translateClassPaths += beforeProject.jar.archivePath
+                this.translateClasspaths += beforeProject.jar.archivePath
```
advayDev1: this doesn't make sense.  either you want to call j2objcConfig and leave out the this.X OR you want to use this.X and remove j2objcConfig

brunobowden: Good catch. Not sure what I was attempting to do here. Reverted back to what the line was before.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/CycleFinderTask.groovy
```diff
@@ -46,18 +46,11 @@ class CycleFinderTask extends DefaultTask {
...
-        }
-        generatedSourceDirs.each { String sourceDir ->
-            allFiles = allFiles.plus(project.files(sourceDir))
-        }
-        translateClassPaths.each { String classPath ->
-            allFiles = allFiles.plus(project.files(classPath))
-        }
+        allFiles = allFiles.plus(project.files(getTranslateClasspaths()))
+        allFiles = allFiles.plus(project.files(getTranslateSourcepaths()))
+        // Only care about changes in the generatedSourceDirs paths and not the contents
```
advayDev1: are you sure that's with this does?
L53 seems like it would care about the contents right?

we should test this either way, i.e. does an @InputFiles containing a project.files('dir') care about contents of 'dir'?

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/CycleFinderTask.groovy
```diff
@@ -103,25 +96,53 @@ class CycleFinderTask extends DefaultTask {
             windowsOnlyArgs.add("${getJ2ObjCHome()}/lib/cycle_finder.jar")
         }
 
-        String sourcepath = Utils.sourcepathJava(project)
+        if (getCycleFinderArgs().isEmpty()) {
+            String message =
+                    "CycleFinder is difficult to setup and use, though it's hoped to improve\n" +
```
advayDev1: though we hope to improve.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
```diff
@@ -133,23 +133,36 @@ class J2objcConfig {
...
      */
-    void translateArgs(String... args) {
-        appendArgs(this.translateArgs, 'translateArgs', args)
+    void translateArgs(String... translateArgs) {
+        appendArgs(this.translateArgs, 'translateArgs', translateArgs)
     }
 
     /**
-     *  Libraries from ${projectDir}/lib/, e.g.: "json-20140107.jar", "somelib.jar".
+     *  Local jars for translation e.g.: "lib/json-20140107.jar", "lib/somelib.jar".
```
confile: You should mention that assume you have the jars in your project ``lib/`` folder you do ``lib/...``.

confile:
see my comment then LGTM

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
brunobowden: Added "/lib/..." prefix to example

confile: LGTM

advayDev1:
Can you please point out the functional changes only?  The refactors generally look good, but it would be great to split the pure refactors from the changes that affect what the plugin does into two separate PRs.  AFAICT there are some changes regarding the source files vs. directories vs. trees but they are hard to find.  Can we make this a pure refactor only?

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/CycleFinderTask.groovy
brunobowden: Dropped this

brunobowden:
Please review #262 first. The second commit within this PR is more constrained. I could separate out some of the changes in J2objcConfig but it's hard to do much more of that. Hopefully looking at the 2nd commit is separated enough.

brunobowden:
Fixes #229

advayDev1:
I will PTAL this once all your other PRs are in and this is rebased.

brunobowden:
Sure thing Advay. I will follow up on this and the others on Monday.

brunobowden:
@advayDev1 - PTAL again for this. It's not quite a perfect refactor but pretty close. Some points I'd like to highlight for you:

1. I've removed the sourcepathJava method and inlined calls to `.getSrcDirs()`. I was very confused by this before and I think it helps to make it more explicit. So for two extra lines of code, you get to remove a method and it's test code.
1. What does it mean to add this to the classpathArgs: `:${project.buildDir}/classes`? Note that I'm also adding this to CycleFinderTask when it wasn't present there before. I want to add a comment better explaining its' purpose.
1. I'd like to refactor out common code between CycleFinderTask and TranslateTask in a future PR. I'm undecided between a common base class or util methods. I'm interested in your thoughts on this.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
```diff
@@ -166,43 +143,44 @@ class Utils {
...
-    static String absolutePathOrEmpty(Project proj, List<String> relativePaths) {
-        if (relativePaths.isEmpty()) {
-            return ''
-        } else {
-            // TODO: see if it works to return ':' for empty relativePaths
-            return ':' + relativePaths.collect({ proj.file(it).path }).join(':')
+    static List<String> j2objcLibs(String j2objcHome,
+                                   List<String> libraries) {
+        return libraries.collect { String library ->
+            return j2objcHome + "/lib/" + library
```
advayDev1: gsting interpolation?

advayDev1:
Responses:
1.  Yep SGTM, looks much more readable.
2.  No idea why we add /build/classes.  I assume it has something to do with autogenerated code where the final class file is present there but no source exists.  Have you tried translation without that argument?  I suspect in the normal case it isn't needed.  @confile - thoughts?
3.  I think we should go inheritance or composition, but not utility methods.  The long Utils class is one of the less readable parts of our codebase.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
brunobowden: fixed

brunobowden:
I'll try inheritance. Merging now...

PR #237, opened 2015-06-26
Rename non-Plugin classes to remove J2objc prefix.

### LINKED ISSUE #210: Remove "J2objc" prefix from all filenames (kept for config and tasks)
The `J2objc` was useful when it was a single file. Since moving in to the new Plugin architecture, it obscures the purpose of classes and files, as you have to look in more detail to parse it. Since the plugin is in an isolated GitHub repository and doesn't have to interact with other code except as a plugin, I don't believe there's any concerns about collisions.

My plan is to remove `J2objc` from all files and classes.

The `j2objcConfig` config and all tasks will retain the `j2objc` prefix.

advayDev1:
Fixes #210 

Hey @brunobowden , I went ahead and did this b/c I wanted to test the final case and also a subsequent PR fixes a 0.3-blocking issue.  I've confirmed that blame still works with and after this commit: 
https://github.com/advayDev1/j2objc-gradle/blame/fixMP/src/main/groovy/com/github/j2objccontrib/j2objcgradle/PluginConfigExtension.groovy

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ Tracked issue: https://github.com/j2objc-contrib/j2objc-gradle/issues/130**
         xcodeTarget "IosApp"                     // Xcode application target name
     
         // Help information on other settings:
-        // https://github.com/j2objc-contrib/j2objc-gradle/blob/master/src/main/groovy/com/github/j2objccontrib/j2objcgradle/PluginConfigExtension.groovy#L25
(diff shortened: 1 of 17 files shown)


brunobowden:
LGTM - with one caveat to rename `PluginConfigExtension` => `J2objcConfig`. That obviously has a closer connection to `j2objcConfig` and I think the J2objc prefix is merited in that case.

advayDev1:
SG.  Not squashing commits to ensure history is preserved for renames.  Merging now.

advayDev1:
Retested, and merging.

Issue #236, opened 2015-06-26
multi-project translateClassPaths broken


advayDev1:
I think this is due to #206

@brunobowden this blocks 0.3; working on it.

```
Caused by: groovy.lang.MissingMethodException: No signature of method: com.github.j2objccontrib.j2objcgradle.tasks.TranslateTask$_getAllInputFiles_closure3.doCall() is applicable for argument types: (java.io.File) values: [/Users/X/build/libs/lib-common.jar]
Possible solutions: doCall(java.lang.String), findAll(), findAll(), isCase(java.lang.Object), isCase(java.lang.Object)
	at com.github.j2objccontrib.j2objcgradle.tasks.TranslateTask.getAllInputFiles(TranslateTask.groovy:69)
	at com.github.j2objccontrib.j2objcgradle.tasks.TranslateTask_Decorated.getAllInputFiles(Unknown Source)
	at org.gradle.internal.reflect.JavaMethod.invoke(JavaMethod.java:75)
	at org.gradle.api.internal.project.taskfactory.AnnotationProcessingTaskFactory$PropertyInfo$4.create(AnnotationProcessingTaskFactory.java:460)
	at org.gradle.util.SingleMessageLogger.whileDisabled(SingleMessageLogger.java:166)
	at org.gradle.api.internal.project.taskfactory.AnnotationProcessingTaskFactory$PropertyInfo.getValue(AnnotationProcessingTaskFactory.java:458)
	at org.gradle.api.internal.project.taskfactory.AnnotationProcessingTaskFactory$Validator$1.call(AnnotationProcessingTaskFactory.java:262)
	at org.gradle.api.internal.file.collections.DefaultFileCollectionResolveContext.doResolve(DefaultFileCollectionResolveContext.java:120)
	at org.gradle.api.internal.file.collections.DefaultFileCollectionResolveContext.resolveAsFileTrees(DefaultFileCollectionResolveContext.java:75)
	at org.gradle.api.internal.file.collections.BuildDependenciesOnlyFileCollectionResolveContext$BuildableFileTreeInternalConverter.convertInto(BuildDependenciesOnlyFileCollectionResolveContext.java:50)
	at org.gradle.api.internal.file.collections.DefaultFileCollectionResolveContext.doResolve(DefaultFileCollectionResolveContext.java:98)
[...]

PR #238, opened 2015-06-26
BUGFIX - Fix multiproject class paths

advayDev1:
Do PR #237 first please.  After that, this is a single-line fix.


brunobowden:
LGTM for commit 92a3287 to fix #236.

PR #245, opened 2015-06-27
Fixes #244 - test beforeProjects first

### LINKED ISSUE #244: When project A depends on B, B's tests should run before A's build

advayDev1:
NOTE: This is a 0.3 bugfix and so the merge target is release_v0.3.0-alpha

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy
@@ -142,6 +142,15 @@ class J2objcPlugin implements Plugin<Project> {
             }
             lateDependsOn(project, 'assemble', 'j2objcAssemble')
 
+            // If users need to depend on this project to build other j2objc projects, they can use this
+            // marker task.
+            tasks.create(name: 'j2objcBuild', type: DefaultTask,
+                    dependsOn: ['j2objcAssemble', 'j2objcTest']) {
+                group 'build'
+                description "Marker task for all tasks that take part in regular j2objc builds"
+            }
+            lateDependsOn(project, 'build', 'j2objcBuild')
+
             // TODO: Where shall we fit this task in the plugin lifecycle?
             tasks.create(name: 'j2objcXcode', type: XcodeTask,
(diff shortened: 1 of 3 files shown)


src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy
```diff
@@ -142,6 +142,15 @@ class J2objcPlugin implements Plugin<Project> {
...
+            tasks.create(name: 'j2objcBuild', type: DefaultTask,
+                    dependsOn: ['j2objcAssemble', 'j2objcTest']) {
+                group 'build'
+                description "Marker task for all tasks that take part in regular j2objc builds"
+            }
+            lateDependsOn(project, 'build', 'j2objcBuild')
+
             // TODO: Where shall we fit this task in the plugin lifecycle?
             tasks.create(name: 'j2objcXcode', type: XcodeTask,
                     dependsOn: 'j2objcAssemble') {
```
brunobowden: I wonder about making j2objcXcode dependsOn: j2objcBuild... I'm concerned about updating Xcode when there might be failing tests.

brunobowden:
LGTM - but feel free to comment / push back on the j2objcXcode dependsOn: j2objcBuild

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy
advayDev1: i think our earlier decision was that users can choose to separate X's build from X's testing steps, and i think we should stick with that.  besides, xcode pods don't have the code in them.  i don't see a benefit to not updating the xcode project spec when tests are failing - it might even help debugging in Xcode if you prefer looking at the objective-c in that IDE.  it is just as bad to not update the xcode spec - the project is equally broken (possibly more broken in terms of user confusion).

advayDev1:
commented above and merging in.  if you feel strongly on the xcode -> build dependency, please make a new issue.  this PR is trying to fix a bug in multi-project which is part of the 0.3.0 release.  if we want to change xcode dependencies, IMO we can do that in master and not rush it in to this release.

PR #249, opened 2015-06-28
CycleFinderTask and TranslateTask Unit Tests

brunobowden:
Testing Improvements:
- Dependency injection for Project.exec {…} method
- MockProjectExec testing utility class
- First use of the Groovy MockFor system
- CycleFinderTaskTest good coverage
- TranslateTaskTest only basic coverage

Other Changes:
- exec args to have proper iteration for adding as args
- J2objcHome() standardize capitalization

PR #250, opened 2015-06-28
CycleFinderTask and TranslateTask Unit Tests

brunobowden:
Testing Improvements:
- Dependency injection for Project.exec {…} method
- MockProjectExec testing utility class
- First use of the Groovy MockFor system
- CycleFinderTaskTest good coverage
- TranslateTaskTest only basic coverage

Other Changes:
- exec args to have proper iteration for adding as args
- J2objcHome() standardize capitalization

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/CycleFinderTask.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/CycleFinderTask.groovy
@@ -16,13 +16,16 @@
 
 package com.github.j2objccontrib.j2objcgradle.tasks
 
+import com.google.common.annotations.VisibleForTesting
 import org.gradle.api.DefaultTask
+import org.gradle.api.Project
 import org.gradle.api.file.FileCollection
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.InputFiles
 import org.gradle.api.tasks.Optional
 import org.gradle.api.tasks.OutputFile
 import org.gradle.api.tasks.TaskAction
+import org.gradle.process.internal.ExecException
 
 /**
  * CycleFinder task checks for memory cycles that can cause memory leaks
@@ -70,7 +73,7 @@ class CycleFinderTask extends DefaultTask {
     int getCycleFinderExpectedCycles() { return project.j2objcConfig.cycleFinderExpectedCycles }
 
     @Input
-    String getJ2ObjCHome() { return Utils.j2objcHome(project) }
+    String getJ2objcHome() { return Utils.j2objcHome(project) }
 
     @Input
     @Optional
@@ -94,13 +97,18 @@ class CycleFinderTask extends DefaultTask {
 
     @TaskAction
     void cycleFinder() {
+        cycleFinderWithExec(project)
+    }
 
-        String cycleFinderExec = "${getJ2ObjCHome()}/cycle_finder"
+    @VisibleForTesting
+    void cycleFinderWithExec(Project projectExec) {
+
+        String cycleFinderExec = "${getJ2objcHome()}/cycle_finder"
         List<String> windowsOnlyArgs = new ArrayList<String>()
         if (Utils.isWindows()) {
             cycleFinderExec = 'java'
             windowsOnlyArgs.add('-jar')
-            windowsOnlyArgs.add("${getJ2ObjCHome()}/lib/cycle_finder.jar")
+            windowsOnlyArgs.add("${getJ2objcHome()}/lib/cycle_finder.jar")
         }
 
         String sourcepath = Utils.sourcepathJava(project)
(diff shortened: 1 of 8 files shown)


advayDev1:
@brunobowden - this is a large change to merge into a release branch without code review.
was that what you intended?

(i'll be able to review it thoroughly tomorrow...)

brunobowden:
This was already reviewed in PR #249. I commented in that PR that I was moving it but I should've added a more explicit comment here as well. Here's my comment from the old PR:

screen shot 2015-06-28 at 10 10 50 am

brunobowden:
My apologies @advayDev1 - this was my mistake and I didn't get it reviewed properly first. My confusion. I'll make sure that I'm careful about not doing this again.

advayDev1:
I might just be missing something - I don't think anyone actually reviewed #249 at all.
Do you still want us to review these changes?  Or do you just want to release them?

brunobowden:
They should still have review - my mistake at not getting that. This is a significant change when it comes to adding test coverage to the execution of command line tools. Important to have your review here.

brunobowden:
@advayDev1 - start with this review first... the other PR is still in progress

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/CycleFinderTask.groovy
```diff
@@ -121,22 +129,27 @@ class CycleFinderTask extends DefaultTask {
...
+                }
 
-                args windowsOnlyArgs
+                // Arguments
                 args "-sourcepath", sourcepath
-
                 if (classPathArg.size() > 0) {
                     args "-classpath", classPathArg
                 }
+                getCycleFinderArgs().each { String cycleFinderArg ->
```
advayDev1: i don't understand why you had to make these changes?  whats wrong with `args getCycleFinderArgs()` ?

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/CycleFinderTask.groovy
```diff
@@ -165,7 +180,7 @@ class CycleFinderTask extends DefaultTask {
                         "j2objcConfig {\n" +
                         "    cycleFinderExpectedCycles $cyclesFound\n" +
                         "}\n"
-                throw new Exception(message)
+                throw new IllegalArgumentException(message)
```
advayDev1: This is not an IllegalArgument.  At the end of the day the issue is bad user data (i.e. bad source code).  So just use a standard InvalidUserDataException from gradle (or however it is spelled).

Repeat elsewhere.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/CycleFinderTask.groovy
```diff
@@ -121,22 +129,27 @@ class CycleFinderTask extends DefaultTask {
         }
 
         String classPathArg = Utils.getClassPathArg(
-                project, getJ2ObjCHome(), getTranslateClassPaths(), getTranslateJ2objcLibs())
+                project, getJ2objcHome(), getTranslateClassPaths(), getTranslateJ2objcLibs())
```
advayDev1: still using project

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/CycleFinderTask.groovy
```diff
@@ -94,13 +97,18 @@ class CycleFinderTask extends DefaultTask {
 
     @TaskAction
     void cycleFinder() {
+        cycleFinderWithExec(project)
+    }
+
+    @VisibleForTesting
+    void cycleFinderWithExec(Project projectExec) {
```
advayDev1: See my overall comment on this CL.

advayDev1:
@brunobowden , my overall comments for this CL:

Having a separate projectExec vs. project variable and using both of them in the VisibleForTesting methods you have introduced makes the code much harder to understand.  I'd instead just wrap the thing we need to mock instead of introducing two projects that can get out of sync.  For example: a utility class ProcessExecutor.  The production implementation takes in a Project and its own exec method merely calls the configurable exec DSL method on the Project.  The test implementation mocks the exec method on that ProcessExecutor.  

Since this wasn't reviewed anyway, I'd prefer reverting this PR and using that method instead or discussing an alternative.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/CycleFinderTask.groovy
brunobowden: It makes the compare more difficult, so calling:

    args a
    args([b, c])

The mocked args() method simply does an addAll() operation, then you end up with [a, [b, c]] which isn't flattened. I'll need to rework the args function to flatten as it adds.

brunobowden:
Happy to revert this and fix it another way. This is getting too complex to work on a branch. So let's punt this to a 0.4 release. I'm leaving on vacation in a couple of days time, so I'd at least like to get a proper fix merged in to master.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/CycleFinderTask.groovy
advayDev1: I think we need to fix the mock to support the functionality of the actual implementation then.  Because it is easy for someone to use the list version of this method mistakenly.

advayDev1:
Sounds good.  Let's do the revert and get 0.3 out without this.

PR #253, opened 2015-06-28
Merge current v0.3.0-alpha branch into master

advayDev1:
There have been major changes to the release branch, merging early/often will make sure our work on master doesn't get disrupted.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,42 +36,42 @@ Further Customizations:
 
 * Function calls should use parentheses:
 ```
-logging.captureStandardOutput(LogLevel.INFO)  // approved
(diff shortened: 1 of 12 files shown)


advayDev1:
Note I'm not deleting the release branch after this - just keeping master up to date with the fixes

brunobowden:
I checked the PR numbers and not the contents and this looks correct. It's more complicated as we have merges going in both directions. In future, I don't think we should branch until we believe we're absolutely done.

brunobowden:
LGTM

advayDev1:
Yes, you're right about branching late.
We do need to think about what to do with features we don't want in the release though...

PR #262, opened 2015-07-01
Refactor of project.exec mocking

brunobowden:
- Proper mocking of project.exec using metaClass.invokeMethod
- Remove projectExec, cycleFinderWithExec and translateWithExec
- This has some ugly code but it's confined to testing


brunobowden:
@advayDev1, @confile - this is now ready for review. It took me four days to work out how to mock project.exec, shouldn't have been that difficult. The solution I've come up with does work but the code is ugly and I hope we can improve it in the future. I've asked around online to try and figure out a better way of doing this.

confile:
looks good except the test part which I have no clue of.

brunobowden:
@confile - the testing is a very good thing to learn about. Strong testing
makes it much easier to maintain and avoid breaking a large project. With
testing, you can make extensive changes and be confident that it still
works correctly. I'd suggest starting by looking at
`J2objcXcodeTaskTest.groovy` - where it verifies that a Podfile is changed
in the expected way.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/MockProjectExec.groovy
```diff
@@ -43,6 +45,57 @@ public class MockProjectExec {
...
+                    }
+                } else {
+                    return metaMethod?.invoke(delegate, args)
+                }
+            }
+        }
+    }
+
+    private static void logInvokeMethod(String name, Object[] args, MetaMethod metaMethod) {
+        String call = "call: $name, ${args.class}"
```
advayDev1: A few more comments in this method to explain what is going on please.

advayDev1:
I've read this through a few times now, but except for the 1 comment above, each thing that is odd or confusing already has TODO comments to make it better, so ship it :)

advayDev1:
LGTM

PR #271, opened 2015-07-10
Allow j2objc built libraries to link to other Gradle native libraries.

advayDev1:
Note the other libraries could be C[++], Objective-C[++], etc.

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
@@ -420,6 +420,23 @@ class J2objcConfig {
         }
     }
 
+    /**
+     * Additional native libraries to link to.
+     */
+    List<Map> extraNativeLibs = []
+    /**
+     * Add additional native library to link to.
+     * <p/>
+     * For example, if you built native library 'utils' in project 'common':
+     * <pre>
+     * extraNativeLib project: 'common', library: 'utils', linkage: 'static'
+     * </pre>
+     */
+    void extraNativeLib(Map spec) {
+        extraNativeLibs.add(spec)
+    }
+
+
     // XCODE
     /**
      * Directory of the target Xcode project.
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/NativeCompilation.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/NativeCompilation.groovy
@@ -239,6 +239,9 @@ class NativeCompilation {
                             beforeProjects.each { Project beforeProject ->
                                 lib project: beforeProject.path, library: "${beforeProject.name}-j2objc", linkage: 'static'
                             }
+                            j2objcConfig.extraNativeLibs.each { Map nativeLibSpec ->
+                                lib nativeLibSpec
+                            }
                         }
                     }
 
@@ -258,6 +261,9 @@ class NativeCompilation {
                             // Resolve dependencies from all java j2objc projects using the compiled static libs.
                             beforeProjects.each { Project beforeProject ->
                                 lib project: beforeProject.path, library: "${beforeProject.name}-j2objc", linkage: 'static'
+                            }
+                            j2objcConfig.extraNativeLibs.each { Map nativeLibSpec ->
+                                lib nativeLibSpec
                             }
 
                             // J2ObjC provided libraries for testing only


confile:
LGTM

src/main/groovy/com/github/j2objccontrib/j2objcgradle/NativeCompilation.groovy
```diff
@@ -329,6 +335,12 @@ class NativeCompilation {
         boolean added = beforeProjects.add(beforeProject)
         assert added
     }
+
+    private List<Map> nativeLibs = []
+    @PackageScope
+    void dependsOnNativeLib(Map spec) {
```
brunobowden: Why do you need this function? Why not inline it?

src/main/groovy/com/github/j2objccontrib/j2objcgradle/NativeCompilation.groovy
```diff
@@ -329,6 +335,12 @@ class NativeCompilation {
         boolean added = beforeProjects.add(beforeProject)
         assert added
     }
+
+    private List<Map> nativeLibs = []
```
brunobowden: Move private map alongside the `extraNativeLib` method.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
```diff
@@ -412,6 +412,14 @@ class J2objcConfig {
         }
     }
 
+    /**
+     * Additional native libraries to link to.
+     */
+    void extraNativeLib(Map spec) {
```
brunobowden: Rename to `addNativeLib`.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
```diff
@@ -412,6 +412,14 @@ class J2objcConfig {
         }
     }
 
+    /**
+     * Additional native libraries to link to.
```
brunobowden: Add fuller example of how this could be used. Ideally with an example of using a common native library.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
brunobowden: Let me take that back, `extra` is the common prefix used, so keep it that way to be consistent.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
brunobowden: I think this is what the example should be, though I don't have a good idea for the common native library, perhaps FormatterKit but I'm not sure?
```
extraNativeLib(library: 'FormatterKit', linkage: 'static')
```

brunobowden:
LGTM with minor comments. Ping me back if you question any of my feedback.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/NativeCompilation.groovy
```diff
@@ -239,6 +239,9 @@ class NativeCompilation {
                             beforeProjects.each { Project beforeProject ->
                                 lib project: beforeProject.path, library: "${beforeProject.name}-j2objc", linkage: 'static'
                             }
+                            nativeLibs.each { Map spec ->
```
brunobowden: spec => nativeLibSpec

src/main/groovy/com/github/j2objccontrib/j2objcgradle/NativeCompilation.groovy
```diff
@@ -259,6 +262,9 @@ class NativeCompilation {
...
```
brunobowden: spec => nativeLibSpec

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
advayDev1: the relevant example IMO is adding to a native library in another project.  otherwise you would just use the extra...Args form.  added such an example.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
advayDev1: fyi

src/main/groovy/com/github/j2objccontrib/j2objcgradle/NativeCompilation.groovy

src/main/groovy/com/github/j2objccontrib/j2objcgradle/NativeCompilation.groovy

src/main/groovy/com/github/j2objccontrib/j2objcgradle/NativeCompilation.groovy
advayDev1: note that unlike tasks, we keep the state of native compilation in the NativeCompilation object itself.  see 'dependsOnJ2objcLib' function above for consistency.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/NativeCompilation.groovy
advayDev1: see comment below

src/main/groovy/com/github/j2objccontrib/j2objcgradle/NativeCompilation.groovy
advayDev1: hmm i take that back.  we don't do it for extra...Args.  ok will inline

advayDev1:
done all, unit tests pass, merging

PR #272, opened 2015-07-10
Fully split debug and release builds.

advayDev1:
Do #271 first please.

```
- Use local.properties: j2objc.debugEnabled=false and j2objc.releaseEnabled=false to globally control
- Both flavors are enabled by default unless explicitly disabled
- j2objcTestDebug and j2objcTestRelease test debug & release binaries.
- j2objcAssembleDebug and j2objcAssembleRelease copy debug & release libraries.  Both copy generated source.
- native compiles also enabled/disabled as needed.
- markers j2objcTest, j2objcAssemble, and j2objcBuild remain and do both debug & release.
```

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
@@ -387,8 +387,8 @@ class J2objcConfig {
      * @param extraObjcSrcDirs add directories for Objective-C source to be compiled
      */
     void extraObjcSrcDirs(String... extraObjcSrcDirs) {
-        for (String arg in args) {
-            extraObjcSrcDirs += arg
+        for (String arg in extraObjcSrcDirs) {
+            this.extraObjcSrcDirs += arg
         }
     }
     /**
@@ -402,8 +402,8 @@ class J2objcConfig {
      * @param extraObjcCompilerArgs add arguments to pass to the native compiler.
      */
     void extraObjcCompilerArgs(String... extraObjcCompilerArgs) {
-        for (String arg in args) {
-            extraObjcCompilerArgs += arg
+        for (String arg in extraObjcCompilerArgs) {
+            this.extraObjcCompilerArgs += arg
         }
     }
     /**
@@ -417,8 +417,8 @@ class J2objcConfig {
      * @param extraLinkerArgs add arguments to pass to the native linker.
      */
     void extraLinkerArgs(String... extraLinkerArgs) {
-        for (String arg in args) {
-            extraLinkerArgs += arg
+        for (String arg in extraLinkerArgs) {
+            this.extraLinkerArgs += arg
         }
     }
 
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
@@ -16,9 +16,11 @@
 
 package com.github.j2objccontrib.j2objcgradle
 
+import com.github.j2objccontrib.j2objcgradle.tasks.Utils
 import com.google.common.annotations.VisibleForTesting
 import org.gradle.api.InvalidUserDataException
 import org.gradle.api.Project
+import org.gradle.api.Task
 import org.gradle.api.tasks.util.PatternSet
 import org.gradle.util.ConfigureUtil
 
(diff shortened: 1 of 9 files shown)


advayDev1:
I've been testing this for a while on my own builds.  I'm up to 17 projects now multiplied by {Android, iOS, and OSX}, and building debug only makes a huge difference in code-compile-test loops.

confile:
@advayDev1 What is the difference between debug and release build? Is there anything working different?

confile:
What is the difference?

brunobowden:
Usually there's a number of differences. Release is optimised for speed and
debug is optimised for debugging. For example, typically only debug has
debug symbols, extra runtime checks and so forth. This makes it easier to
debug but also slower during runtime and a bigger binary.

confile:
The question is what makes the plugin different?

advayDev1:
Previously you could not just build/test the debug vs. of the final libraries vs. the release ones.  You could only do both.

In a tight modify/compile/test cycle, you are usually only running the debug version.  This change cuts down the wait time for each cycle in half.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
```diff
@@ -79,8 +77,15 @@ class Utils {
             localPropertiesFile.withInputStream {
                 localProperties.load it
             }
-            result = localProperties.getProperty('j2objc.home')
+            result = localProperties.getProperty('j2objc.' + key)
         }
+        return result
+    }
```
brunobowden: Add dedicated unit test for this... see `testJ2objcHome_LocalProperties()` in UtilsTest for how to do this.

brunobowden: Add tests for both property exists and property doesn't exist... though I wonder if using the getProperty(String key, String defaultValue) might be neater.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
```diff
@@ -435,6 +445,28 @@ class J2objcConfig {
...
+
+        // For convenience, disable all debug and/or release tasks if the user desires.
+        // Note all J2objcPlugin-created tasks are of the form `j2objc.*(Debug|Release)?`
+        // however Native plugin-created tasks (on our behalf) are of the form `.*((D|d)ebug|(R|r)elease).*(j|J)2objc.*'
+        // so these patterns find all such tasks.
+
+        // Disable only if explicitly present and not true.
+        String debugEnabledProp = Utils.getLocalProperty(project, "debugEnabled")
+        boolean debugEnabled = debugEnabledProp == null || Boolean.parseBoolean(debugEnabledProp)
+        String releaseEnabledProp = Utils.getLocalProperty(project, "releaseEnabled")
```
brunobowden: Add support for default values and then reimplement as follows:

```
boolean debugEnabled = Boolean.parseBoolean(Utils.getLocalProperty(project, "debugEnabled", "true))
boolean releaseEnabled = Boolean.parseBoolean(Utils.getLocalProperty(project, "releaseEnabled", "true))
```

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy
```diff
@@ -104,15 +101,27 @@ class J2objcPlugin implements Plugin<Project> {
                 srcGenDir = j2objcSrcGenDir
             }
 
-            // Note the 'debugTestJ2objcExecutable' task is dynamically created by the Objective-C plugin applied
-            // on the above line.  It is specified by the testJ2objc native component.
-            tasks.create(name: 'j2objcTest', type: TestTask,
+            // Note the '(debug|release)TestJ2objcExecutable' tasks are dynamically created by the Objective-C plugin.
+            // It is specified by the testJ2objc native component.
```
brunobowden: Expand comment:

    // It is specified by the testJ2objc native component in NativeCompilation.groovy.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/AssembleLibrariesTask.groovy
```diff
@@ -0,0 +1,63 @@
...
+ * limitations under the License.
+ */
+
+package com.github.j2objccontrib.j2objcgradle.tasks
+import org.gradle.api.DefaultTask
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.InputDirectory
+import org.gradle.api.tasks.OutputDirectory
+import org.gradle.api.tasks.TaskAction
+/**
```
brunobowden: One line gap between import and class groovydoc.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/AssembleLibrariesTask.groovy
```diff
@@ -0,0 +1,63 @@
...
+    // Generated ObjC binaries
+    @InputDirectory
+    File libDir
+
+    @InputDirectory
+    File packedLibDir
+
+    @OutputDirectory
+    File getDestLibDir() {
+        return project.file(destLibDirPath)
```
brunobowden: Explicit method call:

    return project.file(getDestLibDirPath())

brunobowden:
LGTM with comments.

@advayDev1 - this is a welcome change. Couple of questions:

When disabling the release build for example, should the test binary be deleted? Not sure if there's any convention around this.

Please also update the README with this as a suggestion.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy
advayDev1: done

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/AssembleLibrariesTask.groovy
advayDev1: done

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/AssembleLibrariesTask.groovy
advayDev1: done

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
advayDev1: done - added 3 cases: file not present, key not present, key present

advayDev1:
Re the question: No, when you disable builds I don't see outputs being cleaned/deleted for other plugins.

All comments done, rebasing and merging.

advayDev1:
Builds fixed; now merging.

PR #277, opened 2015-07-14
TestTask exec unit tests, fixes #270

### LINKED ISSUE #270: Unit test check fails if project has only one unit test
Relatively minor issue, but caused me a bit of grief getting started with the plugin :)

In the TestTask line 134 the regex checks for the following:
```
int testCount = Utils.matchNumberRegex(outputStr, /OK \((\d+) tests\)/)
```
However if you have only one unit test the success message would be:

OK (1 test)

Which in turn fails the build.

brunobowden:
- Refactor testSetup() code for TestTask

### DIFF
--- a/src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TestTaskTest.groovy
+++ b/src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TestTaskTest.groovy
@@ -16,10 +16,10 @@
 
 package com.github.j2objccontrib.j2objcgradle.tasks
 
+import com.github.j2objccontrib.j2objcgradle.J2objcConfig
 import org.gradle.api.Project
 import org.gradle.api.file.FileCollection
 import org.gradle.testfixtures.ProjectBuilder
-import org.junit.Before
 import org.junit.Test
 
 /**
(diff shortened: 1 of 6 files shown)


brunobowden:
@alexmetcalfe - here's your fix. It's not even a 1 line fix as it's only a 1 character fix! Along with 50 lines of testing that needed to be added anyway. Not sure when 0.4 comes but look for that.

@advayDev1, @confile - feel free to look as well

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TestTaskTest.groovy
```diff
@@ -81,4 +88,71 @@ class TestTaskTest {
 
         assert expectedTestNames == testNames
     }
+
+    private void taskSetup() {
```
advayDev1: verb first - setupTask

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TestTaskTest.groovy
```diff
@@ -81,4 +88,71 @@ class TestTaskTest {
...
+        // For Utils.throwIfNoJavaPlugin()
+        proj.pluginManager.apply(JavaPlugin)
+
+        // For Utils.J2objcHome()
+        j2objcHome = File.createTempDir('J2OBJC_HOME', '').path
```
advayDev1: i've seen this in some of your other testing code too.  maybe refactor to a TestingUtils class?

advayDev1:
LGTM with minor comments

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TestTaskTest.groovy
brunobowden: fixed

brunobowden:
I've refactored to a TestingUtils class the shared code.

PR #278, opened 2015-07-15
Fix merge issues with #448fb04

brunobowden:
- Unit tests passed both before and after this commit


brunobowden:
@advayDev1 - I noticed a couple of issues with the release branch merge. The commits it produced were a lot more extensive than I was expecting. In future, we need to limit complex changes on release branches to avoid complications like this.

advayDev1:
agreed, i think in the future if we can't do a simple critical-bug-fix only release, we should abandon the branch and re-branch from HEAD.

LGTM.

advayDev1:
(btw, merging the PR since the build is broken)

Issue #230, opened 2015-06-24
Multi-project integration testing


advayDev1:
This is a caricature of my internal projects that use this plugin.
![image](https://cloud.githubusercontent.com/assets/11729521/8337843/36d4a230-1a62-11e5-9952-04529b664134.png)

You'll note it exercises many features:

- Simple java Project dependencies
- Duplicate dependencies (i.e. X depends on C directly but also on C transitively through Y)
- Build-closure translation of an external source jar
- Migrating JNI code to pure objective c
- Combining hand-written objective c code with j2objc translated objective c code
What else should we be testing for a 'complex' project?

confile:
@advayDev1 could you please post some more details of the build.gradle files. That would at least help me with the sample branch I am pushing.

brunobowden:
- Generated code, e.g. Dagger2 and AutoValue, not sure if we should just ignore Dagger1
- The incremental build after files are deleted or added in various different ways

brunobowden:
@advayDev1 - I've taken to modifying your original comment

advayDev1:
@brunobowden - please let's not edit comments from other authors in the future.  i know both of us /can/ but it is really confusing since there is no edit history AFAICT.  nor indication that someone else touched the comment.  scary for github really :P

I unedited the comment (as close as i could).

advayDev1:
@brunobowden , this is I think what I removed.  Also those aren't already in the current test project, but are things we'd like to also have:

- Generated code, e.g. Dagger2 and AutoValue, not sure if we should just ignore Dagger1
- The incremental build after files are deleted or added in various different ways
- Cycle Finder task, this involves creating two intentional memory cycles, marking one invalid and checking the other is counted

confile:
@brunobowden We should not support Dagger1. As far as I remember you cannot translate it because it uses reflection which is not available in objective-c. So forget about Dagger1.

confile:
@advayDev1 I do not understand you drawings. The legend shows symbols which occurs in every box. Does not help. Better to tell where the j2objc plugin is located.

brunobowden:
@advayDev1 - apologies Advay, I'll refrain in the future from modifying comments

advayDev1:
@confile - every dashed box is a distinct application of the j2objc plugin.

advayDev1:
[...]

PR #284, opened 2015-07-16
A few multi-project integration tests.  #230

advayDev1:
This is not high coverage, but gives a flavor of what these tests will look like.
NativeCompilation will also require integration tests of a similar kind
since our interaction with the ObjectiveC plugin cannot be unit-tested effectively.

### DIFF
--- a/src/test/groovy/com/github/j2objccontrib/j2objcgradle/MultiProjectTest.groovy
+++ b/src/test/groovy/com/github/j2objccontrib/j2objcgradle/MultiProjectTest.groovy
@@ -1,0 +1,92 @@
+package com.github.j2objccontrib.j2objcgradle
+
+import com.github.j2objccontrib.j2objcgradle.tasks.TestingUtils
+import groovy.util.logging.Slf4j
+import org.gradle.api.InvalidUserDataException
+import org.gradle.api.Project
+import org.gradle.nativeplatform.NativeBinarySpec
+import org.gradle.nativeplatform.NativeLibraryBinary
+import org.gradle.platform.base.BinarySpec
+import org.junit.Before
+import org.junit.Test
+
+/**
+ * Integration tests with multiple projects.
+ */
+@Slf4j
+class MultiProjectTest {
+
+    private Project rootProj, proj1, proj2
+    private J2objcConfig j2objcConfig1, j2objcConfig2
+
+    @Before
+    void setUp() {
+        // We can't use _ for unused slots because the other variables have already been declared above.
+        Object unused
+
+        // The root needs a local.properties, but nothing else.
+        (rootProj, unused, unused) = TestingUtils.setupProject(new TestingUtils.ProjectConfig())
+
+        // Project 1 will have its J2objcPlugin applied dynamically if needed.
+        (proj1, unused, j2objcConfig1) = TestingUtils.setupProject(new TestingUtils.ProjectConfig(
+                applyJavaPlugin: true,
+                createReportsDir: true,
+                rootProject: rootProj
+        ))
+        (proj2, unused, j2objcConfig2) = TestingUtils.setupProject(new TestingUtils.ProjectConfig(
+                applyJ2objcPlugin: true,
+                createReportsDir: true,
+                rootProject: rootProj
+        ))
+    }
+
+    @Test(expected = InvalidUserDataException)
+    void twoProjectsWithDependsOnJ2objcLib_MissingPluginOnProject1() {
+        j2objcConfig2.dependsOnJ2objcLib(proj1)
+    }
+
+    @Test
+    void twoProjectsWithDependsOnJ2objcLib_Works() {
+        proj1.pluginManager.apply(J2objcPlugin)
+        j2objcConfig1 = proj1.extensions.getByName('j2objcConfig')
+        j2objcConfig1.finalConfigure()
+
+        // Will force evaluation of proj1.
+        j2objcConfig2.dependsOnJ2objcLib(proj1)
+        j2objcConfig2.finalConfigure()
+
+        boolean evaluated = false
+        proj2.afterEvaluate { Project project ->
+            evaluated = true
+            // This forces the native plugin to convert the model rules to actual binaries.
+            project.tasks.realize()
+            project.bindAllModelRules()
+        }
+
+        // Should force evaluation of proj2.
+        rootProj.evaluationDependsOnChildren()
+        assert evaluated
+
+        // proj2 should build after proj1
+        assert TestingUtils.getTaskDependencies(proj2, 'j2objcPreBuild').contains(proj1.tasks.getByName('jar'))
+        assert TestingUtils.getTaskDependencies(proj2, 'j2objcPreBuild').contains(proj1.tasks.getByName('j2objcBuild'))
+
+        // proj2's Objective C libraries should link to proj1's Objective C libraries
+        proj2.binaries.each { BinarySpec binary ->
+            // Some of the binaries are Java classes, don't test those.
+            if (binary instanceof NativeBinarySpec) {
+                NativeBinarySpec nativeBinary = (NativeBinarySpec) binary
+                // Some of the binaries may be the user's custom native binaries, don't test those.
+                if (nativeBinary.name.contains('j2objc')) {
+                    // There may be other dependencies, so we use 'any'.
+                    assert nativeBinary.libs.any {
+                        NativeLibraryBinary bin = it.binary
+                        // The rest of the name indicates debug/release, architecture, etc.
+                        return bin.displayName.startsWith("static library '${proj1.name}-j2objc:")
+                    }
+                }
+            }
+        }
+    }
+
+}
--- a/src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TestingUtils.groovy
+++ b/src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TestingUtils.groovy
@@ -17,10 +17,14 @@
 package com.github.j2objccontrib.j2objcgradle.tasks
 
 import com.github.j2objccontrib.j2objcgradle.J2objcConfig
+import com.github.j2objccontrib.j2objcgradle.J2objcPlugin
 import groovy.util.logging.Slf4j
 import org.gradle.api.Project
+import org.gradle.api.Task
 import org.gradle.api.plugins.JavaPlugin
 import org.gradle.testfixtures.ProjectBuilder
+
+import java.util.concurrent.atomic.AtomicInteger
 
 /**
  * Util methods for testing projects.
@@ -31,36 +35,73 @@ import org.gradle.testfixtures.ProjectBuilder
 @Slf4j
 class TestingUtils {
 
+    static AtomicInteger projectIndex = new AtomicInteger()
+
+    static final class ProjectConfig {
+        /** to apply the J2objcPlugin */
+        boolean applyJ2objcPlugin = false
+        /** to apply the JavaPlugin (forced true if applyJ2objcPlugin = true) */
+        boolean applyJavaPlugin = false
+        /** to create the j2objcConfig extension (forced true if applyJ2objcPlugin = true) */
+        boolean createJ2objcConfig = false
+        /** to create projectDir/build/reports for writing output */
+        boolean createReportsDir = false
+        /** the parent project for this project, if any */
+        Project rootProject = null
+    }
+
     /**
      * Setup the project for testing purposes
      *
      * Configuration in setup() or individual test:
-     * (proj, j2objcHome, j2objcConfig) = TestingUtils.setupProject(true)
+     * (proj, j2objcHome, j2objcConfig) = TestingUtils.setupProject(new ProjectConfig(...))
      *
-     * @param createReportsDir to create projectDir/build/reports for writing output
      * @return [Project proj, String j2objcHome, J2objcConfig j2objcConfig] multiple assignment
      */
-    static Object[] setupProject(boolean createReportsDir) {
+    static Object[] setupProject(ProjectConfig config) {
         // Builds temporary project
-        Project proj = ProjectBuilder.builder().build()
+        Project proj = ProjectBuilder.builder()
+                .withParent(config.rootProject)
+                .withName("testProject${projectIndex.getAndIncrement()}")
+                .build()
+        if (config.rootProject != null) {
+            // When using a child project, the project's directory (a subdirectory of the root)
+            // is not created automatically.
+            assert proj.projectDir.mkdir()
+        }
 
-        // To satisfy Utils.throwIfNoJavaPlugin()
-        proj.pluginManager.apply(JavaPlugin)
+        if (config.applyJavaPlugin || config.applyJ2objcPlugin) {
+            // To satisfy Utils.throwIfNoJavaPlugin()
+            proj.pluginManager.apply(JavaPlugin)
+        }
 
         // To satisfy Utils.J2objcHome()
         String j2objcHome = File.createTempDir('J2OBJC_HOME', '').path
         File localProperties = proj.file('local.properties')
         localProperties.write("j2objc.home=$j2objcHome\n")
 
-        // For configuration of plugin inputs
-        J2objcConfig j2objcConfig = proj.extensions.create('j2objcConfig', J2objcConfig, proj)
+        J2objcConfig j2objcConfig = null
+        if (config.applyJ2objcPlugin) {
+            proj.pluginManager.apply(J2objcPlugin)
+            j2objcConfig = proj.extensions.getByName('j2objcConfig')
+        } else if (config.createJ2objcConfig) {
+            // can't use the config from the J2objcPlugin:
+            j2objcConfig = proj.extensions.create('j2objcConfig', J2objcConfig, proj)
+        }
 
-        if (createReportsDir) {
+        if (config.createReportsDir) {
             // Can't write @OutputFile in "projectDir/build/reports" unless directory exists
             File reportsDir = new File(proj.buildDir, "reports")
             assert reportsDir.mkdirs()
         }
 
+
         return [proj, j2objcHome, j2objcConfig]
+    }
+
+    static Set<? extends Task> getTaskDependencies(Project proj, String name) {
+        // Strange API requires passing the task as a parameter to getDependencies.
+        Task task = proj.tasks.getByName(name)
+        return task.taskDependencies.getDependencies(task)
     }
 }
--- a/src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/CycleFinderTaskTest.groovy
+++ b/src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/CycleFinderTaskTest.groovy
@@ -34,7 +34,11 @@ class CycleFinderTaskTest {
 
     @Before
     void setUp() {
-        (proj, j2objcHome, j2objcConfig) = TestingUtils.setupProject(true)
+        (proj, j2objcHome, j2objcConfig) = TestingUtils.setupProject(new TestingUtils.ProjectConfig(
+                applyJavaPlugin: true,
+                createReportsDir: true,
+                createJ2objcConfig: true
+        ))
     }
 
     // TODO: add java source files to the test cases
--- a/src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TestTaskTest.groovy
+++ b/src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TestTaskTest.groovy
@@ -84,7 +84,11 @@ class TestTaskTest {
     }
 
     private void setupTask() {
-        (proj, j2objcHome, j2objcConfig) = TestingUtils.setupProject(true)
+        (proj, j2objcHome, j2objcConfig) = TestingUtils.setupProject(new TestingUtils.ProjectConfig(
+                applyJavaPlugin: true,
+                createJ2objcConfig: true,
+                createReportsDir: true,
+        ))
 
         j2objcTest = (TestTask) proj.tasks.create(name: 'j2objcTest', type: TestTask) {
             testBinaryFile = proj.file("${proj.buildDir}/testJ2objc")
--- a/src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTaskTest.groovy
+++ b/src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTaskTest.groovy
@@ -35,7 +35,8 @@ class TranslateTaskTest {
 
     @Before
     void setUp() {
-        (proj, j2objcHome, j2objcConfig) = TestingUtils.setupProject(false)
+        (proj, j2objcHome, j2objcConfig) = TestingUtils.setupProject(
+                new TestingUtils.ProjectConfig(applyJavaPlugin: true, createJ2objcConfig: true))
     }
 
     // TODO: add java source files to the test cases


src/test/groovy/com/github/j2objccontrib/j2objcgradle/MultiProjectTest.groovy
```diff
@@ -0,0 +1,83 @@
...
+class MultiProjectTest {
+
+    private Project rootProj, proj1, proj2
+    private J2objcConfig j2objcConfig1, j2objcConfig2
+
+    @Before
+    void setUp() {
+        // We can't use _ for unused slots because the other variables have already been declared above.
+
+        def unused
```
brunobowden: Define as `Object unused` so that searching for "def" can find any stray uses.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TestingUtils.groovy
```diff
@@ -38,22 +44,48 @@ class TestingUtils {
...
      *
      * @param createReportsDir to create projectDir/build/reports for writing output
+     * @param createJ2objcConfig to create the j2objcConfig extension
+     * @param applyJavaPlugin to apply the JavaPlugin
+     * @param applyJ2objcPlugin to apply the J2objcPlugin
+     * @param rootProject the parent project for this project, if any
      * @return [Project proj, String j2objcHome, J2objcConfig j2objcConfig] multiple assignment
      */
-    static Object[] setupProject(boolean createReportsDir) {
+    static Object[] setupProject(boolean createReportsDir,
```
brunobowden: Default on first param?

     boolean createReportsDir = false

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TestingUtils.groovy
```diff
@@ -38,22 +44,48 @@ class TestingUtils {
...
+                                 boolean createJ2objcConfig = true,
+                                 boolean applyJavaPlugin = true,
+                                 boolean applyJ2objcPlugin = false,
+                                 Project rootProject = null) {
```
brunobowden: Lots of parameters with different types of parameters invites mistakes. Perhaps that's not as much of an issue for tests but it would nice to be better on this. Perhaps a closure or parameter map? Not sure what to suggest as an alternative.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TestingUtils.groovy
```diff
@@ -38,22 +44,48 @@ class TestingUtils {
...
         // Builds temporary project
-        Project proj = ProjectBuilder.builder().build()
+        Project proj = ProjectBuilder.builder()
+                .withParent(rootProject)
+                .withName("test${projectIndex.getAndIncrement()}")
+                .build()
+        if (rootProject != null) {
+            // When using a child project, the project's directory (a subdirectory of the root)
+            // is not created automatically.
+            proj.projectDir.mkdir()
```
brunobowden: Make parent directories if needed (adding `s` at end of method) and then assert it worked:

    assert proj.projectDir.mkdirs()

advayDev1: assert yes.
but we should not switch to mkdirs, since that would open the possibility that the root project did not exist.  the child project must be directly under the root

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TestingUtils.groovy
advayDev1: i really don't think we should use closures or maps.  they are not type safe.
i'll switch to using a POGO that is itself constructed using a map (so the type safety is at the caller's level not within the callee's implementation.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TestingUtils.groovy
advayDev1: see next reply

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TestingUtils.groovy
```diff
@@ -38,22 +44,48 @@ class TestingUtils {
...
+            proj.pluginManager.apply(J2objcPlugin)
+        }
+
         // For configuration of plugin inputs
-        J2objcConfig j2objcConfig = proj.extensions.create('j2objcConfig', J2objcConfig, proj)
+        J2objcConfig j2objcConfig = applyJ2objcPlugin ?
+                                    proj.extensions.getByName('j2objcConfig') :
+                                    (createJ2objcConfig ?
+                                     proj.extensions.create('j2objcConfig', J2objcConfig, proj) :
+                                     null)
```
brunobowden: I think this logic is clearer:

    J2objcConfig j2objcConfig = null
    if (applyJ2objcPlugin) {
        proj.pluginManager.apply(J2objcPlugin)
        j2objcConfig = proj.extensions.getByName('j2objcConfig')
    } else if (createJ2objcConfig) {
        // can't use the config from the J2objcPlugin:
        j2objcConfig = proj.extensions.create('j2objcConfig', J2objcConfig, proj)
    }

src/test/groovy/com/github/j2objccontrib/j2objcgradle/MultiProjectTest.groovy
advayDev1: done

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TestingUtils.groovy
advayDev1: ok

advayDev1:
repushed

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TestingUtils.groovy
```diff
@@ -38,22 +44,48 @@ class TestingUtils {
...
+    static Object[] setupProject(boolean createReportsDir,
+                                 boolean createJ2objcConfig = true,
+                                 boolean applyJavaPlugin = true,
+                                 boolean applyJ2objcPlugin = false,
+                                 Project rootProject = null) {
         // Builds temporary project
-        Project proj = ProjectBuilder.builder().build()
+        Project proj = ProjectBuilder.builder()
+                .withParent(rootProject)
+                .withName("test${projectIndex.getAndIncrement()}")
```
brunobowden: Perhaps name "projectX" rather than "testX"... or would it be better to make the test name as part of the project name? Might be helpful for debugging but not sure if it's worth the hassle.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/MultiProjectTest.groovy
```diff
@@ -0,0 +1,83 @@
...
+        // The root needs a local.properties, but nothing else.
+        (rootProj, unused, unused) = TestingUtils.setupProject(false, false, false, false)
+
+        // Project 1 will have its J2objcPlugin applied dynamically if needed.
+        (proj1, unused, j2objcConfig1) = TestingUtils.setupProject(true, false, true, false, rootProj)
+        (proj2, unused, j2objcConfig2) = TestingUtils.setupProject(true, true, true, true, rootProj)
```
brunobowden: This is hard to understand:

    TestingUtils.setupProject(true, true, true, true, rootProj)

Perhaps a builder style would work better:

    (proj2, unused, j2objcConfig2) = TestingUtils.buildProject().applyJ2objcPlugin().build()

In this case, the `applyJ2objcPlugin()` implicitly does `applyJavaPlugin().createJ2objcConfig()`. It is certainly easier to maintain and understand. I think using a long parameter list is extremely difficult to remember and error prone.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/MultiProjectTest.groovy
```diff
@@ -0,0 +1,83 @@
...
+import org.junit.Before
+import org.junit.Test
+
+/**
+ * Integration tests with multiple projects.
+ */
+@Slf4j
+class MultiProjectTest {
+
+    private Project rootProj, proj1, proj2
```
brunobowden: Rename to make it clearer:

    proj1 => projNonJ2objc
    proj2 => projJ2objc

src/test/groovy/com/github/j2objccontrib/j2objcgradle/MultiProjectTest.groovy
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    @Test(expected = InvalidUserDataException)
+    void twoProjectsWithDependsOnJ2objcLib_MissingPluginOnProject1() {
+        j2objcConfig2.dependsOnJ2objcLib(proj1)
+    }
+
+    @Test
+    void twoProjectsWithDependsOnJ2objcLib_Works() {
+        proj1.pluginManager.apply(J2objcPlugin)
```
brunobowden: Project projJ2objc2 = projNonJ2objc.pluginManager.apply(J2objcPlugin)

src/test/groovy/com/github/j2objccontrib/j2objcgradle/MultiProjectTest.groovy
```diff
@@ -0,0 +1,83 @@
...
+        // Should force evaluation of proj2.
+        rootProj.evaluationDependsOnChildren()
+        assert evaluated
+
+        // proj2 should build after proj1
+        assert TestingUtils.getTaskDependencies(proj2, 'j2objcPreBuild').contains(proj1.tasks.getByName('jar'))
+        assert TestingUtils.getTaskDependencies(proj2, 'j2objcPreBuild').contains(proj1.tasks.getByName('j2objcBuild'))
+
+        // proj2's Objective C libraries should link to proj1's Objective C libraries
+        proj2.binaries.each { def binary ->
```
brunobowden: def => Object

src/test/groovy/com/github/j2objccontrib/j2objcgradle/MultiProjectTest.groovy
```diff
@@ -0,0 +1,83 @@
...
+            // Some of the binaries are Java classes, don't test those.
+            if (binary instanceof NativeBinarySpec) {
+                NativeBinarySpec nativeBinary = (NativeBinarySpec) binary
+                // Some of the binaries may be the user's custom native binaries, don't test those.
+                if (nativeBinary.name.contains('j2objc') || nativeBinary.name.contains('J2objc')) {
+                    // There may be other dependencies, so we use 'any'.
```
brunobowden: Do you need a check like "assert evaluated" to make sure that the code gets here? Otherwise if an outer loop or conditional fails for any reason, it won't get to the proper test.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/MultiProjectTest.groovy
```diff
@@ -0,0 +1,83 @@
...
+        assert TestingUtils.getTaskDependencies(proj2, 'j2objcPreBuild').contains(proj1.tasks.getByName('jar'))
+        assert TestingUtils.getTaskDependencies(proj2, 'j2objcPreBuild').contains(proj1.tasks.getByName('j2objcBuild'))
+
+        // proj2's Objective C libraries should link to proj1's Objective C libraries
+        proj2.binaries.each { def binary ->
+            // Some of the binaries are Java classes, don't test those.
+            if (binary instanceof NativeBinarySpec) {
+                NativeBinarySpec nativeBinary = (NativeBinarySpec) binary
+                // Some of the binaries may be the user's custom native binaries, don't test those.
+                if (nativeBinary.name.contains('j2objc') || nativeBinary.name.contains('J2objc')) {
```
brunobowden: Where does the variation in capitalization come from?

How about using a regex? I haven't tested this:

    if (nativeBinary.name.matches("j2objc|J2objc")) {

brunobowden:
LGTM with comments... but likely you'll have pushback on some of them.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/MultiProjectTest.groovy
advayDev1: this is not the case.  both have j2objc.  in one test proj1 does not have it.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/MultiProjectTest.groovy
advayDev1: i'm not sure what this comment means.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/MultiProjectTest.groovy
advayDev1: done

src/test/groovy/com/github/j2objccontrib/j2objcgradle/MultiProjectTest.groovy
advayDev1: it has been refactored, PTAL

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TestingUtils.groovy
advayDev1: i don't think we should get the test name.  the callstack is clear enough.
i will call it testProjectX though.  testX is confusing.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/MultiProjectTest.groovy
advayDev1: correct, if none of those outer loops run, we won't assert the innermost one.
however testing that will be part of the NativeCompilation integration tests, not the MultiProject ones.

advayDev1:
thanks
can you please reply to: https://github.com/j2objc-contrib/j2objc-gradle/pull/284#discussion_r34754432
and clarify: https://github.com/j2objc-contrib/j2objc-gradle/pull/284#discussion-diff-34754553

src/test/groovy/com/github/j2objccontrib/j2objcgradle/MultiProjectTest.groovy
advayDev1: my error - tasks have varied capitalization, binaries are lowercase

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TestingUtils.groovy
```diff
@@ -31,36 +35,82 @@ import org.gradle.testfixtures.ProjectBuilder
...
 
+    static AtomicInteger projectIndex = new AtomicInteger()
+
+    static final class ProjectConfig {
+        /** to create projectDir/build/reports for writing output */
+        boolean createReportsDir = false
+        /** to create the j2objcConfig extension */
+        boolean createJ2objcConfig = true
+        /** to apply the JavaPlugin */
+        boolean applyJavaPlugin = true
```
brunobowden: To make the logic easier to understand. I think that everything should default to false and then be enabled if set to true, or forced to true by another option.

For example, applyJ2objcPlugin both forces applyJavaPlugin and createJ2objcConfig:

    applyJ2objcPlugin => applyJavaPlugin
    applyJ2objcPlugin => createJ2objcConfig

src/test/groovy/com/github/j2objccontrib/j2objcgradle/MultiProjectTest.groovy
```diff
@@ -0,0 +1,102 @@
...
+        Object unused
+
+        // The root needs a local.properties, but nothing else.
+        (rootProj, unused, unused) = TestingUtils.setupProject(new TestingUtils.ProjectConfig(
+                createReportsDir: false,
+                createJ2objcConfig: false,
+                applyJavaPlugin: false,
+                applyJ2objcPlugin: false,
+                rootProject: null
+        ))
```
brunobowden: By changing all defaults to false, this can be simplified to:

    TestingUtils.setupProject(new TestingUtils.ProjectConfig()

src/test/groovy/com/github/j2objccontrib/j2objcgradle/MultiProjectTest.groovy
```diff
@@ -0,0 +1,102 @@
...
+
+        // Project 1 will have its J2objcPlugin applied dynamically if needed.
+        (proj1, unused, j2objcConfig1) = TestingUtils.setupProject(new TestingUtils.ProjectConfig(
+                createReportsDir: true,
+                createJ2objcConfig: false,
+                applyJavaPlugin: true,
+                applyJ2objcPlugin: false,
+                rootProject: rootProj
+        ))
```
brunobowden: Simplify to:

    TestingUtils.setupProject(new TestingUtils.ProjectConfig(
        createReportsDir: true,
        applyJavaPlugin: true,
        rootProject: rootProj)

src/test/groovy/com/github/j2objccontrib/j2objcgradle/MultiProjectTest.groovy
```diff
@@ -0,0 +1,102 @@
...
+        (proj2, unused, j2objcConfig2) = TestingUtils.setupProject(new TestingUtils.ProjectConfig(
+                createReportsDir: true,
+                createJ2objcConfig: true,
+                applyJavaPlugin: true,
+                applyJ2objcPlugin: true,
+                rootProject: rootProj
+        ))
```
brunobowden: Simplify to:

    TestingUtils.setupProject(new TestingUtils.ProjectConfig(
        createReportsDir: true,
        applyJ2objcPlugin: true,
        rootProject: rootProj)

src/test/groovy/com/github/j2objccontrib/j2objcgradle/MultiProjectTest.groovy
brunobowden: The point is that with proj1 and proj2, you can't tell from the name what the difference is. The name should reflect the starting state of the project when it's initially created. If you then modify projJavaPlugin later on to apply(J2objcPlugin), then assign it to another variable with a different name.

    proj1 => projJavaPlugin
    proj2 => projJ2objcPlugin

src/test/groovy/com/github/j2objccontrib/j2objcgradle/MultiProjectTest.groovy
brunobowden: The point is to take `projJavaPlugin`, add the J2objcPlugin and then assign it to the `projJ2objc2` local variable. So that the name indicates that it's a different type of project than before.

    Project projJ2objc2 = projJavaPlugin.pluginManager.apply(J2objcPlugin)

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TestingUtils.groovy
```diff
@@ -31,36 +35,82 @@ import org.gradle.testfixtures.ProjectBuilder
...
 class TestingUtils {
 
+    static AtomicInteger projectIndex = new AtomicInteger()
+
+    static final class ProjectConfig {
+        /** to create projectDir/build/reports for writing output */
+        boolean createReportsDir = false
+        /** to create the j2objcConfig extension */
+        boolean createJ2objcConfig = true
+        /** to apply the JavaPlugin */
```
brunobowden: Expand comment:

    /** to apply the JavaPlugin, forced true if applyJ2objcPlugin is true */

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TestingUtils.groovy
```diff
@@ -31,36 +35,82 @@ import org.gradle.testfixtures.ProjectBuilder
 @Slf4j
 class TestingUtils {
 
+    static AtomicInteger projectIndex = new AtomicInteger()
+
+    static final class ProjectConfig {
+        /** to create projectDir/build/reports for writing output */
+        boolean createReportsDir = false
+        /** to create the j2objcConfig extension */
```
brunobowden: Expand comment:

    /** to create the j2objcConfig extension, forced true if applyJ2objcPlugin is true */

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TestingUtils.groovy
```diff
@@ -31,36 +35,82 @@ import org.gradle.testfixtures.ProjectBuilder
...
+                .build()
+        if (config.rootProject != null) {
+            // When using a child project, the project's directory (a subdirectory of the root)
+            // is not created automatically.
+            assert proj.projectDir.mkdir()
+        }
 
-        // To satisfy Utils.throwIfNoJavaPlugin()
-        proj.pluginManager.apply(JavaPlugin)
+        if (config.applyJavaPlugin) {
```
brunobowden: Force adding JavaPlugin if applyJ2objcPlugin is true:

    if (config.applyJavaPlugin || config.applyJ2objcPlugin) {

brunobowden:
I like the explicit config. Just see to my last round of comments and then it should be an LGTM.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/MultiProjectTest.groovy
advayDev1: for all three cases here, i kept all the assignments around to be clearer.
we're doing some odd setup here, and i think clarity is more important than conciseness for this setup part.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/MultiProjectTest.groovy
advayDev1: see above

src/test/groovy/com/github/j2objccontrib/j2objcgradle/MultiProjectTest.groovy
advayDev1: see below

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TestingUtils.groovy
advayDev1: done

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TestingUtils.groovy
advayDev1: somewhat different, but done

src/test/groovy/com/github/j2objccontrib/j2objcgradle/MultiProjectTest.groovy
advayDev1: hmm.  i don't like this.  the two names are aliases for one object.  it would not be the case that projJ2Objc2  has j2objc and projNonJ2objc does not have j2objc - they both do or they both don't at various points in time.  (unless i'm not understanding pluginManager.apply, i thought it was just returning the mutated object for configuration closure fluency, such as in a build.gradle script)

src/test/groovy/com/github/j2objccontrib/j2objcgradle/MultiProjectTest.groovy
advayDev1: i don't agree for the reason stated in my response below

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TestingUtils.groovy
advayDev1: done

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TestingUtils.groovy
advayDev1: done

advayDev1:
@brunobowden - on 2nd thought, did all comments except renaming the project variables for the reasons in my reply above.  also corresponding j2objcConfig2 to proj2 and j2objcConfig1 to proj1 makes it easier to read.

do you want to push back on the push back, or is this good to go?

src/test/groovy/com/github/j2objccontrib/j2objcgradle/MultiProjectTest.groovy
advayDev1: done

src/test/groovy/com/github/j2objccontrib/j2objcgradle/MultiProjectTest.groovy
advayDev1: nvm, done

src/test/groovy/com/github/j2objccontrib/j2objcgradle/MultiProjectTest.groovy
advayDev1: nvm, done

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TestingUtils.groovy
```diff
@@ -31,36 +35,73 @@ import org.gradle.testfixtures.ProjectBuilder
...
+    static final class ProjectConfig {
+        /** to create projectDir/build/reports for writing output */
+        boolean createReportsDir = false
+        /** to create the j2objcConfig extension (forced true if applyJ2objcPlugin = true) */
+        boolean createJ2objcConfig = false
+        /** to apply the JavaPlugin (forced true if applyJ2objcPlugin = true) */
+        boolean applyJavaPlugin = false
+        /** to apply the J2objcPlugin */
+        boolean applyJ2objcPlugin = false
+        /** the parent project for this project, if any */
```
brunobowden: Order the fields alphabetically, i.e. applyJavaPlugin, applyJ2objcPlugin, createJ2objcConfig, createReportsDir, rootProject. This also puts the "apply" bit as the first part of the config, which highlights a more important term than the "createReportsDir":

    (proj1, unused, j2objcConfig1) = TestingUtils.setupProject(new TestingUtils.ProjectConfig(
            applyJavaPlugin: true,     // Most important element should go first to be more distinctive
            createReportsDir: true,    // This is less important, so it's ok to be buried in the middle
            rootProject: rootProj
    ))

brunobowden:
I'd prefer otherwise but I'm ok with the variable names. The other improvements are definitely welcome. I think it helped a lot to be collaborating here.

Well deserved LGTM.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TestingUtils.groovy
advayDev1: done - important first makes sense.  don't think it would be generally alpha though.

advayDev1:
@brunobowden - thanks, marging..  btw, to nuance my response re: naming.  i don't have a general issue with those names, if the return value from apply(...) was doing what the names made it seem like (i.e. if apply did not return its argument mutated).

PR #287, opened 2015-07-16
Switch to @CompileStatic in most production code

### LINKED ISSUE #283: Consider @CompileStatic annotation for more type safety
Would help us catch more errors at compile-time.  But for the specific classes that use this, we would lose some Groovy magic.

@brunobowden - thoughts?

advayDev1:
- Found lots of issues with our (mis)use of FileCollection-ish
- Keeps project.exec and project.copy in dynamically typed extracted methods
- Does not touch the J2objcPlugin or NativeCompilation, which would become
  very unreadable or require use of internal classes for conversion.

Fixes #283 

Do #284 first please.
Unit tests pass.  My projects internally seem to build...

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
@@ -18,16 +18,23 @@ package com.github.j2objccontrib.j2objcgradle
 
 import com.github.j2objccontrib.j2objcgradle.tasks.Utils
 import com.google.common.annotations.VisibleForTesting
+import groovy.transform.CompileStatic
 import org.gradle.api.InvalidUserDataException
 import org.gradle.api.Project
 import org.gradle.api.Task
+import org.gradle.api.tasks.bundling.AbstractArchiveTask
 import org.gradle.api.tasks.util.PatternSet
 import org.gradle.util.ConfigureUtil
 
 /**
  * Further configuration uses the following fields, setting them in j2objcConfig within build.gradle
  */
+@CompileStatic
 class J2objcConfig {
+
+    static J2objcConfig from(Project project) {
+        return project.extensions.findByType(J2objcConfig)
+    }
 
     final protected Project project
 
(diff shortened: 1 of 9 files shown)


brunobowden:
This is great @advayDev1 - I'm very happy to see all these changes. Good luck trying to do this on MockProjectExec ;-)

How much of the project does this cover? Are there many exceptions in the test code?

Definitely add something to the style guidelines on how extensively @CompileStatic should be used and the hopefully narrow number of exceptions.

advayDev1:
@brunobowden - thanks!  it makes me feel much better about the correctness of the the build catching all of those File errors.  This covers all production code that I think we should convert.  I do not think converting Plugin or NativeCompilation gets anything.  their entire point is using the gradle DSL.

I do not think we should convert any of the tests to CompileStatic for 2 reasons:
- Where they use the Gradle DSL, they are using the code the way our users will use it. build.gradle files are never CompileStatic.
- test code should be fluent/compat where possible to encourage more tests to be written.  Gradle magic makes them shorter.
We should have the opposite policy for tests as for production: Only where useful, mark individual methods @CompileStatic.

i will update CONTRIBUTING.md if that looks good to you.

advayDev1:
@brunobowden ... sad.  i figured out a way to make all of the tasks fully statically typed.  and it works in production.

but it breaks mocking of project.exec (not my switch to strong typed closures, instead because it happens to now call project.exec on a static typed scope).  by making anything that calls project.exec statically typed, the mock executor's modification of the metaclass has no effect.  by extension all groovy-based mocking will not work in any CompileStatic code is my assumption.  check out commit 9ab0a75
if this is true, the only way we will be able to do true unit-tests is Mockito-style mocking, not Groovy-style mocking.  are we willing to pay that penalty for static typing?

brunobowden:
When I went to the Gradle Summit, I met the author of Mockito... who also
happened to work at Gradle for 3 years. So he knows both systems inside
out. I sent him an email two days ago looking for a better way to mock
project.exec. I haven't heard back from him yet but I'll extend the
discussion to cover this as well.

We definitely must have mocking of project.exec. It's such an important
part of the testing. I'm open to something that preserves that... or an
alternative which does similar coverage.

Let's keep the PR open for now while we look in to alternatives. I'll also
try patching it in and see if I can make it work.

advayDev1:
i did not incorporate 9ab0a75 into this PR.  mocking project.exec works correctly with the current version of this PR and all unit tests pass.  once we get this PR into an acceptable state, I think we should submit it.  and then wait for the Mockito response to do a better version of 9ab0a75 independently as a separate PR.

then we don't wait for perfect compile type checking while still covering all code except the calls to project.exec with static type checking.

brunobowden:
Agreed. LGTM.

brunobowden:
LGTM. Go ahead with the merge @advayDev1.

advayDev1:
dealing with conflicts, will then merge

advayDev1:
unit tests pass; own projects build.

however, it would be nice to get some more practical testing on this @brunobowden after i merge in (now)

PR #291, opened 2015-07-17
Typo fix.

advayDev1:
Everything works without this change because our configuration uses delegates not params as written.


brunobowden:
LGTM

Issue #267, opened 2015-07-03
Output stderr, stdout and cmd line for projectExec failures


brunobowden:
When cycleFinderTask or translateTask fails with the command line call. The internal error is obscured, making it harder to debug. The output should be captured and then output as part of the thrown exception. It should also show the full command line and perhaps the suggestion of trying to run it with javac to test that it works correctly. This is a lesson from issue #261 by @confile.

Currently a user will have to enable --debug mode and then dig through the debug output to see what command is run and then try debugging it.

Example of the current output which doesn't show any helpful messages:

```
:j2objcTranslate FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':j2objcTranslate'.
> Process 'command '/Applications/J2Objc/j2objc-0.9.7b/j2objc'' finished with non-zero exit value 1

* Try:
Run with --info or --debug option to get more log output.
```

confile:
This is a great idea!

brunobowden:
I've been testing this by adding `--nonsense` flags to each of the tasks. My feeling is that this is going to be a great change for tracking down tricky build issues.

Previously it would say:

    Process 'command '/Users/brunobowden/dev/j2objc-versions/j2objc-0.9.7/j2objc'' finished with non-zero exit value 1

Now it produces some very helpful error messages as follows:

    ...
    :base:j2objcCycleFinder UP-TO-DATE
    :base:j2objcTranslate
    Standard Output:

    Error Output:
    j2objc: invalid flag: --nonsense
    Usage: j2objc <options> <source files>
    use --help for a list of possible options

    :base:j2objcTranslate FAILED

    FAILURE: Build failed with an exception.

    * What went wrong:
    Execution failed for task ':base:j2objcTranslate'.
    > Error Output:
      j2objc: invalid flag: --nonsense
      Usage: j2objc <options> <source files>
      use --help for a list of possible options

      Command Line:
      /Users/brunobowden/dev/j2objc-versions/j2objc-0.9.6.1/j2objc -d ...
      Caused by:
      org.gradle.process.internal.ExecException: Process 'command ...

brunobowden:
@advayDev1 - see example above for how the error output has changed. This is going to be a great help for tricky issues.

advayDev1:
@brunobowden - totally agree.  waiting for you to merge it in :)

PR #294, opened 2015-07-17
Helpful output from project.exec failures

brunobowden:
- Catch ExecException and output useful information for failures
- For successes, debug log everything

Failure Output
- Command Line
- Error Output
- Standard Output

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/CycleFinderTask.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/CycleFinderTask.groovy
@@ -27,8 +27,6 @@ import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.InputFiles
 import org.gradle.api.tasks.OutputFile
 import org.gradle.api.tasks.TaskAction
-import org.gradle.process.ExecResult
-import org.gradle.process.internal.ExecException
 
 /**
  * CycleFinder task checks for memory cycles that can cause memory leaks
@@ -127,22 +125,48 @@ class CycleFinderTask extends DefaultTask {
         // TODO: comment explaining ${project.buildDir}/classes
         String classpathArg = Utils.joinedPathArg(classpathFiles) + ":${project.buildDir}/classes"
 
-        ByteArrayOutputStream output = new ByteArrayOutputStream()
+        ByteArrayOutputStream stdout = new ByteArrayOutputStream()
+        ByteArrayOutputStream stderr = new ByteArrayOutputStream()
         try {
-            execCycleFinder(cycleFinderExec, windowsOnlyArgs, sourcepathArg, classpathArg, fullSrcFiles, output)
+            logger.debug('CycleFinderTask - projectExec:')
+            Utils.projectExec(project, stdout, stderr, {
+                executable cycleFinderExec
+                windowsOnlyArgs.each { String windowsOnlyArg ->
+                    args windowsOnlyArg
+                }
+
+                // Arguments
+                args "-sourcepath", sourcepathArg
+                args "-classpath", classpathArg
+                getCycleFinderArgs().each { String cycleFinderArg ->
+                    args cycleFinderArg
+                }
+
+                // File Inputs
+                fullSrcFiles.each { File file ->
+                    args file.path
+                }
+
+                setStandardOutput stdout
+                setErrorOutput stderr
+            })
 
             logger.debug("CycleFinder found 0 cycles")
             assert 0 == getCycleFinderExpectedCycles()
 
-        } catch (ExecException exception) {
-            // Expected exception for non-zero exit of process
+        } catch (Exception exception) {
+            // ExecException is converted to InvalidUserDataException in Utils.projectExec(...)
 
-            // TODO show output for build failure instead of matchNumberRegex exception
-            String outputStr = output.toString()
-            // matchNumberRegex throws exception if regex isn't found
-            int cyclesFound = Utils.matchNumberRegex(outputStr, /(\d+) CYCLES FOUND/)
+            // Suppress rethrow of exception if it's expected non-zero exit matching "CYCLES FOUND"
+            String cyclesFoundStr = Utils.matchRegexOutputStreams(stdout, stderr, /(\d+) CYCLES FOUND/)
+            if (!(exception instanceof InvalidUserDataException) || !cyclesFoundStr?.isInteger()) {
+                // Didn't match (XX CYCLES FOUND)
+                throw exception
+            }
+
+            // Matched (XX CYCLES FOUND), so assert on cyclesFound
+            int cyclesFound = cyclesFoundStr.toInteger()
             if (cyclesFound != getCycleFinderExpectedCycles()) {
-                logger.error(outputStr)
                 String message =
                         "Unexpected number of cycles found:\n" +
                         "Expected Cycles:  ${getCycleFinderExpectedCycles()}\n" +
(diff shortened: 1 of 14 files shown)


brunobowden:
Fixes #267

@advayDev1 - this is still a work in progress. It still needs unit tests, some adjustment for CycleFinderTask and to get all the tests passing again. Please take an initial look.

@confile - this should be helpful for debugging some of the problems you've been seeing. See the example before and after below:

screen shot 2015-07-16 at 10 19 27 pm

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
```diff
@@ -237,13 +239,45 @@ class Utils {
     // TODO: In Gradle 2.5, we can switch to strongly-typed Actions, like:
     // https://docs.gradle.org/2.5/javadoc/org/gradle/api/Project.html#copy(org.gradle.api.Action)
     @CompileStatic(TypeCheckingMode.SKIP)
-    static ExecResult projectExec(Project proj,
-                                  @ClosureParams(value = SimpleType.class, options = "org.gradle.process.ExecSpec")
-                                  @DelegatesTo(ExecSpec)
-                                          Closure closure) {
-        proj.exec {
-            (delegate as ExecSpec).with closure
+    static void projectExec(Project proj,
```
advayDev1: why remove the ExecResult return value?  this should be an enhanced stand-in for project.exec, so if callers need it, they can use it.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
```diff
@@ -237,13 +239,45 @@ class Utils {
...
+                            @ClosureParams(value = SimpleType.class, options = "org.gradle.process.ExecSpec")
+                            @DelegatesTo(ExecSpec)
+                                    Closure closure) {
+
+        // Uses internal classes to get the command line and other closure parameters
+        ExecHandleBuilder execHandleBuilder = new ExecHandleBuilder()
```
advayDev1: Can you provide a pointer to which internal class instantiates this using an ExecHandleBuilder?

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
```diff
@@ -237,13 +239,45 @@ class Utils {
...
+        ConfigureUtil.configure(closure, execHandleBuilder)
```
advayDev1: IMPORTANT: the closure can have side effects; it is not correct to do this twice (also on L254).
looking through Javadocs, I don't see a correct way to do this without manually wrapping the ExecSpec that proj.exec gives as the delegate, and then applying those modifications to your execHandleBuilder, then printing the command line, then returning from the proj.exec closure.

In pseudo-code:
````
try {
proj.exec {
    WrappedExecSpec wrapped = new WrappedExecSpec(delegate as ExecSpec)
    wrapped.with closure
    ExecHandleBuilder execHandleBuilder = new EHB()
    wrapped.applyConfigsTo(execHandleBuilder)
    printExecHandleBuilderCommandLine
}

...
```

advayDev1:
@brunobowden - unfortunately i don't think this works.  see my suggestion inline

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
advayDev1: alternate pseudo-design:
````
try {
proj.exec {
ExecHandleBuilder execHandleBuilder = new EHB()
TeeExecSpec bridge = new TeeExecSpec(delegate as ExecSpec, execHandleBuilder)
bridge.with closure
printExecHandleBuilderCommandLine
}
}
````

**Comment by brunobowden** [comment]
@confile - please have a serious look at this now. The new version is much simpler and should be side effect free. I still need to do some more testing on it but this is good for a thorough review.

File: src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy [hunk 4]
```diff
@@ -237,12 +246,65 @@ class Utils {
...
+            @DelegatesTo(ExecSpec)
+                    Closure closure) {
+
+        // Uses internal classes to get the command line and other closure parameters
+        ExecSpec execSpec = null
+        ExecResult execResult
+
+        try {
+            execResult = proj.exec {
+                execSpec = delegate as ExecSpec
```
advayDev1: Nice - way simpler!!

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
```diff
@@ -237,12 +246,65 @@ class Utils {
...
-            (delegate as ExecSpec).with closure
+    static ExecResult projectExec(
+            Project proj,
+            ByteArrayOutputStream stdout,
+            ByteArrayOutputStream stderr,
+            @ClosureParams(value = SimpleType.class, options = "org.gradle.process.ExecSpec")
+            @DelegatesTo(ExecSpec)
+                    Closure closure) {
+
+        // Uses internal classes to get the command line and other closure parameters
```
advayDev1: Don't think this comment is relevant anymore

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
```diff
@@ -237,12 +246,65 @@ class Utils {
...
+    @VisibleForTesting
+    static void logExecSpecOutput(
+            ByteArrayOutputStream stdout, ByteArrayOutputStream stderr, ExecSpec execSpec) {
+
+        if (execSpec == null) {
+            log.debug('execSpec is null')
+            return
+        }
+
+        log.debug('Command Line: ' + execSpec.getCommandLine().join(' '))
```
advayDev1: Would be nice to split the commandLine from the "Command line: " prefix with a newline.  that way people can directly copy the line to test manually.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/CycleFinderTask.groovy
```diff
@@ -127,22 +125,42 @@ class CycleFinderTask extends DefaultTask {
         // TODO: comment explaining ${project.buildDir}/classes
         String classpathArg = Utils.joinedPathArg(classpathFiles) + ":${project.buildDir}/classes"
 
-        ByteArrayOutputStream output = new ByteArrayOutputStream()
+        ByteArrayOutputStream stdout = new ByteArrayOutputStream()
+        ByteArrayOutputStream stderr = new ByteArrayOutputStream()
         try {
-            execCycleFinder(cycleFinderExec, windowsOnlyArgs, sourcepathArg, classpathArg, fullSrcFiles, output)
+            logger.debug('CycleFinderTask - projectExec:')
+            Utils.projectExec(project, stdout, stderr, {
```
advayDev1: any particular reason you re-inlined all these methods?
i liked the separation. but it doesn't matter to me that much.

advayDev1:
LGTM with comments.  no need to rereply with any push back, i don't feel that strongly to oppose.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
brunobowden: Didn't realize it but there's nothing internal now.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
brunobowden: I'm so much happier with this as well

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
brunobowden: Newline after commandline

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/CycleFinderTask.groovy
brunobowden: I preferred not to have the separation... as the configuration of projectExec is such an integral part of the task that I preferred it inline. It was separated out before because of the TypeCheckingMode.SKIP, which is no longer needed.

brunobowden:
projectExec improvements merged in. I did a heck of a lot of testing and found a few historical bugs in the process. Curiously, JUnit and xcrun write their errors to stdout!?

For future reference, here's how you can reproduce a command line failures in all the tasks:

j2objcConfig {
    cycleFinderArgs '--nonsense'
    translateArgs '--nonsense'
    translateJ2objcLibs = []
    testArgs '--nonsense'
}
j2objcPackLibrariesDebug {
    buildType 'INVALID'
}

brunobowden:
Fixes #267

PR #295, opened 2015-07-18
projectExec regex match for unexpected output

brunobowden:
If the command line succeeds but the task regex match fails, e.g. can’t find (OK XX tests) for example, then the error is missing the command line, stdout or stderr. This change pushes the regex match in to projectExec, so it can produce more meaningful error messages. This duplicates the regex match done in the task itself but this is a small tradeoff.

projectExec
- Intercept all exceptions so that they all get useful error messages
- matchRegexOutputsRequired parameter
- Distinguish between command line failure or regex failure

Tasks
- Rework exceptions and error messages to be more descriptive
- Reworked CycleFinderTask exception handling

TESTED=added test: projectExec_MatchRegexFailed

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/CycleFinderTask.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/CycleFinderTask.groovy
@@ -127,9 +127,10 @@ class CycleFinderTask extends DefaultTask {
 
         ByteArrayOutputStream stdout = new ByteArrayOutputStream()
         ByteArrayOutputStream stderr = new ByteArrayOutputStream()
+        String cyclesFoundRegex = /(\d+) CYCLES FOUND/
         try {
             logger.debug('CycleFinderTask - projectExec:')
-            Utils.projectExec(project, stdout, stderr, {
+            Utils.projectExec(project, stdout, stderr, cyclesFoundRegex, {
                 executable cycleFinderExec
                 windowsOnlyArgs.each { String windowsOnlyArg ->
                     args windowsOnlyArg
@@ -157,11 +158,18 @@ class CycleFinderTask extends DefaultTask {
         } catch (Exception exception) {
             // ExecException is converted to InvalidUserDataException in Utils.projectExec(...)
 
-            // Suppress rethrow of exception if it's expected non-zero exit matching "CYCLES FOUND"
-            String cyclesFoundStr = Utils.matchRegexOutputStreams(stdout, stderr, /(\d+) CYCLES FOUND/)
-            if (!(exception instanceof InvalidUserDataException) || !cyclesFoundStr?.isInteger()) {
-                // Didn't match (XX CYCLES FOUND)
+            if (!Utils.isProjectExecNonZeroExit(exception)) {
                 throw exception
+            }
+
+            String cyclesFoundStr = Utils.matchRegexOutputs(stdout, stderr, cyclesFoundRegex)
+            if (!cyclesFoundStr?.isInteger()) {
+                String message =
+                        exception.toString() + '\n' +
+                        "CycleFinder completed could not find expected output.\n" +
+                        "Failed Regex Match cyclesFoundRegex: /$cyclesFoundRegex/" +
+                        "Found: $cyclesFoundStr"
+                throw new InvalidUserDataException(message, exception)
             }
 
             // Matched (XX CYCLES FOUND), so assert on cyclesFound
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
@@ -23,6 +23,7 @@ import groovy.transform.stc.ClosureParams
 import groovy.transform.stc.SimpleType
 import groovy.util.logging.Slf4j
 import org.gradle.api.InvalidUserDataException
+import org.gradle.api.Nullable
 import org.gradle.api.Project
 import org.gradle.api.file.CopySpec
 import org.gradle.api.file.FileCollection
@@ -204,15 +205,23 @@ class Utils {
     }
 
     // Matches regex, return first match as string, must have >1 capturing group
-    // Compares against stderr first then stdout
-    static String matchRegexOutputStreams(
-            ByteArrayOutputStream stdout, ByteArrayOutputStream stderr, String regex) {
+    // Return first capture group, comparing stderr first then stdout
+    // Returns null for no match
+    static String matchRegexOutputs(
+            ByteArrayOutputStream stdout,
+            ByteArrayOutputStream stderr,
+            @Nullable String regex) {
+
+        if (regex == null) {
+            return null
+        }
 
         Matcher stdMatcher = (stdout.toString() =~ regex)
         Matcher errMatcher = (stderr.toString() =~ regex)
         // Requires a capturing group in the regex
-        assert 1 <= stdMatcher.groupCount(), "regex must have '(...)' capturing group: $regex"
-        assert 1 <= errMatcher.groupCount(), "regex must have '(...)' capturing group: $regex"
+        String assertFailMsg = "matchRegexOutputs must have '(...)' capture group, regex: '$regex'"
+        assert stdMatcher.groupCount() >= 1, assertFailMsg
+        assert errMatcher.groupCount() >= 1, assertFailMsg
 
         if (errMatcher.find()) {
             return errMatcher.group(1)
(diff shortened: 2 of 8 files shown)


brunobowden:
@confile - this should provide better help in debugging #261
@advayDev1 - this is the final bit of work that I wanted to do in improving the command line error messages. Let me know what you think of them and feel free to make style changes to it.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/CycleFinderTask.groovy
```diff
@@ -157,13 +159,23 @@ class CycleFinderTask extends DefaultTask {
         } catch (Exception exception) {
             // ExecException is converted to InvalidUserDataException in Utils.projectExec(...)
 
-            // Suppress rethrow of exception if it's expected non-zero exit matching "CYCLES FOUND"
-            String cyclesFoundStr = Utils.matchRegexOutputStreams(stdout, stderr, /(\d+) CYCLES FOUND/)
-            if (!(exception instanceof InvalidUserDataException) || !cyclesFoundStr?.isInteger()) {
-                // Didn't match (XX CYCLES FOUND)
+            if (!(exception instanceof InvalidUserDataException) ||
+                // Should this be a better test to isolate non-zero exits?
```
advayDev1: would prefer to put this check in its own method say in Utils, like didProcessNonZeroExit or similar

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/UtilsTest.groovy
```diff
@@ -273,6 +276,37 @@ class UtilsTest {
...
+        try {
+            Utils.projectExec(proj, stdout, stderr, matchRegexOutputs, {
+                executable 'echo'
+                args 'echo-stdout'
+                setStandardOutput stdout
+                setErrorOutput stderr
+            })
+        } catch (InvalidUserDataException exception) {
+            String expected =
+                    'org.gradle.api.InvalidUserDataException: Command Line Succeeded (failure cause listed below):\n' +
```
advayDev1: do we really need to test the exact full string in the exception?
what is the minimal high priority information to get from the exception text?

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
```diff
@@ -226,46 +231,56 @@ class Utils {
...
     // TODO: In Gradle 2.5, we can switch to strongly-typed Actions, like:
     // https://docs.gradle.org/2.5/javadoc/org/gradle/api/Project.html#copy(org.gradle.api.Action)
     @CompileStatic(TypeCheckingMode.SKIP)
     static ExecResult projectExec(
-            Project proj,
-            ByteArrayOutputStream stdout,
-            ByteArrayOutputStream stderr,
+            Project proj, ByteArrayOutputStream stdout, ByteArrayOutputStream stderr,
+            // Fail if outputs don't match this regex
+            String matchRegexOutputsRequired,
```
advayDev1: provide an @Nullable annotation (or whichever flavor we link in), and comment as to null's meaning
please use javadoc for this method's params, it is now pretty complicated.  be very clear what "outputs" (plural) means when matched against a regex.  are the stdout and stderr interleaved (nearly impossible for us to do due to buffering)?  is it just the stdout only? etc.

you should however keep my above comments as non-javadoc since they are implementation-only.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
```diff
@@ -226,46 +231,56 @@ class Utils {
...
-
-            if (execSpec == null) {
-                throw exception
+            execSucceeded = true
+            if (matchRegexOutputsRequired) {
+                if (!matchRegexOutputs(stdout, stderr, matchRegexOutputsRequired)) {
+                    // Exception thrown here to output command line
+                    throw new InvalidUserDataException(
+                            'Unable to find expected expected output in stdout or stderr\n' +
+                            "Failed Regex Match: /$matchRegexOutputsRequired/")
```
advayDev1: are you sure the `/BLAH/` is valid here?  Isn't the escaping of strings within slashes slightly different?  would we need to properly escape it here to be precise?

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/CycleFinderTask.groovy
brunobowden: Moved to Utils. The test could be much better but I see your point in doing this.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
```diff
@@ -205,14 +205,19 @@ class Utils {
...
+            return null
+        }
+
         Matcher stdMatcher = (stdout.toString() =~ regex)
         Matcher errMatcher = (stderr.toString() =~ regex)
         // Requires a capturing group in the regex
-        assert 1 <= stdMatcher.groupCount(), "regex must have '(...)' capturing group: $regex"
-        assert 1 <= errMatcher.groupCount(), "regex must have '(...)' capturing group: $regex"
+        String assertFailMsg = "matchRegexOutputs must have '(...)' capture group, regex: '$regex'"
+        assert 1 <= stdMatcher.groupCount(), assertFailMsg
```
advayDev1: the number 1 before the groupCount is unnatural to me at least.
groupCount >= 1 please

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
```diff
@@ -205,14 +205,19 @@ class Utils {
 
     // Matches regex, return first match as string, must have >1 capturing group
```
advayDev1: indicate meaning of regex null and return value of null.

also @Nullable the regex

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/CycleFinderTask.groovy
advayDev1: yeah once we make it better in the future only one place to update

advayDev1:
LGTM with comments.  nothing i feel strongly above, so feel free to merge at your discretion.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/UtilsTest.groovy
brunobowden: From @confile's #261 bug, the command line succeeded but the plugin couldn't find the (OK X tests) in the output. In that case, you still want to output the full command line as it could be important. Perhaps it could hide the `org.gradle.api.InvalidUserDataException: ` but then it's a confusing code path where it suppresses some type information but not others. Feel free to do a PR if you have specific thoughts about this.

advayDev1: that was not what my comment was about.  i was asking whether you really want to do a full string match to the exceptions's text in this unit test.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
brunobowden: Good point. Added escaping for all the regexes. Please note if I've missed anything.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
brunobowden: done for both

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
brunobowden: groupCount >= 1

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
brunobowden: See new javadoc and comments

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/UtilsTest.groovy
brunobowden: This should be pretty stable for now. I'll change if it becomes a hassle.

brunobowden:
PTAL - need to fix one test but code should be in shape

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
advayDev1: hmm, comparing
http://docs.groovy-lang.org/latest/html/documentation/#_slashy_string
to
http://docs.groovy-lang.org/latest/html/api/groovy/json/StringEscapeUtils.html#escapeJava(java.lang.String)

it seems Slashy strings need forward slashes escaped, but escapeJava will not do that.  am i reading that right?  i could not find any SlashyStringEscaper for lack of a better word easily.

advayDev1:
one comment regarding escaping.

brunobowden:
Looking at the regex output, the right is no escaping... since the original string is `/regex.*\n/`, it's shown that way on logger.debug or in string comparison. So no escaping is required.

advayDev1:
no escaping *except* of forward slashes.  you need to escape those.

brunobowden:
I'm pushing now so that @confile can try it out soon in Germany. Everything appears correct to me.

I've removed the escaping and it works as I'd expect it to. The regex string is escaped before it's passed to the Matcher. So this test passes:

screen shot 2015-07-19 at 12 39 50 am

When the unit tests fail, it produces helpful output:

screen shot 2015-07-19 at 12 43 09 am

I'm not sure what you mean by "no escaping except of forward slashes". When I try to escape '/', it fails:

screen shot 2015-07-19 at 12 41 30 am

advayDev1:
Per http://docs.groovy-lang.org/latest/html/documentation/#_slashy_string
the only char needing escaping within such a string is a forward slash.  therefore when your regex has a forward slash, and you output in a message indicating it could be in a slashy string, you should replace
`/` with `\/`

PR #300, opened 2015-07-21
Fix "library not found for -lPods-TARGET"

brunobowden:
@advayDev1 - this didn't fix the issue. Any thoughts on what else might be needed?


advayDev1:
Actually, I should ask:
@brunobowden  @confile - has the j2objcXcode task ever fully worked - requiring no hand modifications - end 2 end?  does someone know what commit that was at?

advayDev1:
Ok, so further investigation (http://www.raywenderlich.com/64546/introduction-to-cocoapods-2) indicates there is a *separate* Pods project.  You have to open the xcworkspace first.

confile:
@advayDev1 Yes it works with the old plugin. There it was the j2objcPod task which I am currently using.

PR #301, opened 2015-07-21
Caused By => Cause

brunobowden:
- Helps with debugging command line failures


advayDev1:
shrug LGTM.
what effect does this have btw?

brunobowden:
I was confused by some of @confile's output and wanted to distinguish between "Caused By" from the exception stacktrace and the error string that the plugin creates.

Issue #306, opened 2015-07-21
After clean build, test binary tasks incorrectly think they are UP-TO-DATE


advayDev1:
the :debugTestJ2objcExecutable task does not recreate the binary after cleaning, if there are no tests.

do a ./gradlew clean build, you will see for example:
```
flatbuffers-java:compileDebugTestJ2objcExecutableTestJ2objcObjc UP-TO-DATE
:flatbuffers-java:compileX86_64DebugFlatbuffers-java-j2objcStaticLibraryFlatbuffers-java-j2objcObjc
:flatbuffers-java:createX86_64DebugFlatbuffers-java-j2objcStaticLibrary
:flatbuffers-java:x86_64DebugFlatbuffers-java-j2objcStaticLibrary
:flatbuffers-java:linkDebugTestJ2objcExecutable UP-TO-DATE
:flatbuffers-java:debugTestJ2objcExecutable UP-TO-DATE
...
:flatbuffers-java:j2objcTestDebug FAILED

FAILURE: Build failed with an exception.

* What went wrong:
A problem was found with the configuration of task ':flatbuffers-java:j2objcTestDebug'.
> File '/Users/X/third_party/flatbuffers-java/build/binaries/testJ2objcExecutable/debug/testJ2objc' specified for property 'testBinaryFile' does not exist.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.


```

advayDev1:
Ah doing:
./gradlew clean build
does not work.
doing
./gradlew clean
./gradlew build

will work.  the problem is the EmptyTest.m file.  It is created during project evaluation, not task execution.  So the clean step deletes that file since it is in the build/ directory.

Not sure how to fix this in the gradle 2.4 model...  removing from 0.4

brunobowden:
Tag it on the beta release to remind us later?

brunobowden:
I've not seen this issue and I do `./gradlew clean build` a good deal.

advayDev1:
do all of your j2objc projects have at least one *Test.ajva file?

brunobowden:
Yes. So that's why I don't see it. Doesn't the flatbuffers project have
tests?

advayDev1:
it has integration tests in a separate directory, but not unit tests with the java runtime.

PR #307, opened 2015-07-21
FAQ: clean build re: #306

advayDev1:


confile:
LGTM

FAQ.md
```diff
@@ -73,6 +73,9 @@ j2objcConfig {
 }
 ```
 
+### Why is my clean build failing?
+This is a [known issue](https://github.com/j2objc-contrib/j2objc-gradle/issues/306).  If you are doing `./gradlew clean build`, try instead `./gradlew clean && ./gradlew build`.  The plugin creates some files during project configuration that the `build` needs but `clean` deletes.
```
brunobowden: Add mention explaining that this doesn't happen when you have any Test.java files and why that is.

brunobowden:
LGTM with comment

FAQ.md

PR #308, opened 2015-07-22
Podspec: split debug / release, remove source

### LINKED ISSUE #127: Debug and Release targets with CocoaPods
@confile - can you take a look at this, I think there's a path to do what's needed here and I'd appreciate you help on this.

The LIBRARY_SEARCH_PATHS should be different for Debug and Release builds in Xcode. Currently it only points to the debug version. The following is based on: https://github.com/CocoaPods/CocoaPods/issues/2847

Podfile changes:

target 'TapTap' do
// You don't need to add targeting for arm64 or x86 for the simulator as the linker will automatically choose the right architecture from the fat binary
pod 'j2objc-base-debug', :path => "${project.buildDir}", :configurations => ['Debug']
pod 'j2objc-base-release', :path => "${project.buildDir}", :configurations => ['Release', 'AppStore']
end

Two podspec files: 'j2objc-base' and 'j2objc-base-debug'. The differences are fairly minor:

Debug:
'LIBRARY_SEARCH_PATHS' => '${getJ2ObjCHome()}/lib ${project.buildDir}/j2objcOutputs/lib/iosDebug'

Release:
'LIBRARY_SEARCH_PATHS' => '${getJ2ObjCHome()}/lib ${project.buildDir}/j2objcOutputs/lib/iosRelease'

brunobowden:
- Proper debug and release versions of the podspec
- Remove all source from podspec
- Podspec style to match documentation examples

Fixes #127

TESTED=yes
Normal Xcode build is debug. Release build is done using “Product > Build For > Profiling”. Tested by hiding libj2objc-shared and then looking at the command line when the build failed.

“Profiling” uses iosRelease:
-L/Users/brunobowden/dev/taptap/base/build/j2objcOutputs/lib/iosRelease

Everything else uses iosRelease:
-L/Users/brunobowden/dev/taptap/base/build/j2objcOutputs/lib/iosDebug


brunobowden:
@advayDev1, @confile, @alexmetcalfe - PTAL

brunobowden:
Does anyone use ClassLoader at the moment? I've never used it and I'm not sure if it works at all. If someone is using that right now and it is working, then perhaps the switch to the resouce_bundles might break it. Need to look in to this further.

brunobowden:
FYI, ClassLoader is how you load files with Java from the `src/main/resources` directory and similar locations.

advayDev1:
Yes, I'm using the class loader indirectly through resource loading.  Will give this a try tomorrow...

brunobowden:
resouce_bundle's are **strongly recommended" partly to prevent collisions. I presume this is done by putting it in subdirectories, which would break the classloader. I'm not sure how to do isolation between multiple projects that are linked in to the same iOS app.

advayDev1:
On the Android, I don't think any separation occurs.  Multiple Java/Android projects all get their resources mixed into the root dir of the APK.

brunobowden:
I've reverted the resouce_bundles part of this change until I'm at a point that I can test this end to end.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy
```diff
@@ -90,30 +93,27 @@ class XcodeTask extends DefaultTask {
...
+
+        String podspecContentsDebug =
+                genPodspec(getPodNameDebug(), libDirDebug, libName, getJ2objcHome(),
+                        srcGenDirRelativeToBuildDir, "$j2objcResourceDirName/**/*")
+        String podspecContentsRelease =
+                genPodspec(getPodNameRelease(), libDirRelease, libName, getJ2objcHome(),
+                        srcGenDirRelativeToBuildDir, "$j2objcResourceDirName/**/*")
+
+        logger.debug("Writing debug podspec... ${getPodspecDebug()}\n$podspecContentsDebug")
+        logger.debug("Writing release podspec... ${getPodspecDebug()}\n$podspecContentsDebug")
```
advayDev1: interleave the logs and the write statements so that an error during writing debug does not show a Writing release podspect log line before it

brunobowden: Done

advayDev1:
LGTM with 1 comment

PR #311, opened 2015-07-22
MockExec (plugin class) => ExecHandleBuilder (internal Gradle class)

brunobowden:
- Logging of workingDir in projectExec

Even though this uses an internal Gradle class, I think it is a better trade off than the time and risks of re-implementing the class within the plugin.

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
@@ -299,51 +299,46 @@ class Utils {
             }
 
         } catch (Exception exception) {
-
             // ExecException is most common, which indicates "non-zero exit"
-            // Add command line and stderr to make the error message more useful
-            // Chain to the original ExecException for complete stack trace
-            String exceptionMsg = ''
-            if (execSucceeded) {
-                exceptionMsg += 'Command Line Succeeded (failure cause listed below):\n'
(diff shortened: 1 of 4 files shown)


brunobowden:
@confile, @advayDev1 - PTAL

confile:
LGTM

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
```diff
@@ -299,51 +299,46 @@ class Utils {
...
-            return
+    static String projectExecLog(
+            ExecSpec execSpec, ByteArrayOutputStream stdout, ByteArrayOutputStream stderr,
+            boolean execSucceeded, Exception exception) {
+        // Add command line and stderr to make the error message more useful
+        // Chain to the original ExecException for complete stack trace
+        String msg = ''
+        // The command line can be long, so highlight more important details below
+        if (execSucceeded) {
+            msg += 'Command Line Succeeded:\n'
```
advayDev1: why init to empty and add here?

advayDev1:
lgtm

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
brunobowden: Fixed

PR #315, opened 2015-07-23
Warning to setup bridging-header for Swift development

brunobowden:
- Warns on setting up bridging header alongside warning about opening .xcworkspace

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -100,6 +100,24 @@ sourceSets {
 ```
 
 
+### How do I develop with Swift?
+
+To work with Swift in Xcode, you need to configure a [bridging header](https://developer.apple.com/library/ios/documentation/Swift/Conceptual/BuildingCocoaApps/MixandMatch.html#//apple_ref/doc/uid/TP40014216-CH10-XID_81).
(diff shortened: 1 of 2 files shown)


brunobowden:
@advayDev1, @confile - PTAL

FAQ.md
```diff
@@ -100,6 +100,24 @@ sourceSets {
...
+Within that bridging header, include the file needed for using the JRE and any classes that you'd like
+to access from Swift code.
+
+```
+// File: IOS-APP-bridging-header.h
+
+// Required for Swift initialization of Java objects (they all inherit from JavaObject)
+#import "JreEmulation.h"
+
+// Swift accessible j2objc translated classes, referenced from `build/j2objcSrcGen`
```
advayDev1: no, they should be references solely from j2objcOutputs/src (not sure what the full path is).
j2objcSrcGen is our intermediate implementation detail

advayDev1:
LGTM with comment

FAQ.md
brunobowden: Changed comment to `build/j2objcOutputs/src/main/objc` - The podspec is pointing to the wrong place, so I'll fix that in another commit.

PR #318, opened 2015-07-24
Assemble and Pack task testing

brunobowden:
- AssembleSourceTask unit tests
- AssembleLibrariesTask unit tests
- PackLibrariesTask unit tests
- MockProjectExec added demandCopyAndReturn(CopySpec closure)

PR #320, opened 2015-07-24
Xcode to use j2objcOutputs + AssembleResourcesTask

### LINKED ISSUE #319: Xcode should read resources from j2objcOutputs subdirectory
Resources should be isolated to a subdirectory of j2objcOutputs. Currently they're being read from:

    s.resources = 'j2objcResources/**/*'

brunobowden:
Please review PR #318 first

- Move Xcode resource loading to j2objcOutputs, fixes #319
- Added AssembleResourcesTask to copy resources to j2objcOutputs
- ‘File’ suffix when @Input or @Output is File rather than string

J2objcConfig path refactor
- Users set directories now point to src/main rather than src/main/objc
- destSrcDir => destSrcMainDir
- destSrcDirTest => destSrcTestDir
- Explicit directory methods for accessing Main/Test, Objc/Resources:
  i.e. j2objcConfig.getDestSrc{Main|Test}{Objc|Resources}Dir()


advayDev1:
i think you need to come up with a convention for the InputFiles naming as well, otherwise it doesn't come off as consistent

advayDev1:
LGTM after that

brunobowden:
@advayDev1 - the distinction I think of is returning a File rather than a path. To that end, I've marked any @Input* returning a file with "File" as a method suffix. Please see what I've done but I don't think it needs to be more than that.

For the J2objcConfig output paths, they all now use project.file and File to produce a safer and better result. Using these methods also prevents a task from forgetting to call project.file to resolve a path for example.

For @OutputDirectory, though some testing, it re-runs the task if the directory path or contents have changed. That allowed me to removed some redundant @Input methods for the output directory path. This was then documented in a Gradle PR: https://github.com/gradle/gradle/pull/485

brunobowden:
@advayDev1 - please take a look again. I think this is in a much better place thanks to your input.

brunobowden:
@advayDev1 - I've now merged the changes. Feel free to push back on anything.

advayDev1:
sorry, I already gave an LGTM above: https://github.com/j2objc-contrib/j2objc-gradle/pull/320#issuecomment-124771494

PR #322, opened 2015-07-24
Exception for invalid j2objc property keys

brunobowden:
- Warn if local.properties contains invalid key with ‘j2objc.’ prefix
- Warn if developer requests invalid key
- Correct unit tests referring to j2objc.debug.enabled

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
@@ -62,7 +62,25 @@ class Utils {
         }
     }
 
+    // Add valid keys here
+    // Use camelCase and order alphabetically
+    private static final List<String> PROPERTIES_VALID_KEYS =
+            Collections.unmodifiableList(Arrays.asList(
+        'debug.enabled',
+        'home',
+        'release.enabled'
+    ))
+
+    private static final String PROPERTY_KEY_PREFIX = 'j2objc.'
+
     static String getLocalProperty(Project proj, String key, String defaultValue = null) {
+
+        // Check for requesting invalid key
+        if (!(key in PROPERTIES_VALID_KEYS)) {
+            throw new InvalidUserDataException(
+                    "Requesting invalid property: $key\n" +
+                    "Valid Keys: $PROPERTIES_VALID_KEYS")
+        }
         File localPropertiesFile = new File(proj.rootDir, 'local.properties')
         String result = defaultValue
         if (localPropertiesFile.exists()) {
@@ -70,7 +88,22 @@ class Utils {
             localPropertiesFile.withInputStream {
                 localProperties.load it
             }
-            result = localProperties.getProperty('j2objc.' + key, defaultValue)
+
+            // Check valid key in local.properties for everything with PROPERTY_KEY_PREFIX
+            localProperties.keys().each { String propKey ->
+                if (propKey.startsWith(PROPERTY_KEY_PREFIX)) {
+                    String withoutPrefix =
+                            propKey.substring(PROPERTY_KEY_PREFIX.length(), propKey.length())
+                    if (!(withoutPrefix in PROPERTIES_VALID_KEYS)) {
+                        throw new InvalidUserDataException(
+                                "Invalid j2objc property: $propKey\n" +
+                                "From local.properties: $localPropertiesFile.absolutePath\n" +
+                                "Valid Keys: $PROPERTIES_VALID_KEYS")
+                    }
+                }
+            }
+
+            result = localProperties.getProperty(PROPERTY_KEY_PREFIX + key, defaultValue)
         }
         return result
     }
--- a/src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/UtilsTest.groovy
+++ b/src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/UtilsTest.groovy
@@ -98,6 +98,34 @@ class UtilsTest {
     }
 
     @Test
+    public void testGetLocalProperty_RequestInvalidKey() {
+        File localProperties = proj.file('local.properties')
+        localProperties.write('#IGNORE')
+
+        expectedException.expect(InvalidUserDataException.class)
+        expectedException.expectMessage('Requesting invalid property: requested-invalid-key')
+        // Check list of valid keys is suggested by checking for a single entry:
+        expectedException.expectMessage('debug.enabled')
+
+        Utils.getLocalProperty(proj, 'requested-invalid-key')
+    }
+
+    @Test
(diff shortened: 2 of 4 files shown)


brunobowden:
@advayDev1, @confile - I'm sure this will save someone from the pain of setting the wrong key. The one downside is that switching between version of the plugin you may need to comment out keys that are invalid in older versions. You could add a "j2objc.warnOnlyInvalidKey" but I think that's unnecessary.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
```diff
@@ -62,7 +62,23 @@ class Utils {
         }
     }
 
+    // Add valid keys here
+    // Use camelCase, no '.' and order alphabetically
+    static String[] PROPERTIES_VALID_KEYS = [
```
advayDev1: final ImmutableList?

this static variable can be mutated

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
```diff
@@ -70,7 +86,20 @@ class Utils {
             localPropertiesFile.withInputStream {
                 localProperties.load it
             }
-            result = localProperties.getProperty('j2objc.' + key, defaultValue)
+
+            // Check local.properties for j2objc.INVALID_KEY
+            localProperties.keys().each { String propKey ->
+                if (propKey.startsWith(PROPERTY_KEY_PREFIX)) {
+                    if (!(propKey.replace(PROPERTY_KEY_PREFIX, '') in PROPERTIES_VALID_KEYS)) {
```
advayDev1: more accurate would be
```
if (!(propKey.substring(PROPERTY_KEY_PREFIX.length(), propKey.length) in PROPERTIES_VALID_KEYS)) {
```

it is possible j2objc. could show up twice and you already checked startsWith

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/UtilsTest.groovy
```diff
@@ -66,38 +66,66 @@ class UtilsTest {
 
     @Test
     public void testGetLocalProperty_FileNotPresent() {
-        String value = Utils.getLocalProperty(proj, 'debug.enabled')
+        String value = Utils.getLocalProperty(proj, 'debugEnabled')
```
advayDev1: please revert the debug|release.enabled --> debug|releaseEnabled changes.
i can very much see us doing things like debug.archs or release.archs, and I think a hierarchical configuration like that allows us to easily pull out common code to for example, parse the 'debug' options congruently to parsing the 'release' options with a single method

advayDev1:
comments provided

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
brunobowden: => final ImmutableList

Didn't realize it was so easy to include Guava. Do we use Guava much elsewhere?

brunobowden: Well, we've been using @VisibleForTesting for a while now.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
brunobowden: Changed to:

    String withoutPrefix =
            propKey.substring(PROPERTY_KEY_PREFIX.length(), propKey.length())
    if (!(withoutPrefix in PROPERTIES_VALID_KEYS)) {

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/UtilsTest.groovy
brunobowden: With my other changes, it will throw `Requesting invalid property: debug.Enabled`. I'd suggest leaving it as is... then when we add hierarchical options later on, then we can revisit this.

brunobowden:
Need LGTM to skip your comment about doing the reversion debugEnabled => debug.Enabled.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
advayDev1: heh, Gradle brings in dozens if not hundreds of libraries ;)

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/UtilsTest.groovy
advayDev1: having architectures configurable is a change I'm currently working on.
the reason I don't think we can delay that is that the client-facing API changes repeatedly.
in that users currently have to switch from debug.enabled to debugEnabled, and then once architecture configuration goes in have to witch to debug.enabled again

advayDev1:
i think that comment is important, please see reply

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/UtilsTest.groovy
brunobowden: What would you suggest as an alternative? Add "debug.Enabled" as a valid key... or add "debug.*" to match any of that? Not sure how to preserve the checks I've added in a way that also serves your needs.

advayDev1: all i'm suggesting is change each "debugEnabled" string back to "debug.enabled"
you don't need to do the support for debug.* or hierarchy in this change at all, sorry for the confusion

brunobowden:
Now I get a `NoClassDefFoundError` for ImmutableList. I'm wary about this as I don't want adding a dependency to bite us when we release this.

```
Caused by: java.lang.NoClassDefFoundError: Could not initialize class com.github.j2objccontrib.j2objcgradle.tasks.Utils
	at java.lang.Class.forName0(Native Method)
	at java.lang.Class.forName(Class.java:270)
        ...

* Where:
Build file '/Users/brunobowden/dev/taptap/base/build.gradle' line: 86

* What went wrong:
A problem occurred evaluating project ':base'.
> com/google/common/collect/ImmutableList
```

brunobowden:
Switched from ImmutableList to...

    final List<String> PROPERTIES_VALID_KEYS =
            Collections.unmodifiableList(Arrays.asList(...

brunobowden:
@advayDev1 - please look again. I think this is what you were asking for.

advayDev1:
LGTM.

PR #324, opened 2015-07-25
projectClearDir

brunobowden:
- Allows mocking of calls to project.mkdir(…)
- Currently only used in —build_closure, which isn’t tested
- Available for testing in the future


brunobowden:
@advayDev1 - this isn't helpful for now unless you add unit testing for build_closure in the future.

advayDev1:
lgtm

PR #325, opened 2015-07-25
CompileStatic for J2objcConfig and UtilsTest

brunobowden:
- Uncovered one mistake in a test

### DIFF
--- a/src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/UtilsTest.groovy
+++ b/src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/UtilsTest.groovy
@@ -16,6 +16,7 @@
 
 package com.github.j2objccontrib.j2objcgradle.tasks
 
+import groovy.transform.CompileStatic
 import org.apache.commons.io.output.TeeOutputStream
 import org.gradle.api.InvalidUserDataException
 import org.gradle.api.Project
@@ -33,6 +34,7 @@ import org.junit.rules.ExpectedException
 /**
  * Utils tests.
  */
+@CompileStatic
 class UtilsTest {
 
     private Project proj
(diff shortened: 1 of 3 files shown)


src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/UtilsTest.groovy
```diff
@@ -182,7 +184,7 @@ class UtilsTest {
                 'src/test/resources/com/github/j2objccontrib/j2objcgradle/tasks/prefixes.properties')
 
         List<String> translateArgs = new ArrayList<String>()
-        translateArgs.add("--prefixes ${prefixesProp.absolutePath}")
+        translateArgs.add('--prefixes ' + prefixesProp.absolutePath)
```
advayDev1: i don't understand this change.

brunobowden: It's to satisfy @CompileStati which wants a String instead of a GString or
is that GroovyString?

advayDev1: ah when we have that, the right things is do to "$WHATEVER".toString()
we don't want to stop the use of GStrings

brunobowden: I've done that in a number of other places already. In this case, as it was only simple concatenation, I thought it was clearer. Compare the two:

    '--prefixes ' + prefixesProp.absolutePath
    "--prefixes ${prefixesProp.absolutePath}".toString()

advayDev1: shrug.  it doesn't really matter to me.

advayDev1:
lgtm

PR #326, opened 2015-07-25
Split debug and test task output directories

brunobowden:
Please review PR #320 first.

Now writes to two separate directories:
- j2objcTest/debug
- j2objcTest/release
- buildType @Input for TestTask

Consolidate resource copying to `syncResources` method. This should
be replaced with a more efficient copying system, e.g. rsync.

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
@@ -333,10 +333,17 @@ class Utils {
                (exception?.getCause() instanceof ExecException)
     }
 
-    static WorkResult copyResources(Project proj, String sourceSetName, File destDir) {
+    // Sync main and or test resources, deleting destination directory and then recreating it
+    // TODO: make the sync more efficient, e.g. Gradle Sync task or rsync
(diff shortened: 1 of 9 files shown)


brunobowden:
@advayDev1, @confile - PTAL

confile:
LGTM

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
```diff
@@ -333,10 +333,17 @@ class Utils {
                (exception?.getCause() instanceof ExecException)
     }
 
-    static WorkResult copyResources(Project proj, String sourceSetName, File destDir) {
+    // Sync main and or test resources, deleting destination directory and then recreating it
+    // TODO: make the sync more efficient, e.g. Gradle Sync task or rsync
+    static WorkResult syncResources(Project proj, List<String> sourceSetNames, File destDir) {
```
advayDev1: i think 'syncResourcesTo' would be a better name.  the sync is one-way, to destDir alone.

advayDev1:
lgtm iwth comment

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
brunobowden: Changed to syncResourcesTo

PR #329, opened 2015-07-26
j2objc => J2ObjC for user facing messages and docs

brunobowden:
- Settle on the proper name ‘J2ObjC for the tool and plugin.
- Rewrapped some line to fit with 100 character limit


advayDev1:
lgtm

PR #330, opened 2015-07-26
Minimum Gradle Version check: 2.4

brunobowden:
- Throws exception for version 2.3 and earlier

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
@@ -34,6 +34,7 @@ import org.gradle.api.tasks.WorkResult
 import org.gradle.process.ExecResult
 import org.gradle.process.ExecSpec
 import org.gradle.process.internal.ExecException
+import org.gradle.util.GradleVersion
 
 import java.util.regex.Matcher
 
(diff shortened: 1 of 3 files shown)


brunobowden:
@advayDev1, @confile - I think this is a good precautionary check to have in the plugin. I've tested it from the command line as well.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
```diff
@@ -48,6 +48,17 @@ class Utils {
     // TODO: ideally bundle j2objc binaries with plugin jar and load at runtime with
     // TODO: ClassLoader.getResourceAsStream(), extract, chmod and then execute
 
+    // Prevent construction of this class, confines usage to static methods
+    private Utils() { }
+
+    static void checkMinGradleVersion(String gradleVersion) {
```
advayDev1: TODO will fail when they get to gradle 10

advayDev1:
lgtm.  you're introducing a y2k =)

brunobowden:
Just for you @advayDev1, I search for the solution to our Y2K problem. There is a DefaultArtifactVersion from Apache... but I didn't want to introduce a new dependency. Then I found the GradleVersion class... which very neatly encapsulates what we need. Now version "10.0" is passing in the newly added test.

PR #338, opened 2015-08-02
j2objcCycleFinder as j2objcTest dependency instead of j2objcTranslate

brunobowden:
The cycle_finder tool doesn’t need to be run in order for translation
and compilation. As a tool for code correctness, it should be run as a
dependency of j2objcTest instead.

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy
@@ -81,22 +81,11 @@ class J2objcPlugin implements Plugin<Project> {
                 description "Marker task for all tasks that must be complete before j2objc building"
             }
 
-            // j2objcCycleFinder is disabled by default as it's complex to use and understand.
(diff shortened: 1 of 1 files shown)


brunobowden:
@advayDev1, @confile - this will make the assemble task slightly faster.

advayDev1:
lgtm

brunobowden:
Tested on my main project and it works correctly.

What do you think the test ordering should be? Should we run the debug/release j2objc tests first or cycle_finder? My feeling is that whatever is more likely to fail, should be run first so that it can give useful feedback more quickly. In this case though, it should be relatively rare that either one fails... the java test have already passed and cycle_finder should change only rarely.

advayDev1:
i've never run cycle finder so i don't know how long it takes.  i think j2objc tests should fail very rarely and unless you are in power-user land probably indicate a bug or unfinished functionality in j2objc.  however, a rapidly changing program can easily introduce (legitimately) new reference loops requiring developer attention (either by waiving the loop as OK, or putting in an annotation somewhere).

so i think: cycleFinder, debug tests, release tests

brunobowden:
Committing as is then...

PR #343, opened 2015-08-10
OS specific File.pathSeparator

brunobowden:
- Use ":" as path separator for OS X, and ";" for Windows
- Fix unit tests to use ';' path separator on Windows

Minor changes:

- Minor improvements in README
- Corrected link in CONTRIBUTING
- Improve readability of a couple of unit tests

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -125,11 +125,12 @@ String message = "the count is negative"  // WRONG - only needed for $var interp
 ```
 
 * Regexes should be written and displayed as
-[http://docs.groovy-lang.org/latest/html/documentation/#_slashy_string](Slashy Strings):
+[Slashy Strings](http://docs.groovy-lang.org/latest/html/documentation/#_slashy_string):
 ```
-String regex = /dot-star:.*,forward-slash:\/,newline:\n/
+String regex = /dot-star:.*, forward-slash:\/, newline:\n/
 logger.debug('Regex is: ' + Utils.escapedSlashyString(regex))
-// Debug log: '/dot-star:.*,forward-slash:\/,newline:\n/'
+
+// Debug log: '/dot-star:.*, forward-slash:\/, newline:\n/'
 ```
 
 
--- a/README.md
+++ b/README.md
@@ -70,11 +70,11 @@ be run as follows:
 
     ./gradlew shared:build
 
-To update the existing Xcode project to load the libraries and header files:
+During development, to build the libraries and update Xcode (skipping the tests):
 
     ./gradlew shared:j2objcXcode
 
-Typically they should both be used together:
+For a complete build, run both:
 
     ./gradlew shared:build shared:j2objcXcode
 
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/CycleFinderTask.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/CycleFinderTask.groovy
@@ -124,7 +124,8 @@ class CycleFinderTask extends DefaultTask {
                 project.files(Utils.j2objcLibs(getJ2objcHome(), getTranslateJ2objcLibs()))
         ])
         // TODO: comment explaining ${project.buildDir}/classes
-        String classpathArg = Utils.joinedPathArg(classpathFiles) + ":${project.buildDir}/classes"
+        String classpathArg = Utils.joinedPathArg(classpathFiles) +
+                              File.pathSeparator + "${project.buildDir}/classes"
 
         ByteArrayOutputStream stdout = new ByteArrayOutputStream()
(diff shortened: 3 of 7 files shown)


brunobowden:
FYI: @advayDev1. This fix was suggested by another user who will also review it.

gocursor:
Hi Bruno! This is much better than my original proposal. I haven't built and tried it out on Windows yet, but I agree with all the changes...

BTW: Can I build the plugin on Windows? Are there any instructions on how a build can be made?

brunobowden:
@gocursor - I have no idea if the build will work but it's worth trying. Fork and download it, then follow the build instructions: https://github.com/j2objc-contrib/j2objc-gradle/blob/master/CONTRIBUTING.md#local-development

@advayDev1 - please review as well

gocursor:
Ok, I tried building on Windows and of course it can not work, because of "acceptable tradeoff" mentioned above... :-) Anyway: I could built it on Windows by turning off the tests...

Then I tried the plugin (under Windows) and the file separator works now so the translation phase is working under Windows and I get see the ObjectiveC output files.

The build task fails however when "xcrun" is run, because it is not present under Windows.

Question: Would it be possible to disable compilation, so I could use this plugin under Windows with "gradle build" and get the translated ObjectiveC without build error? In this case "xcrun" and other OSX specific commands for compilation whould be just skipped? Could you point me to the point in source code where I could prepare such switch or is maybe something similar already implemented?

gocursor:
Maybe off topic, but I will continue here: I could disable these tasks under Windows:

j2objcPackLibrariesDebug { enabled = false }
j2objcAssembleDebug { enabled = false }
j2objcPackLibrariesRelease { enabled = false }
j2objcAssembleRelease { enabled = false }

but I run into another error when plugin wants to run clang and I don't know how to disable this task...

Here is the error output:

:app-ios:j2objcPreBuild UP-TO-DATE
:app-ios:j2objcTranslate
:app-ios:j2objcAssembleSource
:app-ios:j2objcBuildObjcDebug UP-TO-DATE
:app-ios:j2objcPackLibrariesDebug SKIPPED
:app-ios:j2objcAssembleDebug SKIPPED
:app-ios:j2objcBuildObjcRelease UP-TO-DATE
:app-ios:j2objcPackLibrariesRelease SKIPPED
:app-ios:j2objcAssembleRelease SKIPPED
:app-ios:j2objcAssembleResources
:app-ios:j2objcAssemble
:app-ios:jar
:app-ios:assemble
:app-ios:j2objcCycleFinder SKIPPED
:app-ios:compileDebugTestJ2objcExecutableTestJ2objcObjc FAILED

* What went wrong:
Execution failed for task ':app-ios:compileDebugTestJ2objcExecutableTestJ2objcObjc'.
> No tool chain is available to build for platform 'x86_64':
    - Tool chain 'clang' (Clang): Could not find C compiler 'clang' in system path.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED

brunobowden:
You'll need to use a Mac in order to use xcrun and build any libraries or run the clang compiler. You don't need to disable commands, just run up to the limit of what the plugin can do on a PC:

./gradlew shared:j2objcAssembleSource shared:j2objcAssembleResources
    ...
    :base:j2objcPreBuild
    :base:j2objcTranslate
    :base:j2objcAssembleSource
    :base:j2objcAssembleResources

This should work for you. If it does, then please submit a PR to add something to CONTRIBUTING.md for "How do I build the plugin on Windows?" Also in the PR, add a separate item to the FAQ for "How do I develop on Windows?" You'll need to explain that only translation is supported - the tests aren't run and libraries aren't created. Also in both items, strongly suggest doing development on a Mac, which is much easier.

brunobowden:
@gocursor - I've updated the code so that it should now be able to run and pass all the unit tests on Windows. See the changes in the `MockProjectExec.groovy` file for doing this.

advayDev1:
@brunobowden - In terms of correctness, the code looks fine.

However:
Do we really need to make our supported development surface larger than OS X?
Either the system works or it doesn't, there's no half-way - and integration tests will not work on anything but a system supporting j2objc.

Specifically, I suggest we do not add anything about development on non-OS X to the FAQ or CONTRIBUTING, and instead preserve the minimum requirements of j2objc itself
(incl. "Mac OS X 10.9 or higher" - see http://j2objc.org/#requirements)  I don't think the benefit of non-OS X development is worth the can of works in terms of support it would entail.

gocursor:
@advayDev1: I just need the translation phase on Windows and maybe also cycle finder, so I can verify that my code would translate to ObjectiveC. This is my use case. You can however decide if you would like to support this use case.

gocursor:
@brunobowden: I have tried running all unit tests on Windows with the latest version: 82 tests completed, 35 failed... But "gradle build -x test" still works... If you will decide to support Windows, I can look into the details and fix it.

PR #344, opened 2015-08-12
Can I develop on Windows - FAQ new item

brunobowden:
- Explain that Windows is officially unsupported by may work
- PackLibrariesTask to fail with useful error message on Windows

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -235,3 +235,19 @@ source and whitelist. Note how this gives and expected cycles of zero.
 For more details:
 - http://j2objc.org/docs/Cycle-Finder-Tool.html
 - https://groups.google.com/forum/#!msg/j2objc-discuss/2fozbf6-liM/R83v7ffX5NMJ
+
+
+### Can I develop on Windows?
+
+Development is officially supported only on Mac OS X. This matches the
+[J2ObjC Requirements](http://j2objc.org/#requirements).
+In practice, the j2objcCycleFinder and j2objcTranslate tasks may work, though it is
+strongly suggested to use this for experimentation only and do all proper development
+on OS X. The team will reject any bugs related to systems that don't meet the
+[minimum requirements](README.md#minimum-requirements).
+
+To experiment on Windows (skipping the unsupported tasks), the following may work:
+
+    ./gradlew shared:j2objcAssembleSource shared:j2objcAssembleResources
+
+Please note that this will not build the packed libraries or update your Xcode project.
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/PackLibrariesTask.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/PackLibrariesTask.groovy
@@ -19,6 +19,7 @@ package com.github.j2objccontrib.j2objcgradle.tasks
 import com.github.j2objccontrib.j2objcgradle.J2objcConfig
 import groovy.transform.CompileStatic
 import org.gradle.api.DefaultTask
+import org.gradle.api.InvalidUserDataException
 import org.gradle.api.file.ConfigurableFileCollection
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.InputFiles
@@ -59,6 +60,12 @@ class PackLibrariesTask extends DefaultTask {
     void packLibraries() {
         Utils.projectDelete(project, getOutputLibDirFile())
         getOutputLibDirFile().mkdirs()
+
+        if (Utils.isWindows()) {
+            throw new InvalidUserDataException(
+                    'Windows is officially unsupported. The packLibraries task can only be run on Mac OS X:\n' +
+                    'https://github.com/j2objc-contrib/j2objc-gradle/blob/master/FAQ.md#can-i-develop-on-windows')
+        }
 
         ByteArrayOutputStream stdout = new ByteArrayOutputStream()
         ByteArrayOutputStream stderr = new ByteArrayOutputStream()
--- a/README.md
+++ b/README.md
@@ -53,6 +53,18 @@ Within the Android application's `android/build.gradle`, make it dependent on th
     }
 
 
+### Minimum Requirements
+
+Gradle 2.4 is required for the compilation support within Gradle. The other necessities
+are the [J2objc Requirements](http://j2objc.org/#requirements).
+
+    * Gradle 2.4
+    * JDK 1.7 or higher
+    * Mac workstation or laptop
+    * Mac OS X 10.9 or higher
+    * Xcode 6 or higher
+
+
 ### J2ObjC Installation
 
 Download the latest version from the [J2ObjC Releases](https://github.com/google/j2objc/releases).
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy
@@ -26,6 +26,7 @@ import com.github.j2objccontrib.j2objcgradle.tasks.TranslateTask
 import com.github.j2objccontrib.j2objcgradle.tasks.Utils
 import com.github.j2objccontrib.j2objcgradle.tasks.XcodeTask
 import org.gradle.api.DefaultTask
+import org.gradle.api.InvalidUserDataException
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.Task
@@ -42,6 +43,11 @@ class J2objcPlugin implements Plugin<Project> {
         // This avoids a lot of "project." prefixes, such as "project.tasks.create"
         project.with {
             Utils.checkMinGradleVersion(GradleVersion.current())
+
+            if (Utils.isWindows()) {
+                logger.warn('Windows is officially unsupported:\n' +
+                            'https://github.com/j2objc-contrib/j2objc-gradle/blob/master/FAQ.md#can-i-develop-on-windows')
+            }
 
             extensions.create('j2objcConfig', J2objcConfig, project)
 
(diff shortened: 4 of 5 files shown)


brunobowden:
@advayDev1, @gocursor - moving the conversation to a separate PR.

I've had several people that have asked for help running translation on Windows. Similar to this situation, it's only taken minor fixes for it to work. It does have some use for when people try it out for the first time and investigate if it's worth further investment. At the very least we should have an FAQ item on "Can I develop on Windows?" - we should just debate on what to put in the answer.

My suggestion is that we say it's officially only OS X is supported (and point to the J2ObjC requirements). Then strongly suggest that if you can to properly use the plugin, you'll need to do development on OS X. Finally indicate in practice that limited parts may work on Windows but use them at your risk and only for experimentation.

I've written something up which I think sets the right tone. Please take a look.

FAQ.md
```diff
@@ -235,3 +235,17 @@ source and whitelist. Note how this gives and expected cycles of zero.
...
+### Can I develop on Windows?
+
+Development is officially supported only on Mac OS X. This matches the
+(J2ObjC Requirements)[http://j2objc.org/#requirements].
+In practice, the cycle_finder and translate tasks may work, though it is strongly
+suggested to use this only for experimentation and do all proper development on OS X.
+
+To experiment on Windows (skipping the unsupported tasks), the following may work:
+
+    ./gradlew shared:j2objcAssembleSource shared:j2objcAssembleResources shared:j2objcTest
```
advayDev1: j2objcTest requires compiling and linking the code - which should only work on Mac since it requires the Apple Frameworks.

FAQ.md
```diff
@@ -235,3 +235,17 @@ source and whitelist. Note how this gives and expected cycles of zero.
 For more details:
 - http://j2objc.org/docs/Cycle-Finder-Tool.html
 - https://groups.google.com/forum/#!msg/j2objc-discuss/2fozbf6-liM/R83v7ffX5NMJ
+
+
+### Can I develop on Windows?
+
+Development is officially supported only on Mac OS X. This matches the
+(J2ObjC Requirements)[http://j2objc.org/#requirements].
```
advayDev1: reverse the order of the hyperlink [anchor](url)

FAQ.md
```diff
@@ -235,3 +235,17 @@ source and whitelist. Note how this gives and expected cycles of zero.
...
+In practice, the cycle_finder and translate tasks may work, though it is strongly
```
advayDev1: as the user sees them, it is j2objcCycleFinder and j2objcTranslate

FAQ.md
```diff
@@ -235,3 +235,17 @@ source and whitelist. Note how this gives and expected cycles of zero.
 For more details:
 - http://j2objc.org/docs/Cycle-Finder-Tool.html
 - https://groups.google.com/forum/#!msg/j2objc-discuss/2fozbf6-liM/R83v7ffX5NMJ
+
+
+### Can I develop on Windows?
```
advayDev1: If we're going to do this, add a line to README with heading of minimum requirements which duplicates the j2objc requirements and also 'Gradle 2.4 or higher'.  I do not think the README should mention Windows at all.

FAQ.md
```diff
@@ -235,3 +235,17 @@ source and whitelist. Note how this gives and expected cycles of zero.
...
+
+Development is officially supported only on Mac OS X. This matches the
```
advayDev1: j2objc I believe does not accept bugs related to systems not meeting the minimum requirements.  I don't think we should either - and such a comment should be added.

advayDev1:
comments added.

advayDev1:
the most important thing i think is to make it clear that we won't add complexity to half-support windows/etc.  seeing:
https://groups.google.com/forum/#!searchin/j2objc-discuss/windows/j2objc-discuss/ryYtHXE07iU/N2V4w-UjPekJ (re this plugin)
https://groups.google.com/forum/#!searchin/j2objc-discuss/windows/j2objc-discuss/LZngJ27gpHU/nMTe_zFMLU4J
https://groups.google.com/forum/#!searchin/j2objc-discuss/windows/j2objc-discuss/Tikm5VqQ8ic/eQ2h9sAmrPgJ

i think we'll cause more headache for us and the j2objc team if we aren't strict about it.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/PackLibrariesTask.groovy
```diff
@@ -60,6 +61,12 @@ class PackLibrariesTask extends DefaultTask {
         Utils.projectDelete(project, getOutputLibDirFile())
         getOutputLibDirFile().mkdirs()
 
+        if (Utils.isWindows()) {
+            throw new InvalidUserDataException(
```
advayDev1: add an identical message, but as a logger.warn in the Plugin itself so everyone sees this, even if they don't successfully get to executing this task

FAQ.md
brunobowden: Done

FAQ.md
brunobowden: Done

FAQ.md
brunobowden: done

FAQ.md
brunobowden: fixed

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/PackLibrariesTask.groovy
brunobowden: Done

FAQ.md
brunobowden: Obviously it's needed for packing the libraries but how so for the tests? Removed that for now.

@gocursor - can you try running j2objcTest and confirm whether it fails.

brunobowden:
PTAL

@gocursor - I'd appreciate your input before merging this PR

advayDev1:
LGTM

FAQ.md
```diff
@@ -235,3 +235,19 @@ source and whitelist. Note how this gives and expected cycles of zero.
...
+[J2ObjC Requirements](http://j2objc.org/#requirements).
+In practice, the j2objcCycleFinder and j2objcTranslate tasks may work, though it is
+strongly suggested to use this for experimentation only and do all proper development
+on OS X. The team will reject any bugs related to systems that don't meet the
+[minimum requirements](README.md#minimum-requirements).
+
+To experiment on Windows (skipping the unsupported tasks), the following may work:
+
+    ./gradlew shared:j2objcAssembleSource shared:j2objcAssembleResources
```
gocursor: I would write:
  gradle shared:j2objcAssembleSource shared:j2objcAssembleResources

gocursor:
OK, I have reviewed the new FAQ, included one comment and you can merge this PR into the master branch.

But - to be honest - this solution has some issues which needs to be further worked out and the question is of course if you would be willing to accept my changes to the project which would enable better integration with Windows...

If I include the IOS subproject into my multi-project gradle structure, then "gradle build" will try to build everything including the IOS subproject and will fail there, because only tasks j2objcAssembleSource and j2objcAssembleResources are working under Windows. So I actually can not use this solution within my project. If I separate the IOS subproject from the rest, I have to put redundant code into its build.gradle...

The ideal solution for my case would be working "gradle build" for IOS subproject where all OSX specific tasks are turned off - something like this, which is already supported:

j2objcPackLibrariesDebug { enabled = false }
j2objcAssembleDebug { enabled = false }
j2objcPackLibrariesRelease { enabled = false }
j2objcAssembleRelease { enabled = false }

but with additional tasks disabled...

The "steering committee" of the j2objc-contrib/j2objc-gradle project will need to decide if better and more long(er) term support for Windows should be included into the project. Since Gradle is supported on Windows and since J2ObjC translator is pure-Java it shouldn't be a problem (See also http://j2objc.org/docs/Getting-Started.html ... "However, since the J2ObjC translator is pure-Java, translation can be done on other systems.")

If you would decide to include better Windows support, I can provide better integration and ensure the plugin will work on Windows (we have some production-grade projects which would use it).

You don't need to include the support because of me... Meanwhile I forked another version of j2objc gradle plugin (found here https://github.com/toddway/FeedRead/tree/master/core - mentioned author is also Bruno Bowden), just replaced ':' with File.pathSeparator and it works exactly as I expected: "gradle build" will run the Cycle Finder and the Translation but nothing else on Windows...

There may however be other people with the same use case as I have...

So please make a decision about Windows support and let me know if I should develop the necessary fixes for Windows...

advayDev1:
@gocursor 
- "but with additional tasks disabled..." - what does that mean?  Does the enabled=false block you provided not do what you need to disable tasks?  Are there other tasks you need to disable to translate what you want on Windows?
- the fork of the plugin you linked to is extremely old - the entire plugin has basically been redone a few times since that version.

FAQ.md
brunobowden: Using ./gradlew is very intentional. It's called the Gradle Wrapper and makes builds easier across multiple platform, different versions of Gradle and on CI servers:

Good example:
http://stackoverflow.com/a/27982996/1509221

From this description:
https://docs.gradle.org/current/userguide/gradle_wrapper.html

_The Gradle Wrapper (henceforth referred to as the “wrapper”) is the preferred way of starting a Gradle build. The wrapper is a batch script on Windows, and a shell script for other operating systems. When you start a Gradle build via the wrapper, Gradle will be automatically downloaded and used to run the build._

_The wrapper is something you should check into version control. By distributing the wrapper with your project, anyone can work with it without needing to install Gradle beforehand. Even better, users of the build are guaranteed to use the version of Gradle that the build was designed to work with. Of course, this is also great for continuous integration servers (i.e. servers that regularly build your project) as it requires no configuration on the server._

brunobowden:
@advayDev1 - @gocursor makes a good point that J2ObjC mentions how translation can be done on Windows and Linux: http://j2objc.org/docs/Getting-Started.html#how-do-i-run-on-windows-or-linux

I'll merge in this change. Let's shift the rest of the discussion to an issue.

gocursor:
OK, I created a new issue for Windows Support and we can continue our discussion here: https://github.com/j2objc-contrib/j2objc-gradle/issues/345

@advayDev1: the plugin may be extremely old, but it works on Windows without problems whereby the latest version does not...

PR #348, opened 2015-08-14
FAQ - Clarify expected cycle count

advayDev1:


brunobowden:
Sorry @advayDev1, I should've done a LGTM instead of merging.

PR #350, opened 2015-08-16
FAQ: Get proper syntax highlighting

advayDev1:


brunobowden:
LGTM... but add in the two build.gradle code samples in README.md, they should also be gradle syntax

advayDev1:
#351 does that, so merging this one.

brunobowden:
Yeah, I saw that right afterwards ;-)

PR #353, opened 2015-08-17
xcodeTarget to also automatically match Tests and WatchKit App

brunobowden:
For an iOS app target of ‘IOS-APP’, this will add linkages for other
iOS targets that start with the same name. Such as ‘IOS-APPTests’,
‘IOS-APP WatchKit App’ and ‘IOS-APP WatchKit Extension’.

- Automatic linkage for Apple Watch apps and extension
- Matches against all targets that start with the same prefix
- Standardize xcodeTarget examples and test to use ‘IOS-APP’

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
@@ -36,7 +36,7 @@ import org.gradle.util.ConfigureUtil
  *
  * j2objcConfig {
  *     xcodeProjectDir '../ios'
- *     xcodeTarget 'IosApp'
+ *     xcodeTarget 'IOS-APP'
  *     finalConfigure()
  * }
  *
@@ -432,6 +432,7 @@ class J2objcConfig {
      * @param extraObjcSrcDirs add directories for Objective-C source to be compiled
      */
     void extraObjcSrcDirs(String... extraObjcSrcDirs) {
+        verifyNoSpaceArgs('extraObjcSrcDirs', extraObjcSrcDirs)
         for (String arg in extraObjcSrcDirs) {
             this.extraObjcSrcDirs += arg
         }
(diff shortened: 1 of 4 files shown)


brunobowden:
@advayDev1, @confile - PTAL, something I needed to make development easier for WatchKit Apps

confile:
What does this pull request?

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
```diff
@@ -565,8 +564,24 @@ class J2objcConfig {
...
+        verifyNoSpaceArgs(nameArgs, args)
+        listArgs.addAll(Arrays.asList(args))
+    }
+
+    // As above but for String[] instead of List<String>
+    @VisibleForTesting
+    static void appendArgs(String[] arrayArgs, String nameArgs, String... args) {
+        verifyNoSpaceArgs(nameArgs, args)
+        for (String arg in args) {
+            arrayArgs += arg
```
advayDev1: does this actually work?  you can modify the arrayArgs reference passed on?

brunobowden: You know, I was wondering the same thing and decided I need to unit test it... having now added the test, it doesn't work. So I've reverted the changes for String[] variables.

brunobowden:
@advayDev1 - it's all fixed up now

@confile - this makes it easier to work with multiple targets. For example, having a WatchKit app as well.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
advayDev1: yeah for other collections this is an .addAll call, and for arrays it is really an assignment and a plus.  one of the annoyances of groovy magic

advayDev1:
LGTM

confile:
LGTM

PR #356, opened 2015-08-17
codeStyleSettings.xml changes from Android Studio upgrade

brunobowden:


brunobowden:
@advayDev1 - I believe this is caused by a recent update for Android Studio (but can't confirm this). It's constantly trying to update the code style for the plugin. Have you seen this issue? Any concern with me merging this PR?

advayDev1:
did you recently install an Objective-C or related plugin in to your IDE?
there is no objective-c code in this plugin, so i don't think you should submit this.

make sure that your objective-c plugin (if any) is not enabled (via facets, frameworks, or modules) for this project.

PR #358, opened 2015-08-18
Enabled architectures, fixes #357

### LINKED ISSUE #357: Allow configuring enabled architectures in the environment
Similar to configuring debug, release, and translateOnly on a per-environment basis for developer convenience, permit enabling architectures in the environment via local.properties.

This is separate and in addition to setting the /supported/ architectures in the build file itself.  The final architectures built are the intersection of the supported architectures and the per-environment enabled ones (if any).

advayDev1:
Similar to configuring debug, release, and translateOnly on a per-environment basis for developer convenience, permit enabling architectures in the environment via local.properties.

This is separate and in addition to setting the /supported/ architectures in the build file itself. The final architectures built are the intersection of the supported architectures and the per-environment enabled ones (if any).

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
@@ -360,6 +360,35 @@ class J2objcConfig {
     // Public to allow assignment of array of targets as shown in example
     List<String> supportedArchs = NativeCompilation.ALL_SUPPORTED_ARCHS.clone() as List<String>
 
+    /**
+     * An architecture is active if it is both supported ({@link #supportedArchs})
+     * and enabled in the current environment via the comma-separated j2objc.enabledArchs
+     * value in local.properties.
+     * <p/>
+     * If no j2objc.enabledArchs value is specified in local.properties, all supported
+     * architectures are also active, otherwise the intersection of supportedArchs
+     * and j2objc.enabledArchs is used.
+     */
+    List<String> getActiveArchs() {
+        // null is the default value, since an explicit empty string means no architectures
+        // are enabled.
+        String archsCsv = Utils.getLocalProperty(project, 'enabledArchs', null)
+        if (archsCsv == null) {
+            return supportedArchs
+        }
+        List<String> enabledArchs = archsCsv.split(',').toList()
+        // Given `j2objc.enabledArchs=` we will have one architecture of empty string,
+        // instead we want no architectures at all in this case.
+        enabledArchs.remove('')
+        List<String> invalidArchs = enabledArchs.minus(
+                NativeCompilation.ALL_SUPPORTED_ARCHS.clone() as List<String>).toList()
+        if (!invalidArchs.isEmpty()) {
+            throw new InvalidUserDataException("Invalid 'enabledArchs' entry: " + invalidArchs.join(', '))
+        }
+        // Keep the return value sorted to prevent changes in intersection ordering
+        // from forcing a rebuild.
+        return supportedArchs.intersect(enabledArchs).toList().sort()
+    }
 
     // TEST
     /**
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/NativeCompilation.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/NativeCompilation.groovy
@@ -224,7 +224,7 @@ class NativeCompilation {
                                 }
                             }
                         }
-                        j2objcConfig.supportedArchs.each { String arch ->
+                        j2objcConfig.activeArchs.each { String arch ->
                             if (!(arch in ALL_SUPPORTED_ARCHS)) {
                                 throw new InvalidUserDataException(
                                         "Requested architecture $arch must be one of $ALL_SUPPORTED_ARCHS")
--- a/src/test/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfigTest.groovy
+++ b/src/test/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfigTest.groovy
@@ -16,6 +16,7 @@
 
 package com.github.j2objccontrib.j2objcgradle
 
+import com.github.j2objccontrib.j2objcgradle.tasks.TestingUtils
 import groovy.transform.CompileStatic
 import groovy.transform.TypeCheckingMode
 import org.gradle.api.InvalidUserDataException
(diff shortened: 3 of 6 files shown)


advayDev1:
@brunobowden  ptal;
for rapid development this lets you build a subset of your architectures

src/test/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfigTest.groovy
```diff
@@ -152,4 +156,52 @@ class J2objcConfigTest {
...
+        J2objcConfig j2objcConfig = new J2objcConfig(proj)
+        Assert.assertEqualsNoOrder(j2objcConfig.enabledArchs.toArray(),
+                j2objcConfig.supportedArchs.toArray())
+    }
+
+    @Test
+    void testSupportedAndEnabledArchs_SubsetEnabledArchsSpecified() {
+        // Multiple assignment not supported with static-typing.
+        Object[] projObjs = TestingUtils.setupProject(
+                new TestingUtils.ProjectConfig(createJ2objcConfig: true,
```
brunobowden: Add method that returns only the j2objcConfig from setting up a project. Something like:

    J2objcConfig j2objcConfig = TestingUtils.setupProjectJ2objcConfig(
        new TestingUtils.ProjectConfig(createJ2objcConfig: true,
        extraLocalProperties: ['j2objc.enabledArchs=ios_armv7,ios_armv7s']))

That'd be much neater than the temp projObjs array.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/NativeCompilation.groovy
```diff
@@ -224,7 +224,7 @@ class NativeCompilation {
                                 }
                             }
                         }
-                        j2objcConfig.supportedArchs.each { String arch ->
+                        j2objcConfig.enabledArchs.each { String arch ->
```
brunobowden: Is the `enabledArchs` => `getEnabledArchs` a way to make it work with `List<String>` type? If so, then can you convert supportedArchs to List<String> also?

brunobowden: In fact, should this be done for other String[] inputs for the native compilation?

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TestingUtils.groovy
```diff
@@ -78,7 +80,9 @@ class TestingUtils {
         // To satisfy Utils.J2objcHome()
         String j2objcHome = File.createTempDir('J2OBJC_HOME', '').path
         File localProperties = proj.file('local.properties')
-        localProperties.write("j2objc.home=$j2objcHome\n")
+        List<String> localPropertiesLines = ["j2objc.home=$j2objcHome"]
+        localPropertiesLines.addAll(config.extraLocalProperties)
+        localProperties.write(localPropertiesLines.join('\n'))
```
brunobowden: Nice addition in functionality.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
```diff
@@ -360,6 +360,27 @@ class J2objcConfig {
...
+        // null is the default value, since an explicit empty string means no architectures
+        // are enabled.
+        String archsCsv = Utils.getLocalProperty(project, 'enabledArchs', null)
+        if (archsCsv == null) {
+            return supportedArchs
+        }
+        String[] enabledArchs = archsCsv.split(',')
+        // Keep the return value sorted to prevent changes in intersection ordering
+        // from forcing a rebuild.
+        return supportedArchs.intersect(enabledArchs.toList()).toList().sort()
```
brunobowden: This should throw exception if there's an unidentified architecture, e.g. typo. Something like:

    List<String> invalidArchs = enabledArchs.toList().minus(NativeCompilation.ALL_SUPPORTED_ARCHS.clone() as List<String>).toList()
    if (!invalidArchs.isEmpty()) {
        throw new InvalidUserDataException("Invalid 'enabledArchs' entry in local.properties: " + invalidArchs.join(', '))
    }

brunobowden:
I like different items in the stack to be more closely related to each other. For example, the name of supportedArchs and enabledArchs should be the same. Given the local.properties configuration, do we still need to have a build.gradle setting?

Perhaps this is all better converted to a disableArch setting? That way, the impact of the j2objcConfig and local.properties is cumulative. It would also have a consistent name and clear behaviour. Please have a ponder about this.

advayDev1:
supportedArchs in build.gradl is necessary.  The project may just want to exist on later iOS platforms, or perhaps just x64 for OSX, etc.  It is a property of the project (in source control) of what platforms it supports.

What subset of those platforms you want to build at any given time I think should still be environment-specific.  I think enabledArchs is better than disabledArchs, because if you want to just do a debug loop on one device/simulator/etc, you just want to enable one architecture.

I agree though on the confusion between names, let me think about that a bit more.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/NativeCompilation.groovy
advayDev1: Not sure what you are asking here.  The accessor notation vs method notation has no impact on the type.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
advayDev1: Fair point.

advayDev1:
Alright here is my suggestion:
- supportedArchs - project setting
- enabledArchs - what the environment wants to build right now
- activeArchs - what will actually get built (intersection of the 2 above)

Why?  This is consistent with XCode from what I can tell.  But in Xcode your only options are having 1 or all architectures that your project supports being.  This is also consistent with our own names like 'debug' or 'release' enabled, being the environment-specific setting.

Updated appropriately.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfigTest.groovy
advayDev1: done

advayDev1:
ptal

brunobowden:
LGTM - I prefer it differently but this is fine

PR #361, opened 2015-08-18
Allow all local.properties flags to be overridden by environment

### LINKED ISSUE #360: All local.properties values should be configurable via environment too
More developer convenience:
If I want to rapidly build and run a debug-only iOS build I should be able to do `J2OBJC_RELEASE_ENABLED=false ./gradlew build`
and not have to change local.properties and then change it back.

The priority order is check environment first, and if not present, check local.properties.
j2objc.home's ordering will be changed appropriately.  This ordering makes a command like above work.

advayDev1:
Fixes #360

BREAKING API CHANGE:
J2OBJC_HOME environment value will now override j2objc.home
in local.properties instead of the other way around.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,11 +113,20 @@ Also see the FAQ note on [developing with Swift](https://github.com/j2objc-contr
 You can reduce the build time by 50% by skipping the release binaries by adding the
 following to your root level `local.properties` file:
 
-    j2objc.release.enabled=false
+```properties
+j2objc.release.enabled=false
+```
 
 This is helpful for a tight modify-compile-test loop using only debug binaries.
 You can also do this for `j2objc.debug.enabled`.
 
+If you'd rather just disable release builds for a particular run of the command line:
+
+```sh
+J2OBJC_RELEASE_ENABLED=false ./gradlew build
+```
+
+The `local.properties` value overrides the environment variable, if present.
 
 ### J2ObjC Standard Libraries
 
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
@@ -87,6 +87,12 @@ class Utils {
 
     private static final String PROPERTY_KEY_PREFIX = 'j2objc.'
 
+    /**
+     * Retrieves the local properties with highest precedence:
+     * 1.  local.properties value like j2objc.name1.name2 when present.
+     * 2.  environment variable like J2OBJC_NAME1_NAME2 when present.
+     * 3.  defaultValue.
+     */
     static String getLocalProperty(Project proj, String key, String defaultValue = null) {
 
         // Check for requesting invalid key
@@ -96,7 +102,7 @@ class Utils {
                     "Valid Keys: $PROPERTIES_VALID_KEYS")
         }
         File localPropertiesFile = new File(proj.rootDir, 'local.properties')
-        String result = defaultValue
+        String result = null
         if (localPropertiesFile.exists()) {
             Properties localProperties = new Properties()
             localPropertiesFile.withInputStream {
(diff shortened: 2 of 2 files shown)


advayDev1:
@brunobowden ptal.  please do #358 first

brunobowden:
Is there a precedence for this precedence? E.g. that the JVM or other tools have similar behaviour?

I've tried searching online but didn't get a good answer for this. The only example I could find seemed to suggest that local properties took precedence over the environment. I'm think that a system that has gone through battle testing on this front would have sound reasons for doing it that way. We should be careful about doing something difference.

http://www.javacodegeeks.com/2012/02/properties-with-spring.html

*4.1 Properties Search Precedence
By default, in Spring 3.1, local properties are search last, after all environment property sources, including properties files. This behavior can be overridden via the localOverride property of the PropertySourcesPlaceholderConfigurer, which can be set to true to allow local properties to override file properties.*

*In Spring 3.0 and before, the old PropertyPlaceholderConfigurer also attempted to look for properties both in the manually defined sources as well as in the System properties. The lookup precedence was also customizable via the systemPropertiesMode property of the configurer:
never – Never check system properties*

*fallback (default) – Check system properties if not resolvable in the specified properties files
override – Check system properties first, before trying the specified properties files. This allows system properties to override any other property source.*

*Finally, note that in case a property is defined in two or more files defined via @PropertySource – the last definition will win and override the previous ones. This makes the exact property value hard to predict, so if overriding is important, the PropertySource API can be used instead.*

advayDev1:
So does Android I guess:
https://android.googlesource.com/platform/tools/build/+/master/gradle/src/main/groovy/com/android/build/gradle/internal/Sdk.groovy

alright, I'll swap the priority order back to properties overrides environment

brunobowden:
Another issue we'll have for the future... is when build.gradle allows you
to specific the J2ObjC version, we'll likely want that to be the preferred
configuration system.

advayDev1:
ptal @brunobowden  - put the order back to the way it was

README.md
```diff
@@ -113,11 +113,19 @@ Also see the FAQ note on [developing with Swift](https://github.com/j2objc-contr
...
 
-    j2objc.release.enabled=false
+```properties
+j2objc.release.enabled=false
+```
 
 This is helpful for a tight modify-compile-test loop using only debug binaries.
 You can also do this for `j2objc.debug.enabled`.
 
+If you'd rather just disable release builds for a particular run of the command line:
```
brunobowden: Empty line afterwards

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
```diff
@@ -116,9 +117,14 @@ class Utils {
                 }
             }
 
-            result = localProperties.getProperty(PROPERTY_KEY_PREFIX + key, defaultValue)
+            result = localProperties.getProperty(PROPERTY_KEY_PREFIX + key, null)
         }
-        return result
+        if (result == null) {
```
brunobowden: Add comment on overall method that it will first get the localProperty with higher precedence and then get the environment variable.

README.md
```diff
@@ -113,11 +113,19 @@ Also see the FAQ note on [developing with Swift](https://github.com/j2objc-contr
...
+```sh
+J2OBJC_RELEASE_ENABLED=false ./gradlew build
+```
+
+The environment variable overrides the `local.properties` value, if present.
```
brunobowden: Need to change the precedence here

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
```diff
@@ -360,6 +360,35 @@ class J2objcConfig {
...
+    /**
+     * An architecture is active if it is both supported ({@link #supportedArchs})
+     * and enabled in the current environment via the comma-separated j2objc.enabledArchs
+     * value in local.properties.
+     * <p/>
+     * If no j2objc.enabledArchs value is specified in local.properties, all supported
+     * architectures are also active, otherwise the intersection of supportedArchs
+     * and j2objc.enabledArchs is used.
+     */
+    List<String> getActiveArchs() {
```
brunobowden: I prefer consistent naming, so I'd name the method getEnabledArchs()

brunobowden:
LGTM - it would be nice to add a test for Utils.groovy for loading a property from an environment variable.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
advayDev1: this is from the previous PR #358  - please see the discussion there, and let me know what you think

README.md
advayDev1: done

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
advayDev1: done

README.md
advayDev1: done

advayDev1:
filed a bug for mocking System.getenv - not sure how to at the moment

brunobowden:
Just add a TODO for now.

Nice update from Travis:

screen shot 2015-08-20 at 10 19 41 am

advayDev1:
Actually, it is baffling as it is a failure in JDK 8.0 only.  JDK 7.0 is
working.  Digging into it...

madvayApiAccess:
well the tests are passing now.  looks like we are flaky on JDK8 - i reran various commits on Travis over and over and JDK8 sometimes fails, sometimes passes.

i'm sending a PR to provide more info on such failures.  right now it just tells you to look at the testreport.html which obviously we can't.

brunobowden:
It would be helpful to have a more detailed record, including the
testreport.html. Can that be maintained for a brief period, e.g. 24 hours?

advayDev1:
We can't see the HTML file, but we can get the exception details:
http://mrhaki.blogspot.com/2013/05/gradle-goodness-show-more-information.html

see PR #373

PR #366, opened 2015-08-20
Travis build on OS X and Linux

advayDev1:


brunobowden:
LGTM.

Presumably this will work for now but may present issues in the future as we add more tests. In that case, we'll just just add a local.properties file for the tests on Linux / Windows.

advayDev1:
yes, until we add the system-level tests for actually translating this will work.
and you can indeed specify a matrix of environment variables per test config (even if we don't specify a local.properties that varies per OS)

this will get turned on once Travis support enables our project (email response pending)

PR #371, opened 2015-08-20
Note re: issues vs. email

advayDev1:
Currently getting direct email...


brunobowden:
LGTM - when we get some more activity, we'll set up a mailing list.

advayDev1:
SG.  Although even then, I think for active debugging, issues are best.
Linking to commits, syntax highlighting with markdown, etc. are super useful.

PR #375, opened 2015-08-20
Revert "Travis build on OS X and Linux"

advayDev1:
Reverts j2objc-contrib/j2objc-gradle#366

Per Travis support, Groovy and multiple JDK versions not yet supported on Mac.


brunobowden:
LGTM

PR #376, opened 2015-08-20
Disable JDK8 until we can solve #374

advayDev1:
I can't debug precisely until I can install JDK 8 and see what happens.
No need for the build to fail until I can do that.


brunobowden:
LGTM

advayDev1:
testin is ~50% faster without JDK8 too

PR #377, opened 2015-08-20
Version => 0.4.1

brunobowden:


advayDev1:
w00t.  lgtm.

brunobowden:
One is perhaps a little too enthused? Now that the Travis of Integration Continuous has finished, can we proudly confirm our progress.

PR #381, opened 2015-08-21
Treat prefixes (in the args and in files) as an input to translation.

### LINKED ISSUE #248: Accurate up-to-date checks for prefixes files
If you change the contents of prefixes properties file basically everything needs to be redone.  Any prefixes files on the translateArgs should be InputFiles to all tasks.

advayDev1:
Fixes #248


brunobowden:
LGTM - but you MUST add the same @input to CycleFinderTask as well.

advayDev1:
@brunobowden - i had considered that, but CycleFinderTask does not use the prefixes in any form obvious to me.  Note the lack of Utils.packagePrefixes or j2objcConfig.translateArgs in that class.  Can you tell me why it should use this input?

brunobowden:
Fair point... it won't use it because that's only for producing the output, not parsing the abstract syntax tree or similar.

PR #384, opened 2015-08-22
CycleFinderTask should indeed write a report on failure

advayDev1:


src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/CycleFinderTask.groovy
```diff
@@ -166,6 +166,11 @@ class CycleFinderTask extends DefaultTask {
             if (!Utils.isProjectExecNonZeroExit(exception)) {
                 throw exception
             }
+            
+            // Cycle finder succeeded so report will be useful to determine
+            // which cycles are appropriate and which are not.
+            getReportFile().write(Utils.stdOutAndErrToLogString(stdout, stderr))
```
brunobowden: Put this code in the `finally` section and then it doesn't have to be duplicated. Since stdout / stderr is declared outside of the `try` it will work correctly.

brunobowden:
LGTM - after merging duplicated code to finally.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/CycleFinderTask.groovy
advayDev1: Actually you'll notice I put that line after L168:
If the process itself fails, we do not want to overwrite the report.

However if the process succeeds but has an unexpected number of cycles, we do want to overwrite.

advayDev1:
see reply above: that isn't deduplication, it changes the semantics of the code.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/CycleFinderTask.groovy
brunobowden: "If the process exec itself fails, we do not want to overwrite the report." - Why not?

I think we should write the error file in all circumstances, including failures and successes. If the process exec has a "NonZeroExit", then it would still be useful to output stdout / stderr.

advayDev1: hmm ok.  i don't agree, but it's not a big deal.

brunobowden: This is a change compared to current behaviour but I didn't understand your
perspective... are you thinking that it shouldn't write the report if it's
a projectExec failure? If anything it might be confusing to leave the old
inaccurate report in place.

advayDev1: my thought was:
if cycle finder has an internal failure, you should see that on the command line.
if your code fails to be accepted by cycle finder, that is a report to analyze, publish, archive (like many people do with tests, coverage, etc. reports).

similar to say unit tests.  if you run junit, and the process itself fails you won't get an XML with analysis.  but if your tests fail, you will.

i agree that we could remove the report altogether in the beginning.
either way, i've made the fix you suggested and merging =)

PR #388, opened 2015-08-23
Version => 0.4.2

### LINKED ISSUE #387: Bad Release for v0.4.1-alpha ?
@brunobowden ,

I have the 0.4.1-alpha dist jar and source jar from plugins.gradle.org and I'm seeing errors from missing functionality though.  For example the source for Utils.groovy is missing the lines regarding enabledArchs, and when using those in my build file they fail with an error message that is older than v0.4.1.

Are you sure you published the right revision to plugins.gradle.org?

brunobowden:
- Fixes #387 (bad 0.4.1 build)

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Configure `shared/build.gradle` in your Java only project:
 // File: shared/build.gradle
 plugins {
     id 'java'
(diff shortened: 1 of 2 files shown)


advayDev1:
lgtm - thanks!

PR #389, opened 2015-08-23
typo

advayDev1:


brunobowden:
LGTM - but also fix the same issue for line no. 120 in FAQ.md

PR #390, opened 2015-08-23
Create CHANGELOG.md

advayDev1:
Having this in source control preserves it independent of github UI.
In addition, this will go into the published plugin (in addition to license, authors, readme, etc.) in a future PR.


brunobowden:
LGTM - want to add a note to the publish steps that this should be updated along with updating the version?

advayDev1:
will do so in the next PR that sets up incorporating this and the other
docs into the artifact

PR #393, opened 2015-08-23
README: reduce flux comment

advayDev1:
I cannot imagine us making major architectural changes at this point
so don't unduly scare people off.


brunobowden:
@advayDev1 - sorry, that should've been a LGTM instead of a merge.

PR #396, opened 2015-08-24
Detect Windows and Linux and warn if not supported

brunobowden:
- Error on tasks when Mac OS X is required
- Corrected link to the FAQ
- FAQ now also mentions Linux alongside Windows
- Fake OS name to unit test for Mac and nonMac systems
- inTestMustFakeOS property to ensure hermetic tests

Testing:
- Mac and nonMac command lines
- Exception thrown when attempting unsupported tasks
  (j2objcTest, j2objcPackLibraries and j2objcXcode)

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
@@ -60,8 +60,43 @@ class Utils {
         }
     }
 
+    @VisibleForTesting
+    static String fakeOSName = ''
+
+    @VisibleForTesting
+    static String getLowerCaseOSName() {
+
+        String osName = System.getProperty('os.name')
+        if (!fakeOSName.isEmpty()) {
+            osName = fakeOSName
+        } else {
+            if (System.getProperty('inTestMustFakeOS') == 'true') {
+                throw new InvalidUserDataException(
+                        "Tests must set Utils.fakeOSName = 'OS NAME'\n" +
+                        "This ensure that tests don't depend on the system environment")
+            }
+        }
+        osName = osName.toLowerCase()
+        return osName
+    }
+
     static boolean isWindows() {
-        return System.getProperty('os.name').toLowerCase().contains('windows')
+        String osName = getLowerCaseOSName()
+        return osName.contains('windows')
+    }
+
+    static boolean isMacOSX() {
+        String osName = getLowerCaseOSName()
+        // From: http://stackoverflow.com/a/18417382/1509221
+        return osName.contains('mac') || osName.contains('darwin')
+    }
+
+    static void requireMacOSX(String taskName) {
+        if (!isMacOSX()) {
+            throw new InvalidUserDataException(
+                    "Mac OS X is required for $taskName. Use `translateOnlyMode` on Windows or Linux:\n" +
+                    'https://github.com/j2objc-contrib/j2objc-gradle/blob/master/FAQ.md#how-do-i-develop-on-windows-or-linux')
+        }
     }
 
     static void throwIfNoJavaPlugin(Project proj) {
--- a/FAQ.md
+++ b/FAQ.md
@@ -259,15 +259,16 @@ For more details:
 - https://groups.google.com/forum/#!msg/j2objc-discuss/2fozbf6-liM/R83v7ffX5NMJ
 
 
-### How do I develop on Windows?
+### How do I develop on Windows or Linux?
 
-Windows support is limited to the j2objcCycleFinder and j2objcTranslate tasks. Windows can't
-support j2objcTest, j2objcXcode or the library packing (see [task descriptions](README.md#tasks)).
-Complete development on iOS is supported only on Mac OS X. This matches the
+Windows and Linux support is limited to the j2objcCycleFinder and j2objcTranslate tasks.
+Mac OS X is required for the j2objcAssemble, j2objcTest and j2objcXcode tasks
+(see [task descriptions](README.md#tasks)) This matches the
 [J2ObjC Requirements](http://j2objc.org/#requirements).
 
-It is recommended that Windows users use translateOnlyMode to reduce the chances of breaking the
-iOS / OS X build. This can be done by the Windows developers configuring their local.properties:
+It is recommended that Windows and Linux users use `translateOnlyMode` to reduce the chances
+of breaking the iOS / OS X build. This can be done by those developers configuring their
+local.properties (the Mac OS X developers should not use this):
 
 ```properties
 # File: local.properties
--- a/build.gradle
+++ b/build.gradle
@@ -48,6 +48,7 @@ test {
         // of just pointing to an HTML file.
         exceptionFormat = 'full'
     }
+    systemProperty 'inTestMustFakeOS', 'true'
 }
 
 pluginBundle {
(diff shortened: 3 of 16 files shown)


brunobowden:
PTAL

advayDev1:
I guess this causes unit-tests to fail:
com.github.j2objccontrib.j2objcgradle.J2objcConfigTest > testFinalConfigure FAILED
    org.gradle.api.InvalidUserDataException: Mac OS X is required for this task. Use `translateOnlyMode` on Windows or Linux:
    https://github.com/j2objc-contrib/j2objc-gradle/blob/master/FAQ.md#how-do-i-develop-on-windows-or-linux
        at com.github.j2objccontrib.j2objcgradle.tasks.Utils.requireMacOSX(Utils.groovy:71)
        at com.github.j2objccontrib.j2objcgradle.J2objcConfig.finalConfigure(J2objcConfig.groovy:565)
        at com.github.j2objccontrib.j2objcgradle.J2objcConfigTest.testFinalConfigure(J2objcConfigTest.groovy:78)

brunobowden:
@advayDev1 - it was quite a bit of work but there's now a Utils.fakeOSName static string that can be set for testing. I was tempted to add some tests that skip using fakeOSName... but I think that's likely best covered by doing some actual project builds.

brunobowden:
I've improved a number of the unsupported error messages, indicating which task is unsupported.

This also pushes us over 100 unit tests... we're now at 104 of them.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/CycleFinderTask.groovy
```diff
@@ -94,11 +94,11 @@ class CycleFinderTask extends DefaultTask {
     @TaskAction
     void cycleFinder() {
         String cycleFinderExec = "${getJ2objcHome()}/cycle_finder"
-        List<String> windowsOnlyArgs = new ArrayList<String>()
-        if (Utils.isWindows()) {
+        List<String> nonMacOSXArgs = new ArrayList<String>()
+        if (!Utils.isMacOSX()) {
             cycleFinderExec = 'java'
```
advayDev1: is the java binary needed on Linux (as opposed to just windows)?

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTask.groovy
```diff
@@ -213,11 +213,11 @@ class TranslateTask extends DefaultTask {
         }
 
         String j2objcExecutable = "${getJ2objcHome()}/j2objc"
-        List<String> windowsOnlyArgs = new ArrayList<String>()
-        if (Utils.isWindows()) {
+        List<String> nonMacOSXArgs = new ArrayList<String>()
+        if (!Utils.isMacOSX()) {
```
advayDev1: ditto

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
```diff
@@ -60,8 +60,26 @@ class Utils {
...
     }
 
-    static boolean isWindows() {
-        return System.getProperty('os.name').toLowerCase().contains('windows')
+    @VisibleForTesting
+    static String fakeOSName = ''
+
+    static boolean isMacOSX() {
+        // From: http://stackoverflow.com/a/18417382/1509221
+        println 'Uppercase OS: ' + System.getProperty('os.name')
```
advayDev1: remove

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
```diff
@@ -60,8 +60,26 @@ class Utils {
         }
     }
 
-    static boolean isWindows() {
-        return System.getProperty('os.name').toLowerCase().contains('windows')
+    @VisibleForTesting
+    static String fakeOSName = ''
+
+    static boolean isMacOSX() {
+        // From: http://stackoverflow.com/a/18417382/1509221
```
advayDev1: 1) https://blog.stackexchange.com/2009/06/attribution-required/
2) I do not know whether http://creativecommons.org/licenses/by-sa/3.0/ is compatible with Apache 2.0.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
```diff
@@ -60,8 +60,26 @@ class Utils {
         }
     }
 
-    static boolean isWindows() {
-        return System.getProperty('os.name').toLowerCase().contains('windows')
+    @VisibleForTesting
+    static String fakeOSName = ''
```
advayDev1: My worry with this is that it is easy to forget to put this in a test setup.
And if any test file is missing it, the test results become a function of the local env, and no longer hermetic.

I'd suggest making this fool-proof by:
1.  Adding to build.gradle:
```gradle
test {
    systemProperty 'inTestMustFakeOS', 'true'
}
```

2.  In isMacOSX, if System.getProperty(above) is true, and fakeOSName is empty, assert fail.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/CycleFinderTask.groovy
brunobowden: 'java' only used on Windows

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTask.groovy
brunobowden: 'java' for Windows only

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
brunobowden: done

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
brunobowden: Attribution is meant for when you're displaying the full question and answer, not just for a hyperlink.
Copying a single line of code indisputably comes under fair use exception for copyright.

brunobowden: In this case, additional fair use protection is afforded by the fact that I changed the code in the process.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
brunobowden: Added 'inTestMustFakeOS'... this'll need to change once we add better integration tests. In that case, we'll want to run the integration tests natively on each OS.

advayDev1: yeah travis supports us doing a 'matrix' of commands on various operating systems.

i think integration tests will also be everywhere (as they are our own classes put together)
and then we'd have commands like systemTestsOsx systemTestsLinux systemTestsWindows each of which would run only on their os.

advayDev1:
lgtm

PR #397, opened 2015-08-24
Run CI only on official branches.

madvayApiAccess:


brunobowden:
LGTM - I'm not sure why this is needed... although we have multiple tags, we only have one master

advayDev1:
Tags are branches as far as Travis is concerned.  See:
https://travis-ci.org/j2objc-contrib/j2objc-gradle/branches

The problem is that revert.* branch you see.  Travis keeps forever all branches ever created on the upstream repo (this one).  There's no way to get rid of that clutter; this keeps it from ever showing up.

Note it does not remove the existing functionality of building every PR - including reverts.

PR #398, opened 2015-08-24
Don't warn on SNAPSHOT after error on unclean Git

brunobowden:


brunobowden:
@advayDev1 - after seeing the error, the warning seems redundant

advayDev1:
sure, lgtm if green

Issue #400, opened 2015-08-24
Setup Windows CI on Appveyor


advayDev1:
Looks simple enough.

advayDev1:
@brunobowden - as before, I will do a direct commit via @madvayApiAccess with the initial config, and PR any config change.  please do post-review is as needed.  I will put the commit sha here for clarity.

brunobowden:
Looks like Windows support on Travis has been an open issue for a long
time. This seems to be the current tracking issue - I'd suggest subscribing
to that: https://github.com/travis-ci/travis-ci/issues/2104

advayDev1:
alright, you can do a post-review of #401 and LMK anything you want fixed.
I'll also get a normal PR out to put the README, etc.
The failing tests are in #402

PR #401, opened 2015-08-24
Initial AppVeyor CI config.

madvayApiAccess:
Generated from AppVeyor UI.

#400

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,23 @@
+version: 1.0.{build}.{branch}
+pull_requests:
+  do_not_increment_build_number: true
+branches:
+  only:
+  - master
+  - /release.*/
+  - /v[0-9].*/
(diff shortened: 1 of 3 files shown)


appveyor.yml
```diff
@@ -0,0 +1,23 @@
+version: 1.0.{build}.{branch}
+pull_requests:
+  do_not_increment_build_number: true
+branches:
+  only:
+  - master
+  - /release.*/
+  - /v[0-9].*/
+os: Windows Server 2012 R2
```
brunobowden: I think a consumer OS would probably be more relevant... but I see this is the only OS version available on Appveyor, so this'll be fine.

brunobowden:
LGTM

appveyor.yml
advayDev1: it is also the default

PR #405, opened 2015-08-24
WIP: Force OS X builds.

advayDev1:
The only way I can kick off Travis builds for testing is via PRs.
Please ignore these for now, @brunobowden


advayDev1:
Made a real PR - this is just investigations.

PR #406, opened 2015-08-24
Support OS X builds on Travis [workaround]

### LINKED ISSUE #399: Setup OS X CI builds on Travis
Until they have formal support for Groovy, some other projects use some workarounds.
I'll investigate.

advayDev1:
Fixes #399

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,36 @@
-language: groovy
+# We have to configure the language in the matrix below.
+language: generic
+
+matrix:
+  include:
+    - os: linux
+      language: groovy
+      jdk: oraclejdk7
+      env: USING_OS=linux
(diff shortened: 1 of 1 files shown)


advayDev1:
@brunobowden  - this is now ready.

You can see the investigation that went into this here:
https://github.com/j2objc-contrib/j2objc-gradle/pull/405

brunobowden:
LGTM - thanks for being persistent on this. While I haven't validated everything, I can't see anything that's prominently out of place.

advayDev1:
it is hard to validate, it is a kluge of experimentally verified workarounds ;)

i will now ask travis to re-enable our multi-os support
and we should be able to natively compile at that point

PR #409, opened 2015-08-24
Our first system test!

advayDev1:
This can be improved a lot, but we have to capture a beach-head first.


advayDev1:
Works locally, works on Travis:
https://travis-ci.org/j2objc-contrib/j2objc-gradle/jobs/77048032#L2741

ptal @brunobowden

systemTests/.gitignore
```diff
@@ -0,0 +1,8 @@
+# Ignore all the wrappers below.
+gradle/
+gradlew.bat
+# Under this directory, we do in fact need to preserve local.properties!
+!local.properties
+# Generated by prep.sh
+common/local.properties
+localJ2objcDist/
```
brunobowden: What are these two ignore lines for?

advayDev1: As it says, those files/directories are generated by prep.sh

systemTests/prep.sh
```diff
@@ -0,0 +1,34 @@
...
+set -e
+
+pushd localJ2objcDist
+
+# For developer local testing, don't keep redownloading the zip file.
+if [ ! -e j2objcDist.zip ]; then
+  curl -L https://github.com/google/j2objc/releases/download/$J2OBJC_VERSION/j2objc-$J2OBJC_VERSION.zip > j2objcDist.zip
+  unzip j2objcDist.zip
+  mv j2objc-$J2OBJC_VERSION j2objcDist
+  echo j2objc.home=$PWD/j2objcDist > ../common/local.properties
```
brunobowden: Does this just append rather than replace? Probably fine for now.

advayDev1: It replaces, which is correct.

systemTests/simple1/src/main/java/com/example/HelloWorld.java
```diff
@@ -0,0 +1,31 @@
...
+
+    public HelloWorld(String nameOfWorld) {
+        this.nameOfWorld = nameOfWorld;
+    }
+
+    @Override
+    public String toString() {
+        return "Hello " + nameOfWorld;
+    }
+}
```
brunobowden: Should have trailing newline

systemTests/prep.sh
advayDev1: append is `>>`

systemTests/simple1/src/main/java/com/example/HelloWorld.java
advayDev1: done

brunobowden:
LGTM

You should also see if you can validate a build error. E.g. by posting a PR that breaks the Hello World... both the 'java' build and then translated, e.g. using some OCNI junk.

advayDev1:
That's a really good point.  Will do

advayDev1:
Errors correctly fail:
https://travis-ci.org/j2objc-contrib/j2objc-gradle/jobs/77052795

merging

Issue #407, opened 2015-08-24
Continuous system tests with integrated j2objc on OS X


advayDev1:
When USING_OS == osx on Travis:
1.  Download and unzip j2objc distribution.
2.  Build plugin via gradle.
3.  Run simplest possible build.

Going to take a crack at this myself, but if I run out of time allocated to this will unassign.

advayDev1:
For now this will be a simple script, and not a Gradle TestKit #339 as I don't have time to do that right now.

PR #411, opened 2015-08-24
Multi-project system test.

advayDev1:
#407


brunobowden:
@advayDev1 - I'll review this later tonight

advayDev1:
SG.  I'm about to PR Automatic Dependency resolution which extends this PR.

brunobowden:
I noticed that commonly a http://remote/lib.jar will have the library
available at http://remote/lib-sources.jar. My hope is that for most common
libraries, you only need to add the `compile` or `testCompile` and
everything works. Some cases might need more configuration, but then you
can manually download and specify the sources jar.

advayDev1:
All done at #420

systemTests/multiProject1/build.gradle
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+buildscript {
+    repositories {
+        jcenter()
+    }
+    dependencies {
+        classpath fileTree(dir: '../../build/libs', include: ['*.jar'])
```
brunobowden: Is this because it's the `build/libs` directory from off ROOT_DIR? Please add explanation for `../../` parent directory.

advayDev1: yes, it is the output of the plugin build.  noted

systemTests/multiProject1/sub1/src/main/java/com/example/Cube.java
```diff
@@ -0,0 +1,50 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example;
+
+import java.lang.Override;
+import java.lang.String;
+
+public class Cube {
```
brunobowden: This class should have minimal complexity needed for the test. More than that, just obfuscates what's happening. I'd replace this will something as simple as a `public String helloWorld() { return "Hello World"; }` method.

systemTests/multiProject1/sub1/build.gradle
```diff
@@ -0,0 +1,32 @@
...
+apply plugin: 'com.github.j2objccontrib.j2objcgradle'
+
+repositories {
+    jcenter()
+}
+
+dependencies {
+    compile 'com.google.guava:guava:17.0'
+    testCompile 'junit:junit:4.12'
+    testCompile "org.mockito:mockito-core:1.9.5"
```
brunobowden: Can mockito be removed?

systemTests/multiProject1/sub1/build.gradle
```diff
@@ -0,0 +1,32 @@
...
+
+apply plugin: 'java'
+apply plugin: 'com.github.j2objccontrib.j2objcgradle'
+
+repositories {
+    jcenter()
+}
+
+dependencies {
+    compile 'com.google.guava:guava:17.0'
```
brunobowden: Can guava be removed?

systemTests/multiProject1/sub2/build.gradle
```diff
@@ -0,0 +1,35 @@
...
+
+repositories {
+    jcenter()
+}
+
+dependencies {
+    compile project(':sub1')
+    compile 'com.google.guava:guava:17.0'
+    testCompile 'junit:junit:4.12'
+    testCompile "org.mockito:mockito-core:1.9.5"
```
brunobowden: Can mockito be removed?

systemTests/multiProject1/sub2/build.gradle
```diff
@@ -0,0 +1,35 @@
...
+apply plugin: 'java'
+apply plugin: 'com.github.j2objccontrib.j2objcgradle'
+
+repositories {
+    jcenter()
+}
+
+dependencies {
+    compile project(':sub1')
+    compile 'com.google.guava:guava:17.0'
```
brunobowden: Can guava be removed?

systemTests/multiProject1/sub2/build.gradle
```diff
@@ -0,0 +1,35 @@
...
+
+apply plugin: 'java'
+apply plugin: 'com.github.j2objccontrib.j2objcgradle'
+
+repositories {
+    jcenter()
+}
+
+dependencies {
+    compile project(':sub1')
```
brunobowden: So `sub2` extends `sub2`. Can we rename the class and projects like that?

For example sub1 => base and sub2 => extends?

systemTests/multiProject1/sub1/build.gradle
advayDev1: instead, I added code that uses it.  I want to make sure this dependency works e2e.

systemTests/multiProject1/sub2/build.gradle
advayDev1: yes done

systemTests/multiProject1/sub2/build.gradle
advayDev1: see above re: guava

systemTests/multiProject1/sub1/build.gradle
advayDev1: yes done

systemTests/run-all.sh
```diff
@@ -17,5 +17,16 @@
...
 
-echo Running test simple1
-pushd simple1 && ./gradlew wrapper && ./gradlew clean && ./gradlew build && popd
+function runTest {
+    TEST_DIR=$1
+    echo Running test $TEST_DIR
+    set -e
+    pushd $TEST_DIR
+    ./gradlew wrapper
+    ./gradlew clean
```
brunobowden: Do you want it to be distinct assemble and build steps, like it does in the CI? Presumably this helps ensure that non-test code can compile independently of doing any testing.

systemTests/run-all.sh
```diff
@@ -17,5 +17,16 @@
...
+    ./gradlew build
+    popd
+}
+
+runTest simple1
+runTest multiProject1
```
brunobowden: Should the list of tests be inferred automatically? If so, then at least add a TODO for that.

systemTests/multiProject1/sub1/src/main/java/com/example/Cube.java
advayDev1: trimmed most of it out.
do want to exercise a class in project 2 overriding methods in project 1 though.

systemTests/run-all.sh
advayDev1: No, it is intentionally not separate like the code for the plugin itself.
I want to run it the way users will - most commonly, just `build`.  And our code does behave differently when tasks are run combined vs. not (ex. #306 ).

systemTests/run-all.sh
advayDev1: I don't think it needs to be so complicated.  I'll add the TODO though.

systemTests/multiProject1/sub2/build.gradle
advayDev1: I'll go with base and extended.  Not sure if using a java reserved word would affect things.

advayDev1:
ptal.  repushed and passing.

systemTests/multiProject1/base/build.gradle
```diff
@@ -0,0 +1,31 @@
...
+
+apply plugin: 'java'
+apply plugin: 'com.github.j2objccontrib.j2objcgradle'
+
+repositories {
+    jcenter()
+}
+
+dependencies {
+    compile 'com.google.guava:guava:17.0'
```
brunobowden: Add note that you're intentionally adding guava to test the e2e compile

brunobowden:
LGTM - though see if you can think of a better name than `multiProject1`

advayDev1:
That is quite intentional.  I think we'll have multiProject2... N

systemTests/multiProject1/base/build.gradle
advayDev1: done

PR #414, opened 2015-08-25
Travis support re-enabled multi-os on our account.

advayDev1:
The UI and build attributes are now correct.  We still have
to do the language workaround and other hacks in here, but the
UI will now respond correctly.

#407


brunobowden:
LGTM

PR #415, opened 2015-08-25
Document dependencies on 3rd party jars

### LINKED ISSUE #372: Automatically handle third-party libraries
@ph1lb4 - could you repost your emailed question re: translating dependent jars here?

Please provide the relevant j2objcConfig, the version of the plugin you are using, your dependencies { } block if possible, and any errors.

advayDev1:
Documentation for #372 

Fixes #372

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -147,12 +147,54 @@ j2objcConfig {
 
 See the following FAQ answers...
 
+### How do I setup a dependency to a third-party Java library?
+
+These are the kinds of dependencies usually specified as follows in build `shared/build.gradle`.
+We'll use `gson` as an example:
+```gradle
+dependencies {
+    compile 'com.google.code.gson:gson:2.3.1'
+}
+```
+
+Per J2ObjC's documentation: "Developers must have source code for their Android app, which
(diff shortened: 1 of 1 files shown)


advayDev1:
@brunobowden @ph1lb4 @jjroloff - please take a look

FAQ.md
```diff
@@ -147,12 +147,54 @@ j2objcConfig {
...
+    ...
+}
+```
+
+In the future, this kind of dependency should be inferred automatically from the corresponding
+Java dependency - [issue 41](https://github.com/j2objc-contrib/j2objc-gradle/issues/41).
+
+Also note that this directly incorporates only the neccessary files from your dependencies into
+your Objective C libraries.  It does not produce a separate `gson` Objective C library that you can
+use standalone.  You would need to create a separate Gradle Java project manually to do that.
```
brunobowden: "separate Gradle Java project" - was there an FAQ item about this? If so, then link to it.

brunobowden:
LGTM

FAQ.md
advayDev1: nah I haven't this yet.  #419

PR #421, opened 2015-08-26
Convert AUTHORS to NOTICE

advayDev1:
Use a standard Apache 2 Sec 4d NOTICE.
Soft link AUTHORS over to NOTICE since we reference that all over.


brunobowden:
LGTM - why does the AUTHORS file show up as a delete and new file rather than a modification to an existing file?

advayDev1:
because technically it is a new file (files are a combo of mode and name - the mode is no longer `regular file` but `link`).

PR #423, opened 2015-08-26
Update appveyor.yml

advayDev1:
Build numbers are sequential - makes no sense to make them decimals,
unless we had the ability to infer the version number of the project.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,4 @@
-version: 1.0.{build}.{branch}
+version: {build}.{branch}
 pull_requests:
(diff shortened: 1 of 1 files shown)


brunobowden:
LGTM - not quite clear to me what this is doing

brunobowden:
LGTM - I think I understand this now. Is there any way to put the commit SHA in there or the build version? Either one or preferably both would be helpful.

advayDev1:
i looked for both, it seems not.
for prs it inexplicably appends a random string instead of the sha

PR #431, opened 2015-08-27
Build external Java libs into standalone native libs.

### LINKED ISSUE #419: Handle duplicate dependencies
Say you have shared1 and shared2 projects.
shared1 depends on shared2.
shared2 depends on external jar say 'gson'.
shared1 also depends on external jar 'gson'.

Because --build-closure will pull in gson functions into both, your final linking fails with duplicate symbol definition.

Fix this.

(the right solution is to define a separate j2objc project for gson - we should provide a way to automate that?)

### LINKED ISSUE #429: Feature to convert an entire library jar via j2objc
`translateAllSources=true` will cause all files in the translateSourcepaths to be compiled.  This should be a good step to solving #419 .  Namely, let's say you have 2 projects depending on GSON.   Instead of each project retranslating GSON, you can create a small gradle project that looks like:
```gradle
// third_party/gson/build.gradle
plugins {
  id 'java'
  id 'j2objc-gradle' // pseudocode
}

dependencies {
  compile "com.google.code.gson:gson:2.3.1"
}

j2objcConfig {
  translateAllSources true
  finalConfigure()
}
```

Then your other projects using gson, instead of directly depending on `com.google.code.gson:gson:2.3.1` would depend on `project(":third_party_gson").

@brunobowden, thoughts on this design?

advayDev1:
Rename the old `j2objcTranslation` to `j2objcTranslationClosure`

Add a new dependency type `j2objcTranslation` that allows you to
convert an entire source jar into a standalone library as a separate Gradle project.
You then depend on that project instead of the external library when
you want to use the functionality elsewhere.

This causes the external library to be translated and compiled only once,
and not partially and duplicatively as with --build-closure.

Fixes #429; fixes #419

TESTED=yes


advayDev1:
ping @brunobowden

src/main/groovy/com/github/j2objccontrib/j2objcgradle/DependencyConverter.groovy
```diff
@@ -26,7 +26,7 @@ import org.gradle.api.artifacts.SelfResolvingDependency
 
 /**
  * Converts `[test]compile` dependencies to their
- * `j2objcTranslation` and/or `j2objcLinkage` equivalents, depending on the type
+ * `j2objcTranslationClosure` and/or `j2objcLinkage` equivalents, depending on the type
```
brunobowden: Improve explanation:

    * Converts `[test]compile` dependencies to one of the following (or their equivalent):
    * j2objcLinkage - EXPLANATION HERE
    * j2objcTranslationClosure - EXPLANATION HERE

src/main/groovy/com/github/j2objccontrib/j2objcgradle/DependencyResolver.groovy
```diff
@@ -42,20 +44,50 @@ class DependencyResolver {
...
 
     void configureAll() {
+        project.configurations.getByName('j2objcTranslationClosure').each { File it ->
+            // These are the resolved files, NOT the dependencies themselves.
+            visitTranslationClosureFile(it)
+        }
         project.configurations.getByName('j2objcTranslation').each { File it ->
             // These are the resolved files, NOT the dependencies themselves.
-            visitTranslateFile(it)
+            visitTranslationLibrary(it)
```
brunobowden: Definite improvement

src/main/groovy/com/github/j2objccontrib/j2objcgradle/DependencyResolver.groovy
```diff
@@ -42,20 +44,50 @@ class DependencyResolver {
     }
 
     void configureAll() {
+        project.configurations.getByName('j2objcTranslationClosure').each { File it ->
+            // These are the resolved files, NOT the dependencies themselves.
+            visitTranslationClosureFile(it)
+        }
         project.configurations.getByName('j2objcTranslation').each { File it ->
             // These are the resolved files, NOT the dependencies themselves.
```
brunobowden: By `resolved files`, can you clarify what these files are likely to be... a jar? Or something else?

src/main/groovy/com/github/j2objccontrib/j2objcgradle/DependencyResolver.groovy
```diff
@@ -42,20 +44,50 @@ class DependencyResolver {
...
 
+    /**
+     * Adds to the main java sourceSet a to-be-generated directory that contains the contents
+     * of `j2objcTranslation` dependency libraries (if any).
+     */
+    static void configureSourceSets(Project project) {
+        JavaPluginConvention javaConvention = project.getConvention().getPlugin(JavaPluginConvention)
+        SourceSet sourceSet = javaConvention.sourceSets.findByName(SourceSet.MAIN_SOURCE_SET_NAME)
+        sourceSet.java.srcDirs(project.file("${project.buildDir}/translationExtraction"))
+        Copy copy = project.tasks.create('j2objcTranslatedLibraryExtraction', Copy,
```
brunobowden: Should be extracted as a `private final String` and referenced like so. Not so great to have it copied by hand in several places throughout the code.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/DependencyResolver.groovy
```diff
@@ -42,20 +44,50 @@ class DependencyResolver {
...
+                {Copy task -> task.into(project.file("${project.buildDir}/translationExtraction"))})
+        project.tasks.getByName(sourceSet.compileJavaTaskName).dependsOn(copy)
+    }
+
+    protected void visitTranslationLibrary(File depFile) {
```
brunobowden: 1) s/depFile/jarSourceFile/ - as this makes the use more clear
2) Would it be better to name the method visitTranslationSourceJar?

Add description comment before method:

    // Copy jar source to J2OBJC_TRANSLATED_LIBRARY_SOURCE

With the unzip, this is is nested folders, such that extracting multiple libraries won't conflict?

src/main/groovy/com/github/j2objccontrib/j2objcgradle/DependencyResolver.groovy
```diff
@@ -42,20 +44,50 @@ class DependencyResolver {
...
+        if (!depFile.absolutePath.endsWith('.jar')) {
+            String msg = "`j2objcTranslation` dependencies can only handle " +
+                         "source jar files, not ${depFile.absolutePath}"
+            throw new InvalidUserDataException(msg)
+        }
+        PatternSet pattern = new PatternSet()
+        pattern.include('**/*.java')
+        pattern.exclude('**/package-info.java')
```
brunobowden: Add note explaining why package-info.java is skipped.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
```diff
@@ -323,11 +323,12 @@ class J2objcConfig {
     /**
      * @see #dependsOnJ2objcLib(org.gradle.api.Project)
      * @deprecated Use `dependencies { j2objcLinkage project(':beforeProjectName') }` or
-     * `autoConfigDeps = true` instead.
+     * `autoConfigureDeps = true` instead.
      */
     // TODO: Do this automatically based on project dependencies.
     @Deprecated
     void dependsOnJ2objcLib(String beforeProjectName) {
+        //noinspection GrDeprecatedAPIUsage
```
brunobowden: Just curious, what was deprecated, anything we need worry about for the future?

systemTests/externalLibrary1/base/src/main/java/com/example/Cube.java
```diff
@@ -0,0 +1,41 @@
...
+    public Cube(int dimension) {
+        this.dimension = dimension;
+    }
+
+    @Override
+    public String toString() {
+        return String.format("[Cube %d]", dimension);
+    }
+
+    public String exerciseGuava() {
```
brunobowden: Create a separate `exerciseGson` method?

systemTests/run-all.sh
```diff
@@ -34,5 +34,11 @@ function runTest {
 runTest simple1
 
 # Two gradle projects, `extended` depends on `base`.  They also both test
-# dependency on built-in j2objc libraries, like Guava.
+# dependency on built-in j2objc libraries, like Guava, and build-closure
+# based translation of an external library, Gson.
 runTest multiProject1
+
+# Two gradle projects, `extended` depends on `base`.  A third gradle
```
brunobowden: Reword as:

Two gradle projects, `extended` depends on `base`.  Both of them depend on project `third_party_gson`, which fully translates and compiles an external library (Google's Gson). This library is used in both extended and base.

brunobowden:
LGTM

src/main/groovy/com/github/j2objccontrib/j2objcgradle/DependencyConverter.groovy
advayDev1: Done, but for DependencyConverter I just detailed how compile and testCompile get converted.
The definitions of j2objcLinkage, etc. are now in DependencyResolver's javadoc.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/DependencyResolver.groovy
advayDev1: done

src/main/groovy/com/github/j2objccontrib/j2objcgradle/DependencyResolver.groovy
advayDev1: done but we use literal strings copied for task names throughout the plugin already

src/main/groovy/com/github/j2objccontrib/j2objcgradle/DependencyResolver.groovy
advayDev1: 1) done
2) done
3) done
4) the idea is you would generally only convert one library per project using j2objcTranslation.  anyway, I added L74 to fail fast on duplicate entries.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/DependencyResolver.groovy
advayDev1: done

systemTests/run-all.sh
advayDev1: done

systemTests/externalLibrary1/base/src/main/java/com/example/Cube.java
advayDev1: done

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
advayDev1: it is calling our own deprecated method bellow.
i.e. it is a deprecated overloading

PR #433, opened 2015-08-27
Guava 18.0 builds locally!

advayDev1:
Also allow users to select which pre-built libraries they want to link in.
For example when testing Guava builds, we can't link to the existing j2objc Guava library

TESTED=system tests added

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
@@ -258,7 +258,7 @@ class J2objcConfig {
     boolean autoConfigureDeps = false
 
     /**
-     * Additional libraries that are part of the j2objc distribution.
+     * Additional Java libraries that are part of the j2objc distribution.
      * <p/>
      * For example:
      * <pre>
@@ -276,6 +276,18 @@ class J2objcConfig {
             // Libraries that don't need CycleFinder fixes
             "javax.inject-1.jar", "jsr305-3.0.0.jar",
             "mockito-core-1.9.5.jar"]
+
+    /**
+     * Additional native libraries that are part of the j2objc distribution.
+     * <p/>
+     * For example:
+     * <pre>
+     * linkJ2objcLibs = ["guava", "jsr305"]
+     * </pre>
+     */
+    // J2objc default libraries, from $J2OBJC_HOME/lib/..., without '.a' extension.
+    // TODO: auto add libraries based on java dependencies, warn on version differences
+    List<String> linkJ2objcLibs = ['guava', 'j2objc_main', 'javax_inject', 'jsr305']
 
 
     // TODO: warn if different versions than testCompile from Java plugin
--- a/systemTests/externalLibrary1/third_party_guava/build.gradle
+++ b/systemTests/externalLibrary1/third_party_guava/build.gradle
@@ -22,19 +22,43 @@ repositories {
 }
 
 dependencies {
+    // Most recent versions of Guava's dependencies, as inferred from Guava 18's POM file
+    // (which does not specify version at all).
+    // They are all annotations only.
+    compile 'com.google.j2objc:j2objc-annotations:0.9.8'
+    compile 'com.google.code.findbugs:jsr305:3.0.0'
+    compile 'com.google.errorprone:error_prone_annotations:2.0.4'
+    compile 'org.codehaus.mojo:animal-sniffer-annotations:1.14'
+
     // j2objcTranslation causes the entire specified library
     // to be translated and built into a standalone Objective C library.
     // j2objcTranslation must always be passed a *source* jar,
     // therefore, note the ':sources' classifier.
-    j2objcTranslation 'com.google.code.gson:gson:2.3.1:sources'
+    j2objcTranslation 'com.google.guava:guava:18.0:sources'
 }
 
 j2objcConfig {
-    filenameCollisionCheck false
+    // Per https://github.com/google/j2objc/blob/0.9.8.1/guava/README.txt
+    translatePattern {
+        exclude 'com/google/common/reflect/ClassPath.java'
+    }
+
     // For now, default config links in prebuilt Guava; that defeats the point.
     translateJ2objcLibs.remove("j2objc_guava.jar")
     linkJ2objcLibs.remove("guava")
 
+    // Approximation of https://github.com/google/j2objc/blob/0.9.8.1/guava/Makefile
+    translateArgs "--segmented-headers",
+            "--extract-unsequenced",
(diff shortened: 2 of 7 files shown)


advayDev1:
@brunobowden  - please do #431 first.

brunobowden:
I'll look at this after #431 is merged in.

advayDev1:
@brunobowden - please read #419 - you can already specify `compile X` from your shared build.gradle and it will work if that is the sole use of library X in all of your j2objc projects.  but if say Guava is used from multiple j2objc projects, each such project downloading, translating, and linking Guava in cannot work; when you link the various projects together, linking will fail with duplicate definitions.  therefore, you must define Guava as a separate project that just translates and compiles Guava on its own, which generates a static library.  you then depend on that static library from your various projects.

advayDev1:
ptal

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
```diff
@@ -277,6 +277,18 @@ class J2objcConfig {
...
 
+    /**
+     * Additional native libraries that are part of the j2objc distribution.
+     * <p/>
+     * For example:
+     * <pre>
+     * linkJ2objcLibs = ["guava", "jsr305"]
+     * </pre>
+     */
+    // J2objc default libraries, from $J2OBJC_HOME/lib/..., plus '.a' extension.
```
brunobowden: `without '.a' extension`

systemTests/externalLibrary1/third_party_guava/build.gradle
```diff
@@ -0,0 +1,56 @@
...
+
+apply plugin: 'java'
+apply plugin: 'com.github.j2objccontrib.j2objcgradle'
+
+repositories {
+    jcenter()
+}
+
+dependencies {
+    compile 'com.google.j2objc:j2objc-annotations:0.9.8'
```
brunobowden: These are dependencies for Guava?

If so, then list comment for that. How should this be handled in the future? Is this a list that needs to be manually maintained for Guava or can it be handled automagically?

brunobowden: Add comment explaining all of this.

systemTests/externalLibrary1/third_party_guava/build.gradle
```diff
@@ -0,0 +1,56 @@
...
+    // j2objcTranslation causes the entire specified library
+    // to be translated and built into a standalone Objective C library.
+    // j2objcTranslation must always be passed a *source* jar,
+    // therefore, note the ':sources' classifier.
+    j2objcTranslation 'com.google.guava:guava:18.0:sources'
+}
+
+j2objcConfig {
+    translatePattern {
+        exclude 'com/google/common/reflect/ClassPath.java'
```
brunobowden: Add comment to explain that the `reflect` APIs aren't supported. Is this a TODO for the future?

systemTests/externalLibrary1/third_party_guava/build.gradle
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    // For now, default config links in prebuilt Guava; that defeats the point.
+    translateJ2objcLibs.remove("j2objc_guava.jar")
+    linkJ2objcLibs.remove("guava")
+
+    translateArgs "--segmented-headers",
```
brunobowden: Presumably these settings came from the j2objc build? If so, reference the original source.

systemTests/run-all.sh
```diff
@@ -40,5 +40,8 @@ runTest multiProject1
 
 # Two gradle projects, `extended` depends on `base`. Both of them depend
 # on project `third_party_gson`, which fully translates and compiles an
-# external library (Google's Gson). This library is used in both `extended` and `base`.
+# external library (Google's Gson); and also `third_party_guava` which
+# does the same for Guava. These libraries are used in both `extended` and `base`.
+mv localJ2objcDist/j2objcDist/include/com/google/common localJ2objcDist/j2objcDist/include/com/google/common-bak
```
brunobowden: What's the purpose of the copying back and forth? Backup of some kind? Please add comments explaining this.

brunobowden:
Thanks for the "sole use of library X" explanation. That makes sense.

LGTM based on comments that you should still add

systemTests/externalLibrary1/third_party_guava/build.gradle
advayDev1: At some point we have to come to terms with the limits of automating building of external libraries from their source alone - source jars are not a complete specification of the resulting library.  Imagine if someone just had the Java source of our code, they wouldn't know our dependencies or importantly that we rely on BuildInfo.java being generated via git-build-info.  Same if an external library has resource files, or a special build step (like auto-gened code), etc.  I don't see a way of automating this in a general fashion.

I will add comments as to how I determined this though.

brunobowden: Completely understand about not fully automating this. Depending on how far
we get with this, it would be good to automate at least the common
libraries and then provide the facility where others can add automation for
libraries they care about. Also, if J2ObjC eventually supports translation
of compiled jars (with the performance hit) that would also broaden support.

systemTests/externalLibrary1/third_party_guava/build.gradle
advayDev1: Added comment instead indicating the reference to https://github.com/google/j2objc/blob/master/guava/README.txt

systemTests/externalLibrary1/third_party_guava/build.gradle
advayDev1: done

systemTests/run-all.sh
advayDev1: i have to rename them away from include/com/google/common otherwise the guava build uses those headers instead of the ones it just generated.  (as in without this rename, i'm not actually verifying *our* Guava build worked)

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy

systemTests/externalLibrary1/third_party_guava/build.gradle
advayDev1: I don't think we can get to the point of automating creation of separate projects (such as third_party_gson|guava) even for common libraries ever within the confines of a Gradle plugin.  A plugin simply can't create such a subproject dynamically (I've looked before).

Instead, I think the best we can do, is provide in the other e2e repo, sample build.gradle files for these common libraries that people can - most likely - simply drop in and use.

confile:
@advayDev1 Could you please add to the docs how this feature works if it does not do automatically?

advayDev1:
yes I will #427 when i get some time

PR #439, opened 2015-08-29
Added jre_emul to default linkJ2objcLibs

confile:


brunobowden:
@advayDev1 - this seems like it would be needed? Wanted to get your LGTM on this.

advayDev1:
Does not LGTM.
https://github.com/j2objc-contrib/j2objc-gradle/blob/master/src/main/groovy/com/github/j2objccontrib/j2objcgradle/NativeCompilation.groovy#L299

I didn't keep this in the flags because you should not be able to turn off JRE.

advayDev1:
Please reopen if you disagree.

brunobowden:
Got it. I was a little surprised it worked without that but your code
reference makes sense.

confile:
It is a little bit confusing. @advayDev1 At least add a comment in the config to make this clear.

PR #444, opened 2015-08-31
DEFAULT_ENABLED_ARCHS="ios_arm64,ios_armv7,ios_x86_64"

### LINKED ISSUE #443: Faster Xcode build: smaller default set for enabledArchs
During development state doing ``gradle shared:j2objcXcode`` for working with code changes in the Xcode project appears to be very slow. Even changing a single line and doing incremental compile takes about 2 minutes for my project. With the old version of this plugin I could do it in 3 to 5 seconds which is much more acceptable. 

Here is what I propose. During development stage it does not seems to be necessary to generate a pre compiles lib from your translated code. We should introduce a flag: 

```
fastDevelopment = true
``` 
When this flag is set the plugin translates the code and the ``j2objcXcode`` task adds the translated sources to the pod in Xcode. 

This makes development super fast. I myself worked this way with the old plugin for a while. 

@advayDev1 @brunobowden Looking forward to your comments on this.

brunobowden:
- Speed build by defaulting to a smaller set of enabled architectures
- Remove J2objcConfig.supportedArchs setting
- Common developer use case of modern Mac but new and old iOS devices
- Fixes #443

Notes:
- ios_i386 is rarely needed to run the iOS simulator on a 32-bit Mac
- ios_armv7s was deprecated in Xcode 6 as a “standard architecture”
- See: http://stackoverflow.com/a/25398898/1509221

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
@@ -53,6 +53,13 @@ class J2objcConfig {
 
     final protected Project project
 
+    /**
+     * Default list of architectures to enable. This supports the most commonly
+     * needed architectures: 32-bit and 64-bit iOS builds and running the
+     * iOS simulator on a 64-bit Mac.
+     */
+    public final String DEFAULT_ENABLED_ARCHS ="ios_arm64,ios_armv7,ios_x86_64"
+
     J2objcConfig(Project project) {
         assert project != null
         this.project = project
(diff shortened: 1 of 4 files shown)


brunobowden:
@confile, @advayDev1 - this'll help speed up the default build for most people by dropping little needed architectures (power users can add them back as needed). It also explains how to disable the release build as well.

confile:
@brunobowden What do you think about the ``fastDevelopment = true`` flag I suggested? #443

confile:
LGTM

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
```diff
@@ -619,6 +611,10 @@ class J2objcConfig {
 
         if (!translateOnlyMode) {
             Utils.requireMacOSX('Native Compilation of translated code task')
+            if (releaseEnabled && debugEnabled) {
+                project.logger.warn("Local development builds can be sped up by disabling release builds:")
```
advayDev1: Warnings show up every time.  There is nothing wrong with building everything, thus this should not be a warning.

advayDev1:
I do not agree we should do this.
The defaults should be configured to build the entire app, as each user has configured in supportedArchs.

I think maybe what you meant is to change the `DEFAULT_SUPPORTED_ARCHS` not `ENABLED`.  That I think would make sense.  (Remember `supported` means what the project builds if the user hasn't customized it, while `enabled` is the subset of supported you want to build for speed alone.)

Otherwise it is impossible to build a full fat library without a local config change - that seems wrong.

brunobowden:
@advayDev1 - we don't have "DEFAULT_SUPPORTED_ARCHS" but only "ALL_SUPPORTED_ARCHS". I certainly I think we should remove 'ios_armv7s' from ALL_SUPPORTED_ARCHS, as it's such little value. I also think we should remove 'ios_i386' as I think it's so unlikely that someone will develop on an i386 machine.

So would be ok with this?

    static final String[] ALL_SUPPORTED_ARCHS = ['ios_arm64', 'ios_armv7', 'ios_x86_64']

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
brunobowden: I understand your point. Downgrade to info... but at that level users are unlikely to see this helpful information.

confile:
I agree with @brunobowden

advayDev1:
I think a much simpler solution (and proper use of the words supported and enabled) is:
https://github.com/j2objc-contrib/j2objc-gradle/pull/448

I'm sure there's a test missing somewhere, but please check it out.

advayDev1:
closed as dupe to #448

PR #448, opened 2015-09-04
Default to only modern architectures

advayDev1:
Better fix for #443 (compare vs. #444).

I think this is much simpler, makes regular users faster, and does not take away power users' ability to use all architectures.

cc: @confile

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
@@ -377,19 +377,28 @@ class J2objcConfig {
      * The three ios_arm* architectures are for iPhone and iPad devices, while
      * ios_i386 and ios_x86_64 are for their simulators.
      * <p/>
-     * Adding an unrecognized new architecture here will fail.
+     * By default, only common modern iOS architectures will be built:
+     * ios_arm64, ios_armv7, ios_x86_64.  You may choose to add any of the remaining
+     * entries from NativeCompilation.ALL_SUPPORTED_ARCHS (ios_i386 and ios_armv7s)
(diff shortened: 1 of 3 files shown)


src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
```diff
@@ -377,11 +377,13 @@ class J2objcConfig {
      * The three ios_arm* architectures are for iPhone and iPad devices, while
      * ios_i386 and ios_x86_64 are for their simulators.
      * <p/>
-     * Adding an unrecognized new architecture here will fail.
+     * By default, only common modern iOS architectures will be built:
+     * ios_arm64, ios_armv7, ios_x86_64.  You may choose to add any of the remaining
+     * entries from NativeCompilation.ALL_SUPPORTED_ARCHS to support all possible
```
brunobowden: It would be better to enumerate the other alternatives as well:

    ...
    entries from NativeCompilation.ALL_SUPPORTED_ARCHS (ios_i386 and ios_armv7s)
    to support all possible iOS architectures. Listing any new architectures outside of
    ALL_SUPPORTED_ARCHS will fail the build.

brunobowden:
@advayDev1 - I agree this is much better. Thanks for the suggestion.

brunobowden:
LGTM

brunobowden:
You'll just need to fix up the unit tests... you can copy some of the changes from my PR.

confile:
LGTM

confile:
@advayDev1 You should add the overview comment that @brunobowden had in his request. It shows which phone type has which architecture.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy

advayDev1:
@confile  - done

Issue #430, opened 2015-08-27
Wildcard matching in prefixes.properties


brunobowden:
Support was added in J2ObjC version 0.9.8.1, so the plugin needs to support this too. The change is primarily needed to get the correct test names for TestTask.groovy.

PR #450, opened 2015-09-08
FAQ Support for package-prefix wildcards added

confile:


confile:
@advayDev1 ptal

FAQ.md
```diff
@@ -263,6 +263,11 @@ com.example.dir: Ced
 com.example.dir.subdir: Ceds
 ```
 
+As of j2objc verion 0.9.8 you can also use wildcard in your prefix file.
```
advayDev1: Capitalize 'J2ObjC'

Say plural 'wildcards' and 'prefixes'.

Add sentence `This will map packages such as ``com.example.dir`` and ``com.example.dir.subdir`` both to Ceds.`

confile:
done

FAQ.md
```diff
@@ -263,6 +263,11 @@ com.example.dir: Ced
 com.example.dir.subdir: Ceds
 ```
 
+As of J2ObjC verion 0.9.8 you can also use wildcards in your prefixes file. This will map packages such ascom.example.dirandcom.example.dir.subdirboth to Ceds.
```
advayDev1: add a new line before 'This will...'

missing spaces between words.
'such as com.example.dir and com.example.dir.subdir'

advayDev1:
after you make that fix, AND combine your commits into a single commit, LGTM and I will merge.

confile:
done

advayDev1:
there's still a space missing but i'll take care of it

brunobowden:
The `getTestNames` function in TestTask.groovy also needs to updated to work correctly for wildcards: #430. Until then, the FAQ will be misleading in that wildcards will cause the tests to fail. I'll fix this next week.

PR #455, opened 2015-09-14
Allow mapping of Java source files.

advayDev1:


confile:
@advayDev1 whar is this good for compared with ``translateSourcepaths``.

advayDev1:
I'm currently using it for the following:
I have a third-party library which almost meets my needs (and comes with a source jar).
I need to however alter some native implementations substantially.  I map just those 3 files to brand new Java files. 

translateSourcepaths works only if I was using this library in exactly one other project AND i was willing to use --build-closure.  Instead I chose to build this third-party library using the new `j2objcTranslation` feature I added in #431.  I've gotten rid of all --build-closure in all of my projects using the new dependency features and it speeds up my build by a factor about 5 when I change a single file (in the past build-closure meant everything had to be rebuilt)

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/CycleFinderTask.groovy
```diff
@@ -45,7 +45,7 @@ class CycleFinderTask extends DefaultTask {
         if (J2objcConfig.from(project).translatePattern != null) {
             ret = allFiles.matching(J2objcConfig.from(project).translatePattern)
```
brunobowden: Change ret => result, I think it's more legible.

brunobowden: Actually, it seems like `ret` is a convention so shall we stick with that?

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
```diff
@@ -585,4 +585,16 @@ class Utils {
...
     }
+
+    static FileCollection mapSourceFiles(Project proj, FileCollection files,
+                                         Map<String, String> sourceMapping) {
+        for (String before : sourceMapping.keySet()) {
+            if (files.contains(proj.file(before))) {
+                // Replace the before file with the after file.
+                files = files.minus(proj.files(before)).plus(
+                        proj.files(sourceMapping.get(before)))
+            }
```
brunobowden: Produce an error if a sourceMapping doesn't exist? Likely that indicates a deeper error.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
```diff
@@ -585,4 +585,16 @@ class Utils {
...
         return proj.mkdir(path)
     }
+
+    static FileCollection mapSourceFiles(Project proj, FileCollection files,
+                                         Map<String, String> sourceMapping) {
+        for (String before : sourceMapping.keySet()) {
+            if (files.contains(proj.file(before))) {
+                // Replace the before file with the after file.
+                files = files.minus(proj.files(before)).plus(
+                        proj.files(sourceMapping.get(before)))
```
brunobowden: How efficient is this for large sets of files? I'm sure it isn't efficient but is it in practice an issue?

brunobowden:
LGTM

Yet another example of why cycleFinder and translate should have a shared base class ;-)

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/CycleFinderTask.groovy
advayDev1: yep it is

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
advayDev1: it would mean a failure of the JDK.  let's not verify basic Java functionality.  (the key definitively came from sourceMapping 4 lines above).

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
advayDev1: i'm using this for large libraries and have noticed no issue.  the slowdown is proportional to the number of sourceMappings, which we expect to be very small.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
brunobowden: Sounds like that should be an assert then ;-)

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
brunobowden: Make sense that performance will not be an issue.

PR #470, opened 2015-09-29
filenameCollisionCheck defaults to false

### LINKED ISSUE #467: filenameCollisionCheck should default to the value of no-package-directories
Currently filenameCollisionCheck defaults to true, even though no-package-directories defaults to false.  This is incorrect.  Only if no-package-directories is set should filenameCollisionCheck be true (by default).

brunobowden:
Defaults to True when translateArgs contains “—no-package-directories”

Fixes #467

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
@@ -292,11 +292,20 @@ class J2objcConfig {
 
     // TODO: warn if different versions than testCompile from Java plugin
     /**
-     * Makes sure that the translated filenames don't collide.
+     * Force filename collision check so prohibit two files with same name.
      * <p/>
-     * Recommended if you choose to use --no-package-directories.
+     * This will automatically be set to true when translateArgs contains
+     * '--no-package-directories'. That flag flattens the directory structure
+     * and will overwrite files with the same name.
      */
-    boolean filenameCollisionCheck = true
+    boolean forceFilenameCollisionCheck = false
+
+    // All access to filenameCollisionCheck should be done through this function
+    boolean getFilenameCollisionCheck() {
+        if (translateArgs.contains('--no-package-directories'))
+            return true
+        return forceFilenameCollisionCheck
+    }
 
     /**
      * Sets the filter on files to translate.
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/CycleFinderTask.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/CycleFinderTask.groovy
@@ -86,7 +86,7 @@ class CycleFinderTask extends DefaultTask {
     Map<String, String> getTranslateSourceMapping() { return J2objcConfig.from(project).translateSourceMapping }
 
     @Input
-    boolean getFilenameCollisionCheck() { return J2objcConfig.from(project).filenameCollisionCheck }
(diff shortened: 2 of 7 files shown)


brunobowden:
@advayDev1 - PTAL

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
```diff
@@ -294,9 +294,18 @@ class J2objcConfig {
     /**
      * Makes sure that the translated filenames don't collide.
      * <p/>
-     * Recommended if you choose to use --no-package-directories.
+     * Set to True when translateArgs contains '--no-package-directories'.
+     * The flag flattens the directory structure and overwrites files with
+     * the same names.
      */
-    boolean filenameCollisionCheck = true
+    boolean filenameCollisionCheck = false
```
advayDev1: would be clearer to call this 'forceFilenameCollisionCheck' and then change the javadoc.
it is not the case that this variable should be set to true, in fact we are just treating it as if it was true when no-package-directories is present.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
```diff
@@ -294,9 +294,18 @@ class J2objcConfig {
...
+
+    // All access to filenameCollisionCheck should be done through this function
+    boolean filenameCollisionCheckIsSet() {
```
advayDev1: name getFilenameCollisionCheck() if you do the comment above.  that way you are also using the gradle property naming convention (get/set prefixes), and accesses to `.filenameCollisionCheck` will automatically call this method

'isSet' is somewhat ambiguous - i thought at first this was checking if the user had the set the variable (as in not accepted a default value).

advayDev1:
lgtm with comments

brunobowden:
So Clang allows compiles with the directory structure from the command line but not through Xcode?

Just waiting for CI to complete and then I'll merge.

brunobowden:
Merging...
The Xcode build now takes 18 minutes which is a long time.

advayDev1:
Xcode actually passes every individual file to clang separately.  So I think it is xcode's 'choice' not to support directories.  Also what is taking 18 minutes?  Which task?

brunobowden:
The recent Xcode builds take 18-20 minutes:
https://travis-ci.org/j2objc-contrib/j2objc-gradle/jobs/82959080

Adding the Guava was responsible, you can see that from the build history:
https://travis-ci.org/j2objc-contrib/j2objc-gradle/builds

advayDev1:
OH: the continuous integration builds are 18 minutes.  Not like your dev Xcode builds for your project.

Yeah, that is true.  But I don't see how we can really exercise the plugin unless it is building multiple projects with dependencies - Guava's our best bet for our various features really.

advayDev1:
This is the first time I'm actually comfortable that changes aren't breaking the plugin.  Unit-tests can only do so much with the amount of inter-tool interaction we have

brunobowden:
The level of complexity in the system is considerable. It's a lot more
complex that your typical library.

advayDev1:
Is 'it' our plugin? Or guava as the system-under-test?

Either way I think that's the whole benefit of Travis.  We don't need to serially spend time (or our local CPU cycles) waiting those 18 minutes.  You push and do other work in parallel.  Like with any presubmit, devs usually have enough pipelined work that a remote system thoroughly testing shouldn't delay work.

Also Travis has a feature such that for example if you are strictly changing only CONTRIBUTING or README, etc. add `'[skip ci]` anywhere in the commit, and Travis will auto-green the CL.

PR #474, opened 2015-09-30
Separate test translation and building from production library

advayDev1:


brunobowden:
I'll hold off reviewing it until you ask me.

advayDev1:
yeah hold off for now, i currently have no way of verifying things except pushing to Travis as Xcode 7 is breaking things locally for me

advayDev1:
The next Travis build *should* pass.  But I'm about to lose internet, so if it turns green, please review.  @brunobowden

advayDev1:
Well Travis (unit tests + system tests do pass).  AppVeyor (windows unit tests) aren't. I'm not certain which of your win32-path utils I should use for this error:
https://ci.appveyor.com/project/madvayApiAccess/j2objc-gradle/build/build27.master-yumdtrni

@brunobowden - could you give me a hint there?

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTaskTest.groovy
```diff
@@ -132,20 +191,45 @@ class TranslateTaskTest {
...
-    private static IncrementalTaskInputs genNonIncrementalInputs() {
+    private static IncrementalTaskInputs genNonIncrementalInputs(Project proj) {
         IncrementalTaskInputs incrementalTaskInputs = new IncrementalTaskInputs() {
             @Override
             boolean isIncremental() { return false }
 
             @Override
-            void outOfDate(Action<? super InputFileDetails> outOfDateAction) {}
+            void outOfDate(Action<? super InputFileDetails> outOfDateAction) {
+                proj.files('src/**/*.java').each { final File file ->
```
brunobowden: The Windows build failed. I think the path may need to be altered to work on Windows. Try and see if the following works:

```
String filesPath = 'src/**/*.java'
if (Utils.isWindows()) {
    filesPath = 'src\**\*.java'
}
proj.files(filesPath).each { final File file ->
...
```

Build failure:
https://ci.appveyor.com/project/madvayApiAccess/j2objc-gradle/build/build27.master-yumdtrni

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTaskTest.groovy
```diff
@@ -132,20 +189,46 @@ class TranslateTaskTest {
...
+    private static IncrementalTaskInputs genNonIncrementalInputs(Project proj) {
         IncrementalTaskInputs incrementalTaskInputs = new IncrementalTaskInputs() {
             @Override
             boolean isIncremental() { return false }
 
             @Override
-            void outOfDate(Action<? super InputFileDetails> outOfDateAction) {}
+            void outOfDate(Action<? super InputFileDetails> outOfDateAction) {
+                proj.files(Utils.isWindows() ? 'src\\**\\*.java' : 'src/**/*.java').each {
+                    final File file ->
```
brunobowden: Can you add warning logs (that show up in the logs for the CI builds on all platforms). Have it show the input to proj.files(...) and the value of the iterator on each file. That'll be fairly verbose but I think that'll help us understand what's going wrong here.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
```diff
@@ -289,6 +290,13 @@ class J2objcConfig {
     // TODO: auto add libraries based on java dependencies, warn on version differences
     List<String> linkJ2objcLibs = ['guava', 'j2objc_main', 'javax_inject', 'jsr305']
```
brunobowden: linkJ2objcLibs => linkJ2objcMainLibs

I think it's better to be explicit about Main and Test at all places in the code, so they aren't confused with each other.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/NativeCompilation.groovy
```diff
@@ -69,7 +69,8 @@ class NativeCompilation {
             // Otherwise the Objective-C plugin skips creation of the compile tasks altogether.
             file("${buildDir}/j2objcHackToForceCompilation").mkdirs()
```
brunobowden: Add "Main" in to string. Repeat for line below also.

j2objcHackToForceCompilation => j2objcHackToForceMainCompilation

src/main/groovy/com/github/j2objccontrib/j2objcgradle/NativeCompilation.groovy
```diff
@@ -209,18 +210,16 @@ class NativeCompilation {
                         sources {
                             objc {
                                 source {
-                                    // Note that contents of srcGenDir are generated by j2objcTranslate task.
-                                    srcDirs "${srcGenDir}", "${buildDir}/j2objcHackToForceCompilation"
+                                    // Note that contents of srcGenMainDir are generated by j2objcTranslate task.
+                                    srcDirs "${srcGenMainDir}", "${buildDir}/j2objcHackToForceCompilation"
```
brunobowden: j2objcHackToForceCompilation => j2objcHackToForceMainCompilation

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/AssembleSourceTask.groovy
```diff
@@ -82,25 +84,16 @@ class AssembleSourceTask extends DefaultTask {
...
             includeEmptyDirs = false
-            from srcGenDir
+            from srcGenMainDir
             into getDestSrcMainObjcDirFile()
-            // TODO: this isn't precise, main source can be suffixed with Test as well.
-            // Would be best to somehow keep the metadata about whether a file was from the
-            // main sourceset or the test sourceset.
-            // Don't copy the test code
-            exclude "**/*Test.h"
-            exclude "**/*Test.m"
```
brunobowden: Glad to get rid of this. It was such a hack.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTask.groovy
```diff
@@ -36,17 +36,16 @@ import org.gradle.api.tasks.incremental.InputFileDetails
 @CompileStatic
 class TranslateTask extends DefaultTask {
 
-    // Source files outside of the Java sourceSets of main and test.
+    // Source files outside of the Java main sourceSet.
     FileCollection additionalSrcFiles
```
brunobowden: Should we rename this additionalMainSrcFiles?

brunobowden: Will we ever need additionalTestSrcFiles?

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTask.groovy
```diff
@@ -36,17 +36,16 @@ import org.gradle.api.tasks.incremental.InputFileDetails
...
 
-    // Source files part of the Java sourceSets of main and test.
+    // Source files part of the Java main sourceSet.
     @InputFiles
-    FileCollection getSrcFiles() {
+    FileCollection getMainSrcFiles() {
         // Note that neither additionalSrcFiles nor translatePattern need
```
brunobowden: This whole comment should be moved up to line 39.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
advayDev1: while in other cases this might apply, it is not the case here.  because library linkage is transitive, all libraries here are linked to both main and test.

also, where such notation is used - I think we should follow Gradle's own convention.
For example your main dependencies are called `compile` and test ones `testCompile`.  `mainCompile` is never used.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/NativeCompilation.groovy

src/main/groovy/com/github/j2objccontrib/j2objcgradle/NativeCompilation.groovy

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTask.groovy
advayDev1: renamed 

so this is the matrix explosion of settings I was taking about.  we /could/ need additionalFiles for potentially every combination of {main, test} x {ios, mac} x {native, simulator} etc etc.

Until we finalize our design for ios vs. mac, I suggest we hold off on adding more variables in this PR - especially until user need is demonstrated.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTaskTest.groovy
advayDev1: done

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTask.groovy
```diff
@@ -183,12 +218,30 @@ class TranslateTask extends DefaultTask {
...
-                        if (removedFileNames.contains(nameWithoutExt)) {
+                        if (removedMainFileNames.contains(nameWithoutExt)) {
                             file.delete()
                         }
                     }
                     // compute the number of translated files
-                    translatedFiles = destFiles.getFiles().size()
+                    translatedFiles += destFiles.getFiles().size()
+                }
+                if (srcGenTestDir.exists()) {
```
brunobowden: This should be extracted as a shared method and then be called twice:

```
deleteRemovedFiles(srcGenMainDir, removedMainFileNames)
deleteRemovedFiles(srcGenTestDir, removedTestFileNames)
```

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTask.groovy
```diff
@@ -130,51 +150,66 @@ class TranslateTask extends DefaultTask {
...
+            logger.debug("Removed main files: " + removedMainFileNames.size())
+            logger.debug("Removed test files: " + removedTestFileNames.size())
 
-            logger.debug("New or Updated files: " + srcFilesChanged.getFiles().size())
-            FileCollection unchangedSrcFiles = originalSrcFiles - srcFilesChanged
-            logger.debug("Unchanged files: " + unchangedSrcFiles.getFiles().size())
+            logger.debug("New or Updated main files: " + mainSrcFilesChanged.getFiles().size())
+            logger.debug("New or Updated test files: " + testSrcFilesChanged.getFiles().size())
+            FileCollection unchangedMainSrcFiles = originalMainSrcFiles - mainSrcFilesChanged
+            logger.debug("Unchanged main files: " + unchangedMainSrcFiles.getFiles().size())
```
brunobowden: I understand why you did it this way but I think it's better to always have the "Main" line of code followed by the "Test" line of code. That allows easier visual inspection if there's a difference and aesthetically it flows better.

```
FileCollection unchangedMainSrcFiles = originalMainSrcFiles - mainSrcFilesChanged
FileCollection unchangedTestSrcFiles = originalTestSrcFiles - testSrcFilesChanged
logger.debug("Unchanged main files: " + unchangedMainSrcFiles.getFiles().size())
logger.debug("Unchanged test files: " + unchangedTestSrcFiles.getFiles().size())
```

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTask.groovy
advayDev1: done

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTask.groovy
```diff
@@ -208,13 +261,46 @@ class TranslateTask extends DefaultTask {
...
+        // we will get duplicate symbol errors.
+        List<String> testTranslateArgs = new ArrayList<>(translateArgs)
+        testTranslateArgs.removeAll("--build-closure")
+        sourcepathDirs = new UnionFileCollection([
+                project.files(Utils.srcSet(project, 'main', 'java').getSrcDirs()),
+                project.files(Utils.srcSet(project, 'test', 'java').getSrcDirs()),
+                project.files(getTranslateSourcepaths()),
+                project.files(getGeneratedSourceDirs())
+        ])
+        doTranslate(sourcepathDirs, srcGenTestDir, testTranslateArgs, testSrcFilesChanged)
```
brunobowden: So this will translate the "Main" code twice? What about duplicating the "Main" code and only translating the "Test" code in the second step? This would reduce the translation but where does this break?

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTask.groovy
```diff
@@ -208,13 +261,46 @@ class TranslateTask extends DefaultTask {
...
+
+        // Translate main code.
+        UnionFileCollection sourcepathDirs = new UnionFileCollection([
+                project.files(Utils.srcSet(project, 'main', 'java').getSrcDirs()),
+                project.files(getTranslateSourcepaths()),
+                project.files(getGeneratedSourceDirs())
+        ])
+        doTranslate(sourcepathDirs, srcGenMainDir, translateArgs, mainSrcFilesChanged)
+
+        // Translate test code.  Tests are never built with --build-closure; otherwise
```
brunobowden: Single space after period.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTask.groovy
advayDev1: ok

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTask.groovy
advayDev1: no.  remember that sourcepath is NOT the list of things being translated (except when using --build-closure and the files in that directory are used transitively).

the list of fiels translated is testSrcFilesChanged in that call

systemTests/run-all.sh
```diff
@@ -24,7 +24,11 @@ function runTest {
     pushd $TEST_DIR
     ./gradlew wrapper
     ./gradlew clean
-    ./gradlew build
+    # If we fail, try again with lots of logging.
+    ./gradlew build || ./gradlew build --debug --stacktrace
```
brunobowden: Can you insert 10 newlines before running the command again? That'll help find where it started in the logs.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTaskTest.groovy
advayDev1: Doesn't work.   proj.files is itself failing, not on an individual file (so log never gets hit).
Adding logs to see if the project dir itself exists now - considering the error message

systemTests/run-all.sh
advayDev1: echoed out a unique string instead - hard to Ctrl-F for 10 newlines

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTask.groovy
advayDev1: done

brunobowden:
I'm not sure what's going on with the Windows build but I believe my instinct is correct about the case. It would be helpful if you could add enough logging so that we can see what it's doing.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
brunobowden: Add comment explaining that this applies to Main and Test builds.

systemTests/run-all.sh
```diff
@@ -24,7 +24,11 @@ function runTest {
     pushd $TEST_DIR
     ./gradlew wrapper
     ./gradlew clean
-    ./gradlew build
+    # If we fail, try again with lots of logging.
+    ./gradlew build || echo FAIL_LOG_FAIL_LOG_FAIL_LOG_FAIL_LOG && ./gradlew build --debug --stacktrace
```
brunobowden: Wrap with "****" as scanning the left hand side of a block of text, it's easier to spot this:

```
**** FAIL_LOG_FAIL_LOG_FAIL_LOG_FAIL_LOG **** 
```

brunobowden: Do that you if you can quote it... if it causes hassle for the script then drop the suggestion.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTask.groovy
brunobowden: Ahh, makes sense. Got it.

systemTests/run-all.sh
advayDev1: used underscores instead of starts for that reason

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTaskTest.groovy
advayDev1: actually i don't see anyway to get logs from our unit tests.  what have you used in the past @brunobowden ?  they all seem piped away to report files we can't access on CI and neither you nor I have a Windows machine :)

brunobowden: Can you access any logs from the machine?

I remember that Info or Debug logs were suppressed... but terrors and
Warnings came through. So abuse the Error logging for this.

advayDev1: its not that.

nothing you output during a unit test is ever displayed in the tes ttask

advayDev1:
@brunobowden - windows fixed.  there were 2 bugs:
0.  Misunderstanding of project.files(...) - it doesn't accept globs, just arrays of individually named files.
1.  MockProjectExec's handling of multiple arguments to project.files

nothing about windows!

advayDev1:
ptal @brunobowden - i think this is good to go

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/MockProjectExec.groovy
```diff
@@ -88,13 +88,14 @@ class MockProjectExec {
...
-                    assert 0 == args.size() || 1 == args.size()
                     if (args.size() == 0 ||  // files()
                         args.first() == null) {  // files(null)
                         return metaMethod?.invoke(delegate, [[] as Object[]] as Object[])
-                    } else {
+                    } else if (args.size() == 1) {
                         // files(ArrayList) possibly, so cast ArrayList to Object[]
                         return metaMethod?.invoke(delegate, [(Object[]) args.first()] as Object[])
+                    } else {
+                        return metaMethod?.invoke(delegate, [args] as Object[])
```
brunobowden: Does `metaMethod?.invoke(delegate, [args] as Object[])` work for the `args.size() == 1` case?

brunobowden:
LGTM

See if you can simplify the line within MockProjectExec. Either way, we're good to go.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/MockProjectExec.groovy
advayDev1: It shouldn't...
L96 is equivalent to invoke(delegate, [[args[0][0], args[0][1], args[0][2]]])
L98 is equivalent to invoke(delegate, [[args[0], args[1], [args[2]]])

PR #479, opened 2015-10-03
version = '0.4.4-alpha-SNAPSHOT'

brunobowden:
I know we'll next go to v0.5.0 but leaving it as the old version until we make that leap.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -40,7 +40,7 @@ group = 'com.github.j2objccontrib.j2objcgradle'
 // A suffix of -SNAPSHOT means this is not an official release of the
 // system, but built from an intermediate source tree.  See
 // CONTRIBUTING.md on how to update this version number.
-version = '0.4.3-alpha'
+version = '0.4.4-alpha-SNAPSHOT'
 
 test {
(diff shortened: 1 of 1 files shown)


advayDev1:
no the correct thing would be to label it `0.4.4-alpha-SNAPSHOT` then.  you have to monotonically increase- and X-SNAPSHOT is always below X.

brunobowden:
version => 0.4.4-alpha-SNAPSHOT

brunobowden:
Can I get an LGTM for this?

advayDev1:
lgtm

advayDev1:
we screwed this up a bit.  the 'bump to snapshot' cl always has to be submitted immediately after the 'bump to released version' cl.  when there are commits in between, namely: 
https://github.com/j2objc-contrib/j2objc-gradle/compare/34e792c40d6c24b904cbc0c877423b1686831429...d9c4bfe50e6fbb3edf67f0934b785e3bf74e4c8f
builds at those commits are falsely attributed to the previous released version (a problem when we debug other users' logs)

PR #480, opened 2015-10-03
Version 0.4.3 readme

advayDev1:


README.md
```diff
@@ -34,7 +34,7 @@ Configure `shared/build.gradle` in your Java only project:
 // File: shared/build.gradle
 plugins {
     id 'java'
-    id 'com.github.j2objccontrib.j2objcgradle' version '0.4.2-alpha'
+    id 'com.github.j2objccontrib.j2objcgradle' version '0.4.3-alpha'
```
brunobowden: Since we can't confirm the end to end build, I'd avoid updating the version number. Instead it would be better to add a temporary note that "Work is in progress for compatibility with Xcdoe 7".

brunobowden:
LGTM with comment

advayDev1:
Not sure what you mean:
 I think your additional comment is definitely a good idea.
But why would we not update the version number? 0.4.2-alpha also does not work with xcode 6 and is not the one in fact being tested in our system tests.

holding off on commiting.

brunobowden:
LGTM but just add the comment about Xcode 7 support.

advayDev1:
done and merging

Issue #466, opened 2015-09-27
j2objc 0.9.8.2 and Xcode 7 not yet supported - "ld: library not found for -lj2objc_main"


andrewfunston:
I've searched a bit and found nothing that seems similar. I've also checked my build environment vs the docs + FAQ and haven't found anything out of line. I'm new to j2objc and was trying to get a minimal build environment running. I followed the recommendations for build environment setup from the j2objc docs.

j2objc version was a pull/build from master from today (9/26/2015), but I tried with the release version ".9.8.2" with the same results.

Sorry for the image of my build.gradle, it kept failing when I had it converted to a .txt, etc. Very odd.

The error is as follows:

:java-core:x86_64DebugJava-core-j2objcStaticLibrary
:java-core:linkDebugTestJ2objcExecutable
ld: library not found for -lj2objc_main
clang: error: linker command failed with exit code 1 (use -v to see invocation)

:java-core:linkDebugTestJ2objcExecutable FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':java-core:linkDebugTestJ2objcExecutable'.
> A build operation failed.
      Linker failed while linking testJ2objc.
  See the complete log at: file:///Users/a/parati/java-core/build/tmp/linkDebugTestJ2objcExecutable/output.txt

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED

![screen shot 2015-09-26 at 10 27 56 pm](https://cloud.githubusercontent.com/assets/4275732/10121390/e2b7499c-649d-11e5-829e-8c368ed71945.png)

advayDev1:
What are the full errors in:
file:///Users/a/parati/java-core/build/tmp/linkDebugTestJ2objcExecutable/output.txt ?

andrewfunston:
I can't attach txt files so I'll have to plop them here, but it says essentially the same thing:

--------------------------------------------------------------------

See file:///Users/a/parati/java-core/build/tmp/linkDebugTestJ2objcExecutable/output.txt for all output for linkDebugTestJ2objcExecutable.
linking testJ2objc failed.
ld: library not found for -lj2objc_main
clang: error: linker command failed with exit code 1 (use -v to see invocation)

Finished linkDebugTestJ2objcExecutable, see full log file:///Users/a/parati/java-core/build/tmp/linkDebugTestJ2objcExecutable/output.txt.

[...]

PR #483, opened 2015-10-05
Support Xcode 7: Refactor of native compilation to support separate ios vs. osx librar…

advayDev1:
…ies.

Issue #466

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
@@ -288,7 +288,7 @@ class J2objcConfig {
      */
     // J2objc default libraries, from $J2OBJC_HOME/lib/..., without '.a' extension.
     // TODO: auto add libraries based on java dependencies, warn on version differences
-    List<String> linkJ2objcLibs = ['guava', 'j2objc_main', 'javax_inject', 'jsr305']
+    List<String> linkJ2objcLibs = ['guava', 'javax_inject', 'jsr305']
 
     /**
      * Additional native libraries that are part of the j2objc distribution to link
@@ -296,7 +296,7 @@ class J2objcConfig {
      */
     // J2objc default libraries, from $J2OBJC_HOME/lib/..., without '.a' extension.
     // TODO: auto add libraries based on java dependencies, warn on version differences
-    List<String> linkJ2objcTestLibs = ['junit', 'mockito']
+    List<String> linkJ2objcTestLibs = ['junit', 'mockito', 'j2objc_main']
 
     // TODO: warn if different versions than testCompile from Java plugin
     /**
@@ -416,9 +416,9 @@ class J2objcConfig {
      * <p/>
      * By default, only common modern iOS architectures will be built:
      * ios_arm64, ios_armv7, ios_x86_64.  You may choose to add any of the remaining
-     * entries from NativeCompilation.ALL_SUPPORTED_ARCHS (ios_i386 and ios_armv7s)
+     * entries from NativeCompilation.ALL_IOS_ARCHS (ios_i386 and ios_armv7s)
      * to support all possible iOS architectures. Listing any new architectures outside of
-     * ALL_SUPPORTED_ARCHS will fail the build.
+     * ALL_IOS_ARCHS will fail the build.
      * <p/>
      * Removing an architecture here will cause that architecture not to be built
      * and corresponding gradle tasks to not be created.
@@ -432,7 +432,7 @@ class J2objcConfig {
      * <li>'ios_i386' => iOS Simulator on 32-bit OS X
      * <li>'ios_x86_64' => iOS Simulator on 64-bit OS X
      * </ul>
-     * @see NativeCompilation#ALL_SUPPORTED_ARCHS
+     * @see NativeCompilation#ALL_IOS_ARCHS
      */
     // Public to allow assignment of array of targets as shown in example
     List<String> supportedArchs = ['ios_arm64', 'ios_armv7', 'ios_x86_64']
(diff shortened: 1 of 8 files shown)


advayDev1:
@brunobowden - you might as well ptal.

good news: xcode 7 and unit-tests-on-osx seem to be working fine.
bad news: travis xcode 7 images are bound to java 8, where our unit-tests still won't succeed.
https://travis-ci.org/j2objc-contrib/j2objc-gradle/jobs/83634361

once we upgrade to j2objc 0.9.8.2.1 - xcode 6 no longer works at all (i think this is an acceptable tradeoff, and i do not believe we should do any further coding to make it work with xcode 6 and 7).

i have a mind just to disable the failing unit test - we haven't been able to figure it out over the last 2 months, and xcode 7 is more important.  thoughts?

advayDev1:
I've decided simply to @Ignore that one test and get on with things.  It isn't a regression, and we now have system tests to really exercise our dependency code anyway.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/NativeCompilation.groovy
```diff
@@ -15,23 +15,27 @@
  */
 
 package com.github.j2objccontrib.j2objcgradle
-
```
brunobowden: We have a blank line gap in all other files. I've seen this happen before for your PRs... is the editor doing something automatically?

advayDev1: yeah IntelliJ being annoying

src/main/groovy/com/github/j2objccontrib/j2objcgradle/NativeCompilation.groovy
```diff
@@ -39,6 +43,84 @@ class NativeCompilation {
...
+                        clangArgs += simulatorClangArgs
+                        linkerArgs += ["-L$j2objcPath/lib"]
+                        break
+                    case TargetSpec.TARGET_OSX:
+                        if (Utils.j2objcHasOsxDistribution(project)) {
+                            linkerArgs += ["-L$j2objcPath/lib/macosx"]
+                        } else {
+                            String msg = "J2ObjC distribution at $j2objcPath lacks a lib/macosx directory.\n" +
+                                         "Please update to J2ObjC 0.9.8.2.1 or higher; earlier versions will\n" +
+                                         "not work correctly with Xcode 7 or higher."
```
brunobowden: Make this an exception rather than a warning. We don't care about Xcode 6 and we're comfortable making a breaking change.

advayDev1: appropriate changelog added for this breaking change as well

systemTests/prep.sh
```diff
@@ -14,21 +14,24 @@
...
 # limitations under the License.
 
-J2OBJC_VERSION=0.9.8.1
+J2OBJC_VERSION=0.9.8.2.1
 mkdir localJ2objcDist
 mkdir common
 
 # Fail if any command fails
-set -e
+set -ev
```
brunobowden: Another good flag to use it `-u`, which treats **unset variables as an error when substituting**.

advayDev1: good to know, but in this case i think we're fine as there are no conditional settings

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
```diff
@@ -592,6 +594,7 @@ class J2objcConfig {
         extraNativeLibs.add(spec)
     }
 
+    String minIosVersion = '8.3'
```
brunobowden: How should we set this? The old version of the previous operating system, e.g. 8.0? Or something else. My vote is 8.0 for the time being.

brunobowden: This is also deserves some JavaDoc as well.

advayDev1: This is not a functional change; we forced it to be fixed before now it is variable;
it has been 8.3 from the beginning of the plugin.  I'd suggest we do not change it.

systemTests/run-all.sh
```diff
@@ -48,6 +48,6 @@ runTest multiProject1
 # does the same for Guava. These libraries are used in both `extended` and `base`.
 # We must rename the include directory while this test runs, otherwise the
 # code builds against the translated Guava headers provided in the j2objc dist.
-mv localJ2objcDist/j2objcDist/include/com/google/common localJ2objcDist/j2objcDist/include/com/google/common-bak
+mv localJ2objcDist/j2objc-0.9.8.2.1/include/com/google/common localJ2objcDist/j2objc-0.9.8.2.1/include/com/google/common-bak
```
brunobowden: Extract J2OBJC_VERSION as an environment variable to make this easier to maintain. I'm a little concern that this is a step backwards in updates... but it does allow you to more easily test switching between multiple j2objc versions.

brunobowden:
LGTM - this code already looks much better organized compared to what it was before. I'll experiment with this tomorrow.

systemTests/run-all.sh
advayDev1: done

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
advayDev1: added javadoc

advayDev1:
cool, as soon as i get a green travis with these changes i will submit.  thanks!

PR #485, opened 2015-10-05
FAQ: missing i386 architecture

advayDev1:


advayDev1:
@brunobowden ptal.

cc @maxbritto

maxbritto:
Nice :)

FAQ.md
```diff
@@ -506,3 +507,23 @@ local.properties (the Mac OS X developers should not use this):
...
+If you see a message similar to:
+```
+ld: warning: ignoring file /PATH/j2objcOutputs/lib/iosDebug/libPROJECT-j2objc.a,
+missing required architecture i386 in file /PATH/j2objcOutputs/lib/iosDebug/libPROJECT-j2objc.a
+(3 slices)
+```
+and additionally get linker errors, you are not building all the neccessary architectures.
+
+By default (for performance), we build only modern iOS device and simulator architectures.
+If you need i386 for older simulators, add the following to your build.gradle file:
```
brunobowden: Make it clearer:

    older simulators (iPhone 5, 5c and earlier devices), add the following...

brunobowden:
LGTM

FAQ.md

PR #486, opened 2015-10-05
document xcode 7 support

advayDev1:


brunobowden:
Let's hold off until we can confirm someone got an end to end build working. I'm working on that now.

advayDev1:
@brunobowden: We have 2 people with an e2e build working btw: myself and @maxbritto

brunobowden:
LGTM then

PR #489, opened 2015-10-05
Test-only dependencies; test source translation and execution

### LINKED ISSUE #465: Add 'test' versions of dependency configurations
Some projects, like Apache ones, have a -test-sources.jar too.  Add a j2objcTranslationTest dependency type for these.

advayDev1:
- j2objcTestLinkage links to a j2objc project only for your test source-set
- j2objcTestTranslation translates and executes a unit test source jar

Fixes #465

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/DependencyConverter.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/DependencyConverter.groovy
@@ -52,7 +52,9 @@ class DependencyConverter {
             'com.google.guava:guava',
             'junit:junit',
             'org.mockito:mockito-core',
-            'com.google.j2objc:j2objc-annotations']
+            'com.google.j2objc:j2objc-annotations',
+            'org.hamcrest:hamcrest-core',
+            'com.google.protobuf:protobuf-java']
 
     DependencyConverter(Project project, J2objcConfig j2objcConfig) {
         this.project = project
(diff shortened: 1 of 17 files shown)


advayDev1:
@brunobowden ptal.

other than documenting dependencies, this is the last thing i will have time for 0.5.0

src/main/groovy/com/github/j2objccontrib/j2objcgradle/DependencyConverter.groovy
```diff
@@ -60,43 +61,53 @@ class DependencyConverter {
...
-            visitExternalModuleDependency(dep as ExternalModuleDependency)
+            visitExternalModuleDependency(dep as ExternalModuleDependency, isTest)
         } else {
             // Everything else
-            visitGenericDependency(dep)
+            visitGenericDependency(dep, isTest)
+        }
+    }
+
+    protected void failOnBuildClosureForTests(boolean isTest, Dependency dep) {
```
brunobowden: Stick with the parameter convention: Dependency first, isTest as second parameter.

brunobowden:
I feel like this should have something in the way of tests.

With j2objcTestTranslation, how is it enabled / disabled? I'm thinking of the case where a library may or may not have a test-sources.jar. I'm wondering if it should download and apply the test translation if and only if the test-sources.jar exists. Then if it doesn't exist, that part of the build is just dropped.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/DependencyConverter.groovy
advayDev1: done

advayDev1:
@brunobowden ptal.
* Added tests for testCompile and j2objcTestLinkage to multiProject1.
multiProject1 system test also checks that duplicate transitive dependencies function correctly.

* Also updated the libraries for 0.9.8.2.1; fixes #492

brunobowden:
LGTM

Is there a way to make the code more clear for the tests? I'm wondering if the joiner should be based on the class, so you'd get a result like aEXTbEXTc, or aBASEbBASEc, which makes the context more clear.

advayDev1:
Ok - done

PR #490, opened 2015-10-06
Dependency management documentation

### LINKED ISSUE #427: Document auto dependencies

advayDev1:
Fixes #427


advayDev1:
@brunobowden 

Well here's a start.  Not sure what else would be useful.  Probably better for someone who didn't write the code itself to think what needs emphasis :)

dependencies.md
```diff
@@ -0,0 +1,255 @@
+This document explains how to configure dependencies between multiple projects, libraries, etc.
+
+# Guide
+This guide explains common use cases, but skips some of the corner-cases and technical detail.
+See the [Reference](#reference) for that additional information.
+
+## TL/DR
+If simple enough, your dependencies may be configurable automatically.  First try adding:
+```gradle
+   autoConfigureDeps true
```
brunobowden: We have a convention of putting example with j2objcConfig blocks to make the structure clear:

    j2objcConfig {
        autoConfigureDeps true
    }

dependencies.md
```diff
@@ -0,0 +1,255 @@
...
+
+## Using built-in libraries only
+__Requirements__
+* One or more j2objc Java projects
+* Only built-in libraries
+
+If you have one or more shared Java J2ObjC Gradle project,
+and depend only on libraries provided in the J2ObjC distribution (see `J2objcConfig.translateJ2objcLibs`
+in your version of the plugin), no further dependency configuration is required beyond that described
+in [README.md](README.md) and the `autoConfigureDeps true` line.
```
brunobowden: Since this set doesn't change that much. I think it would be easier to enumerate them here. Then if it changes in the future, note that (only supported up to 0.9.8) or similar.

brunobowden: Besides, they can't read `J2objcConfig.translateJ2objcLibs` in a compiled version of the plugin.

dependencies.md
```diff
@@ -0,0 +1,255 @@
...
+    testMinExpectedTests 0
+    finalConfigure()
+}
+```
+
+If the library also provides a test source Jar, also add for example:
+```gradle
+dependencies {
+    // Unfortunately: Gson doesn't actually have a test-sources.jar.
+    j2objcTestTranslation 'com.google.code.gson:gson:2.3.1:test-sources'
```
brunobowden: Is there a good working example for this that has a test-sources.jar?

dependencies.md
```diff
@@ -0,0 +1,255 @@
...
+You can see many examples of configuring builds for third-party libraries
+[here](https://github.com/madvay/j2objc-common-libs-e2e-test/tree/master/libraryBuilds).
+
+## Multiple projects and external libraries
+__Requirements__
+* Multiple j2objc Java projects: 'shared' depends on 'base' 
+* Simple third-party libraries
+
+With multiple Java projects and multiple external libraries, break up the problem.
+First add the usual `compile project(':base')` stanza to `shared/build.gradle`,
```
brunobowden: stanza => line

1. This is easier for most people to understand
2. A stanza is a *group* of lines (4 lines for some Greek and Latin meters), which suggests more than a single line

dependencies.md
```diff
@@ -0,0 +1,255 @@
...
+    // ...
+    finalConfigure()
+}
+```
+
+```gradle
+// After: shared/build.gradle
+dependencies {
+    compile project(':base')
+    // MODIFY: You must add a specific version number, if you did not have one before.
```
brunobowden: Don't you also need a `third-party-gson/build.gradle` file? Then you'd move the "You must add a specific version number..." line to that file.

dependencies.md
```diff
@@ -0,0 +1,255 @@
...
+
+## TL/DR
+If simple enough, your dependencies may be configurable automatically.  First try adding:
+```gradle
+   autoConfigureDeps true
+```
+at the top of your Gradle project's `j2objcConfig` section.  If that fails, follow one of the
+sections below.
+
+## Using built-in libraries only
```
brunobowden: I'd normally suggest two blank lines between each instruction so that it's more obvious where the breaks are during editing. Up to you though.

dependencies.md
```diff
@@ -0,0 +1,255 @@
...
+risk of duplicate symbols.
+
+### j2objcTranslationClosure
+Use this only with a Jar of Java source files.  These files will be provided to the j2objc
+translator with `--build-closure`; where your main source set relies on classes defined
+in this source Jar, j2objc will additionally translate that class (and all of its transitive
+dependencies).  The plugin will include the result in your project's own library.
+
+This is only safe to use when you are building a single J2ObjC Java Gradle project.
+If you have multiple projects, when the libraries for those projects are linked together,
```
brunobowden: Might be easier to explain as "diamond shape linkage" where projectA depends on two projects, which both depend on the same project => leads to duplicate symbols error.

dependencies.md
```diff
@@ -0,0 +1,255 @@
...
+## Dependency configurations
+
+### compile
+If `autoConfigureDeps` is false, has no effect related to J2ObjC.  When true,
+* for a project, creates a `j2objcLinkage` dependency on the project.
+* for an external module already included in the J2ObjC distribution, has no additional effect (this will work automatically).
+* for any other external module, creates a `j2objcTranslationClosure` dependency on the associated sources
+Jar for the module.
+
+### testCompile
```
brunobowden: Given the similarity, I believe the "test" and "non-test" configurations should be combined for the explanations. For example:

    ### compile & testCompile
    If `autoConfigureDeps` is false, has no effect related to J2ObjC.  When true,
    * for a project, creates a `j2objcLinkage` (or `j2objcTestLinkage`) dependency on the project.
    * for an external module already included in the J2ObjC distribution, has no additional effect (this will work automatically).
    * for any other external module, creates a `j2objcTranslationClosure` dependency on the associated sources Jar for the module. For testCompile, this fails immediately as we cannot build tests using the --build-closure due to the risk of duplicate symbols.

dependencies.md
```diff
@@ -0,0 +1,255 @@
...
+Use this only with a Jar of Java source files.  These files will be provided to the j2objc
+translator with `--build-closure`; where your main source set relies on classes defined
+in this source Jar, j2objc will additionally translate that class (and all of its transitive
+dependencies).  The plugin will include the result in your project's own library.
+
+This is only safe to use when you are building a single J2ObjC Java Gradle project.
+If you have multiple projects, when the libraries for those projects are linked together,
+you will get duplicate symbol errors.
+
+### j2objcTranslation
```
brunobowden: ### j2objcTranslation & j2objcTestTranslation

dependencies.md
```diff
@@ -0,0 +1,255 @@
...
+source set.  They will be both compiled into your project's compiled Jar file, and also be
+fully translated by j2objc, and the Objective C will be included in your project's static
+library.
+
+### j2objcTestTranslation
+Use this only with a Jar of Java JUnit test source files.  These files will be added to the test Java
+source set, and executed in Java.  These will also be fully translated by j2objc and then executed
+in their Objective C form.  They will not be included in your project's static library.
+
+### j2objcLinkage
```
brunobowden: ### j2objcLinkage & j2objcTestLinkage

brunobowden:
LGTM

brunobowden:
It appears that Github doesn't recognize the [ci skip] command that Travis does recognize.

dependencies.md
advayDev1: unfortunately not

dependencies.md
advayDev1: done

dependencies.md
advayDev1: done

dependencies.md
advayDev1: done, but the list is accurate to a fix in a separate PR

dependencies.md
advayDev1: done

dependencies.md
advayDev1: done

dependencies.md
advayDev1: done

dependencies.md
advayDev1: done

dependencies.md
advayDev1: done

dependencies.md
advayDev1: done

PR #491, opened 2015-10-06
Automatic v0.4.3 => v0.5.0 Podfile upgrade

brunobowden:
This allows automatic update of the Podfile without requiring the developer to recreate a new Podfile. The code is somewhat hacky but is safe to use.

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy
@@ -426,6 +426,15 @@ class XcodeTask extends DefaultTask {
                 withinXcodeTarget = true
 
             } else if (withinXcodeTarget) {
+
+                // For upgrade from v0.4.3 to v0.5.0, basically for Xcode 7
+                // TODO: remove code for plugin beta release as it's not necessary after upgrading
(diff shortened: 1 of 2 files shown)


src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy
```diff
@@ -426,6 +426,15 @@ class XcodeTask extends DefaultTask {
                 withinXcodeTarget = true
 
             } else if (withinXcodeTarget) {
+
+                // For upgrade from v0.4.3 to v0.5.0, basically for Xcode 7
```
advayDev1: are you sure you want to do this?  your comment on my PR was for Xcode 7 was that it was fine to break things for earlier users as it is easier for us to maintain.  I think the same applies here.

For example, if you find one of the old lines, why not just delete the file and start over?

brunobowden: Conceivably they could have other pods in the file that they need for other
purposes.

In this case, I think the automatic fix was easier than describing to
people how to fix it themselves if they're already added other pods. This
seem simple enough to me that it was appropriate as a temporary measure.

advayDev1:
hmm ok lgtm

PR #493, opened 2015-10-06
Need to link dependencies.md

advayDev1:


advayDev1:
ptal @brunobowden 
there's no failure in the CI - i just cancelled it

brunobowden:
I didn't understand the purpose of doing this. Do you intend to show parts of this as error output as explanation on what a developer should do? Why not just link to the GitHub repo instead? This commit definitely could do with more description to explain it's purpose.

advayDev1:
All of our documentation is linked into our plugin.  README, FAQ, LICENSE, etc.
We chose to move text from FAQ to dependencies.md, and FAQ references by name dependencies.md.
We'd be losing documentation if we didn't also link dependencies.md like all the other

brunobowden:
LGTM

So how do people access this? Presumably they'd only see it if they were routing around inside the jar?

advayDev1:
Yes.  For IntelliJ at least, all of the things on your classpath are browseable in the external libraries tree and so easy to read.

PR #495, opened 2015-10-07
Use Java 7 on Travis OS X; get Guava working again

advayDev1:

PR #494, opened 2015-10-07
multiProject1 Xcode workspace + systemTests scripts refactor

brunobowden:
multiProject1 Xcode workspace and project …
- Default configuration with no adjustments
- iOS Application: “Single View Application”
- No watchOS or OS X targets
- .gitignore extended to cover Xcode (copied from github/ignore repo)

Simpler developer running of tests …
- Run all scripts from root directory, without needing push systemTests
- Separate out runTest method as run-test.sh script
- Push J2OBJC_VERSION in to script, can be set by user but still has default

PR #497, opened 2015-10-07
Fix bad merge

advayDev1:
#494 changed the working directory; I didn't merge that in #495


brunobowden:
LGTM This is where it would be nice to have a submit queue.

advayDev1:
yeah having to come back to this half an hour later and remembering to merge it is annoying

PR #501, opened 2015-10-08
Match release optimization behavior of Xcode 7

### LINKED ISSUE #500: Optimize debug vs. release clang builds
We use the same optimization setting for debug and release builds.  One of those is probably wrong :)

advayDev1:
Fixes #500

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/NativeCompilation.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/NativeCompilation.groovy
@@ -288,9 +288,15 @@ class NativeCompilation {
                     linker.args j2objcConfig.extraLinkerArgs
 
                     if (buildType == buildTypes.debug) {
+                        // Full debugging information.
                         objcCompiler.args '-g'
                         objcCompiler.args '-DDEBUG=1'
                     } else {  // release
+                        // Per https://raw.githubusercontent.com/llvm-mirror/clang/8eb384a97cfdc244a5ab81026677bcbaf8cf2ecf/docs/CommandGuide/clang.rst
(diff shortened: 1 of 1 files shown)


src/main/groovy/com/github/j2objccontrib/j2objcgradle/NativeCompilation.groovy
```diff
@@ -290,6 +290,8 @@ class NativeCompilation {
                     if (buildType == buildTypes.debug) {
                         objcCompiler.args '-g'
                         objcCompiler.args '-DDEBUG=1'
+                    } else {  // release
+                        objcCompiler.args '-Os'
```
brunobowden: Can you add a comment referencing where this argument is described and also where it's explained why this is standard for release?

brunobowden:
LGTM - please see comment about adding comments ;-)

src/main/groovy/com/github/j2objccontrib/j2objcgradle/NativeCompilation.groovy
advayDev1: done.  referenced clang source re: what it does.
however, there is no defined standard; i merely referenced we are matching Xcode 7's behavior.

brunobowden:
So you verified the behaviour manually rather than seeing it documented somewhere?

advayDev1:
What 'it' does is per llvm's source.  The fact that we should indeed do 'it' is verified by running Xcode.

advayDev1:
Well found some documentation after the fact that validates my understanding:
https://developer.apple.com/library/ios/qa/qa1795/_index.html#//apple_ref/doc/uid/DTS40014195-CH1-COMPILER

will add as well

PR #511, opened 2015-10-14
Contribution certifications

advayDev1:
Note: This is not legal advice.  Contributors, users, etc. must ensure their own level of comfort with
contributions certified as described here, and should seek their own legal counsel as needed.

See the License for disclaimers and limitations of liability, including but not limited to
Section 7:

```
7. Disclaimer of Warranty. Unless required by applicable law or
      agreed to in writing, Licensor provides the Work (and each
      Contributor provides its Contributions) on an "AS IS" BASIS,
      WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
      implied, including, without limitation, any warranties or conditions
      of TITLE, NON-INFRINGEMENT, MERCHANTABILITY, or FITNESS FOR A
      PARTICULAR PURPOSE. You are solely responsible for determining the
      appropriateness of using or redistributing the Work and assume any
      risks associated with Your exercise of permissions under this License.
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,6 +15,40 @@ the [Apache 2.0 License](LICENSE).
 [preparing-your-pull-request-for-submission](#preparing-your-pull-request-for-submission).
 
 
+### Conditions for accepting pull requests
+Please confirm that you can certify the following, 
+then add the certification at the end of every commit message (with two newlines between
+the main content of your commit message and the notice), entering your own information
+and removing the &lt;angled brackets&gt;. If you can't certify the following, please
+do not submit a pull request.
+
+```
+I, <full name> (<email@example.com>, https://github.com/<github username>), certify that
+a) this Contribution is my original work, and
+b) I have the right to submit this Contribution under the Apache License,
+   Version 2.0 (the "License") available at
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -15,6 +15,40 @@ the [Apache 2.0 License](LICENSE).
 [preparing-your-pull-request-for-submission](#preparing-your-pull-request-for-submission).
 
 
+### Conditions for accepting pull requests
+Please confirm that you can certify the following,
```
brunobowden: Any reason for the strange line wrap?

brunobowden:
Can we make it apply to all future contributions, just to make it easier? It would then act closer to a normal CLA.

brunobowden:
LGTM either way

advayDev1:
I don't know how we'd make it apply to future contributions; as different git repos/branches/forks could have different sequences of commits.  I think we should keep it simpler, and if/when we figure out Apache-style CLAs just use those.

CONTRIBUTING.md
advayDev1: i'm just going to leave it alone to avoid repushing :)

PR #512, opened 2015-10-15
vendored_libraries for podspec

### LINKED ISSUE #460: PodSpec should use vendored_libraries
Compare:
https://guides.cocoapods.org/syntax/podspec.html#vendored_libraries
with what we use right now:
https://guides.cocoapods.org/syntax/podspec.html#libraries

Not sure if vendored_libraries should be used in addition to or instead of libraries, but our libraries do come with the pod, not with the system.

brunobowden:
- spec.libraries relpaced with spec.{ios|osx|watchos}.vendored_libraries
- j2objc lib LIBRARY_SEARCH_PATH xcconfig removed

Fixes #460


brunobowden:
cc @maxbritto

I'm seeing a failure in systemTests/multiProject1. Creating a PR to see if the automatic build sees the same issue.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/PodspecTask.groovy
```diff
@@ -159,7 +159,8 @@ class PodspecTask extends DefaultTask {
                "  spec.resources = '$resourceDir/**/*'\n" +
                "  spec.requires_arc = true\n" +
                "  spec.libraries = " +  // continuation of same line
-               "'ObjC', 'guava', 'javax_inject', 'jre_emul', 'jsr305', 'z', 'icucore', '$libName'\n" +
+               "'ObjC', 'guava', 'javax_inject', 'jre_emul', 'jsr305', 'z', 'icucore'\n" +
+               "  spec.vendored_libraries = '$libName'\n" +
```
advayDev1: did you try what max suggested in the bug?
use the filename as 'lib${libName}.a'

advayDev1: well he used the full path, but you have a library search path

brunobowden:
It appears to be a transient failure, so I'm expecting the build to succeed.

advayDev1:
hmm what was the failure?  Travis isn't building the xc project anyway

brunobowden:
I believe it was due to using an old plugin build. Haven't seen it since.

The vendored_libraries has been separated out by OS. Example:
```
spec.libraries = 'ObjC', 'guava', 'javax_inject', 'jre_emul', 'jsr305', 'z', 'icucore'",
spec.ios.vendored_libraries = 'j2objcOutputs/lib/iosDebug/lib$libName'",
spec.osx.vendored_libraries = 'j2objcOutputs/lib/x86_64Debug/lib$libName'",
spec.watchos.vendored_libraries = 'j2objcOutputs/lib/iosDebug/lib$libName'",
```

brunobowden:
Still need to test this but it should work.

advayDev1:
Are you sure you don't need a `.a` suffix?  It's odd if this is working.

brunobowden:
I haven't tested it yet. Fixed by adding the ".a" suffix.

brunobowden:
I will of course test it before merging.

advayDev1:
Cool, logically LGTM.  Merge whenever it passes.

PR #524, opened 2015-10-21
j2objcXcode now dependency of j2objcBuild

brunobowden:
- This will update the Podfile when the j2objc project dependencies change


brunobowden:
PTAL

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy
```diff
@@ -277,7 +277,7 @@ class J2objcPlugin implements Plugin<Project> {
             }
             lateDependsOn(project, 'build', 'j2objcBuild')
 
-            // TODO: Where shall we fit this task in the plugin lifecycle?
+            // TODO: make this a dependency of j2objcBuild
```
advayDev1: Do you want to just go ahead and do this?

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy
```diff
@@ -114,7 +125,23 @@ class XcodeTask extends DefaultTask {
     void xcodeConfig() {
         Utils.requireMacOSX('j2objcXcode task')
 
-        verifyXcodeArgs()
+        // TODO: figure out how to automatically run this on top-level project
```
advayDev1: what does this comment mean?

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy
```diff
@@ -114,7 +125,23 @@ class XcodeTask extends DefaultTask {
...
+        // TODO: but must be disabled on all dependencies
+        if (!isTaskActive()) {
```
advayDev1: I agree with the goal where j2objcBuild depends on j2objcXcode.  But in that version, !isTaskActive is the way to bail out on this, and should not throw an error

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy
brunobowden: Ideally you'd like this error to apply on the root project, the same that builds the android app. You don't know if any project is the root or a dependency instead. I haven't figured out if there's an appropriate way to figure out when to show this error.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy
brunobowden: I was a little apprehensive about doing this... but going ahead and making the change. Also update the docs too.

advayDev1:
Please add a comment to CHANGELOG.md also; LGTM when passes.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy
```diff
@@ -53,6 +53,18 @@ class XcodeTask extends DefaultTask {
     @Input @Optional
     String getXcodeProjectDir() { return J2objcConfig.from(project).xcodeProjectDir }
 
+    boolean isTaskActive() { return getXcodeProjectDir() != null }
+
+    @Input
+    // List of all dependencies
+    List<PodspecDetails> getPodspecDependencies() {
```
advayDev1: class PodspecDetails must be modified as follows to be an input:
```
@EqualsAndHashCode
static class PodspecDetails implements Serializable {
...
private static final long serialVersionUID = 1L;
...
}
```

brunobowden:
@advayDev1 - it also required implementing `writeObject()` and calling `defaultWriteObject()`. I added deserialization as well, so that I could test it end it end. Curiously, the `@EqualsAndHashCode` wasn't needed, so I wanted your opinion on that.

advayDev1:
1.  Not throwing without EqualsAndHashCode makes sense.  However, I think you use it anyway: while serialization will work, you need it to make sure up-to-date checks are correct and even in Groovy, equality is referential unless overridden.

2.  I'm not understanding why writeObject/defaultWriteObject were needed for you.  I don't have them.  Instead of doing that, consider using @AutoExternalize: http://docs.groovy-lang.org/2.4.5/html/gapi/groovy/transform/AutoExternalize.html

advayDev1:
Aha: you don't have a default no-arg constructor for PodspecDetails.  Ok this makes sense now

brunobowden:
I expect that Gradle is comparing Serialized binary to Serialized binary, that way that can get consistent behaviour over time and across daemon runs - they can't keep the object around in memory. The writeDefaultObject is a necessary part of serialization.

advayDev1:
I'd still suggest using @EqualsAndHashCode - we don't know what the implementation detail of Gradle is (I looked briefly, couldn't find it).  Understood about writeDefaultObject, the point of @AutoExternalizable is to do that for you.  (I'm still trying to figure out why I don't need AutoExternalizable nor implementations of Serializable - but that is an aside).

brunobowden:
I've added back @EqualsAndHashCode. Since the serialization is working (and is tested), I'm fine with doing it the Java way with writeDefaultObject().

One concern I have is that getPodspecDependencies() is called 2-3 times during the gradle run. I'm surprised that this happens. Dependending on how large the dependency trees become, this may need caching.

I'll merge this once the build is green.

advayDev1:
sg - LGTM

PR #527, opened 2015-10-22
Handle autogenerated classes correctly.

### LINKED ISSUE #517: Support for annotation processors' generated code (ex. dagger2)
There are many assorted bugs regarding individual error messages when using annotation processors - but they haven't gone anywhere though.
This bug is instead for end-to-end support of dagger2, as opposed of hacking it into the plugin through various individual j2objc config properties.

While there may be many annotation processors and mechanisms for their integration into Gradle, success here will be measured by the ability to successfully use one of the most common: Dagger2 (https://github.com/google/dagger).

@andrewfunston - If you are willing, please document the process you are using to get dagger2 working right now.  I think it would be helpful to consider contributing a generalized version of that as a PR to the plugin itself instead of just the example project (see [here](https://github.com/j2objc-contrib/j2objc-gradle/blob/94e74ff3aee38435f0aaa80a864be2c28e5631de/CONTRIBUTING.md#conditions-for-accepting-pull-requests)).

advayDev1:
Handle autogenerated classes correctly.
- Make autogenerated files work even without --build-closure
- Handle changes to autogenerated files that are not solely caused by changes to hand-written files (ex. configuration changes).
- Default to build/classes/main for annotation-processor generated source.
- Both CycleFinder and Translate now treat autogen code identically.
- Remove duplicate functionality between 'additional' and 'generated' files.

Fixes #517

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
@@ -127,18 +127,15 @@ class J2objcConfig {
 
     /**
      * Generated source files directories, e.g. from dagger annotations.
-     * <p/>
-     * The plugin will ignore changes in this directory so they must
-     * be limited to files generated solely from files within your
-     * main and/or test sourceSets.
      */
-    List<String> generatedSourceDirs = new ArrayList<>()
+    // Default location for generated source files using annotation processor compilation,
+    // per sourceSets.main.output.classesDir.
+    // However, we cannot actually access sourceSets.main.output.classesDir here, because
+    // the Java plugin convention may not be applied at this time.
+    // TODO: Add a test counterpart for this.
+    List<String> generatedSourceDirs = ['build/classes/main']
     /**
      * Add generated source files directories, e.g. from dagger annotations.
-     * <p/>
-     * The plugin will ignore changes in this directory so they must
-     * be limited to files generated solely from files within your
-     * main and/or test sourceSets.
      *
      * @param generatedSourceDirs adds generated source directories for j2objc translate
      */
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy
@@ -149,10 +149,6 @@ class J2objcPlugin implements Plugin<Project> {
                     dependsOn: 'j2objcPreBuild') {
                 group 'build'
                 description "Translates all the java source files in to Objective-C using 'j2objc'"
-                additionalMainSrcFiles = files(
-                        fileTree(dir: "build/source/apt",
-                                include: "**/*.java")
-                )
                 // Output directories of 'j2objcTranslate', input for all other tasks
                 srcGenMainDir = j2objcSrcGenMainDir
                 srcGenTestDir = j2objcSrcGenTestDir
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
@@ -28,7 +28,9 @@ import org.gradle.api.Nullable
 import org.gradle.api.Project
 import org.gradle.api.file.CopySpec
 import org.gradle.api.file.FileCollection
+import org.gradle.api.file.FileTree
 import org.gradle.api.file.SourceDirectorySet
+import org.gradle.api.internal.file.UnionFileTree
 import org.gradle.api.plugins.JavaPluginConvention
 import org.gradle.api.tasks.SourceSet
 import org.gradle.api.tasks.WorkResult
@@ -408,14 +410,10 @@ class Utils {
     }
 
     // Add list of java path to a FileCollection as a FileTree
-    static FileCollection javaTrees(Project proj, List<String> treePaths) {
-        FileCollection files = proj.files()
-        treePaths.each { String treePath ->
-            log.debug "javaTree: $treePath"
-            files = files.plus(
-                    proj.files(proj.fileTree(dir: treePath, includes: ["**/*.java"])))
-        }
-        return files
+    static FileTree javaTrees(Project proj, List<String> treePaths) {
+        List<? extends FileTree> trees =
+            treePaths.collect({ String treePath -> proj.fileTree(dir: treePath, includes: ["**/*.java"]) })
+        return new UnionFileTree("javaTrees_j2objc", trees)
     }
 
     static List<String> j2objcLibs(String j2objcHome,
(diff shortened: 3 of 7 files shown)


src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcPlugin.groovy
```diff
@@ -149,8 +149,9 @@ class J2objcPlugin implements Plugin<Project> {
                     dependsOn: 'j2objcPreBuild') {
                 group 'build'
                 description "Translates all the java source files in to Objective-C using 'j2objc'"
+                // Default location for generated source files using annotation processor compilation.
```
brunobowden: Can you add a reference to where this is defined?

brunobowden:
LGTM

advayDev1:
Waiting for @andrewfunston's ack on #517 .

This is a guess, we'll see if it works.  If it does indeed work, I'll try to look through Gradle's source to see if we can provide an authoritative reference.

advayDev1:
@brunobowden ptal; completely redone btw.

advayDev1:
Per @andrewfunston at https://github.com/j2objc-contrib/j2objc-gradle/issues/517#issuecomment-150364608 - this is functionally good to go

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
```diff
@@ -127,18 +127,11 @@ class J2objcConfig {
...
     /**
      * Generated source files directories, e.g. from dagger annotations.
-     * <p/>
-     * The plugin will ignore changes in this directory so they must
-     * be limited to files generated solely from files within your
-     * main and/or test sourceSets.
      */
-    List<String> generatedSourceDirs = new ArrayList<>()
+    // Default location for generated source files using annotation processor compilation.
+    List<String> generatedSourceDirs = ['build/classes/main']
```
brunobowden: Is there any convention around writing generated source to `build/classes/test`?

advayDev1: I haven't found anything like that.  I don't think we could distinct support annotation processors in the test source set (that aren't in the main source set) at this moment without significant changes (remember we have a limitation that build-closure is not possible to apply for the test source).

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
```diff
@@ -15,7 +15,6 @@
  */
 
 package com.github.j2objccontrib.j2objcgradle.tasks
-
```
brunobowden: Those blank lines are for keeps...

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
andrewfunston: I personally used injection in my test set on my real project; I'm not sure if it's standard practice or not. I add component that inject test implementations (in memory dbs, etc); that was failure on my part to include that in the sample project, I was attempting to boil everything down pretty thin. Is there any actionable items for this, or is this an accepted shortcoming of the current j2objc plugin?

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
```diff
@@ -408,14 +408,10 @@ class Utils {
...
-        treePaths.each { String treePath ->
-            log.debug "javaTree: $treePath"
-            files = files.plus(
-                    proj.files(proj.fileTree(dir: treePath, includes: ["**/*.java"])))
-        }
-        return files
+    static FileTree javaTrees(Project proj, List<String> treePaths) {
+        List<? extends FileTree> trees =
+            treePaths.collect({ String treePath -> proj.fileTree(dir: treePath, includes: ["**/*.java"]) })
+        return new UnionFileTree("javaTrees_${treePaths.join(';')}", trees)
```
brunobowden: Why have such a long displayName? If I was debugging an issue, I'd rather look at the contents. At the moment, it's a little confusing as the `${treePaths.join(';')}` implies that you're creating a string for a command line with a `;` path separator (which is Windows specific). I'd be quite happy with `j2objc_javaTrees` as the displayName.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/CycleFinderTaskTest.groovy
```diff
@@ -94,7 +94,7 @@ class CycleFinderTaskTest {
                 null,
                 [
                         'INVALID-NEEDS-WINDOWS-SUBSTITUTION',
-                        '-sourcepath', '/PROJECT_DIR/src/main/java;/PROJECT_DIR/src/test/java',
+                        '-sourcepath', '/PROJECT_DIR/src/main/java;/PROJECT_DIR/src/test/java;/PROJECT_DIR/build/classes/main',
```
brunobowden: This is separated by `;` instead of `:` - likely an error.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
advayDev1: How about we do the same as we did here?
i.e. you put out some GitHub/Gist repos of an issue (if there actually is any, it might just work?), and we'll try to fix.  But let's please do it as a separate bug and PR (please file the bug Andy).

Adding build/classes/test here is not a good idea regardless, as it would add those test files to your production library.

It is the annotation processors themselves that we can't have be distinct for the test set.  I think I can get it to work as long as you use for example Dagger in both main and test, but you happen to inject different classes (along with their autogen counterparts) in your test code.  If you were to use SomeOtherInjectionFramework in test (only), I don't think we can make that work in the near future.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/CycleFinderTaskTest.groovy
advayDev1: huh - and the test passes...

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
advayDev1: ok

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
advayDev1: haha

brunobowden:
LGTM after addressing comments

advayDev1:
Unit-tests for all platforms have passed; non-aesthetic changes should have only affected them.  Will get the 30min result on main branch anyway.

merging.

PR #530, opened 2015-10-23
minVersionOsx = 10.6 (new default)

brunobowden:
- minVersionOsx now defaults to 10.6 (oldest version that supports ARC)
- exception thrown for unsupported watchOS (manually tested)

allPlatforms
- clean up shared/build.gradle
- correct symlink for local.properties

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
@@ -628,8 +628,9 @@ class J2objcConfig {
      * <p/>
      * See https://developer.apple.com/library/ios/documentation/DeveloperTools/Conceptual/cross_development/Configuring/configuring.html#//apple_ref/doc/uid/10000163i-CH1-SW2
      */
-    // Matches the oldest version supported in Xcode 7
-    String minVersionOsx = '10.4'
+    // Oldest OS X version that supports automatic reference counting (2009 onwards)
+    // Prevents Xcode error: "-fobjc-arc is not supported on versions of OS X prior to 10.6"
+    String minVersionOsx = '10.6'
 
     /**
      * The minimum Watch OS version to build against.  You cannot use APIs that are not supported
@@ -794,6 +795,13 @@ class J2objcConfig {
         assert destLibDir != null
         assert destSrcMainDir != null
         assert destSrcTestDir != null
+
+        // TODO: watchOS build support
+        if (xcodeTargetsWatchos.size() > 0) {
+            throw new InvalidUserDataException(
+                    "WatchOS isn't yet supported, please unset xcodeTargetsWatchos for now." +
+                    "Follow this issue for updates: https://github.com/j2objc-contrib/j2objc-gradle/issues/525")
+        }
     }
 
     protected void configureNativeCompilation() {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,9 +23,9 @@ New functionality:
 
 Breaking changes/functionality:
 * No longer supports Xcode 6 and lower or j2objc 0.9.8.2 and lower #483
-* Minimum versions of platforms have been reduced to iOS 6.0, OS X 10.4, and WatchOS 1.0 #512
+* Minimum versions of platforms have been configured as iOS 6.0, OS X 10.6, and WatchOS 1.0 #512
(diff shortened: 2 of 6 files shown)


brunobowden:
@advayDev1 - I think we should warn people explicitly that watchOS isn't supported, rather then leave them with some more confusing errors.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy
```diff
@@ -161,6 +161,13 @@ class XcodeTask extends DefaultTask {
 //        "https://github.com/j2objc-contrib/j2objc-gradle/blob/master/FAQ.md#what-is-the-recommended-folder-structure-for-my-app\n"
 //        throw new InvalidUserDataException(message)
 
+        // TODO: watchOS build support
+        if (getXcodeTargetsWatchos().size() > 0) {
```
advayDev1: move this to finalConfigure.

no reason to have people suffer through a long translate/compile then realize they can't use this system for what they want

advayDev1:
comment + LGTM

brunobowden:
@advayDev1 - please take another look. I've add in some other cleanup as well but mostly minor stuff.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/NativeCompilation.groovy
```diff
@@ -35,6 +35,8 @@ import org.gradle.platform.base.Platform
  */
 class NativeCompilation {
 
+    // TODO: @Input dependency on config.minVersionIos and other j2objcConfig fields
```
advayDev1: this is not possible, as these are not tasks.  while conceptually we want to do something like this, or file a bug on gradle for it, this TODO is not it.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
```diff
@@ -785,6 +786,13 @@ class J2objcConfig {
             Utils.throwJ2objcConfigFailure(project,
                     "Found $actualVersionString at $j2objcHome, J2ObjC v$j2objcVersion required.")
         }
+
+        // TODO: watchOS build support
```
advayDev1: validateConfiguration is the right place for this; it is called by finalConfigure

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
```diff
@@ -628,8 +628,9 @@ class J2objcConfig {
      * <p/>
      * See https://developer.apple.com/library/ios/documentation/DeveloperTools/Conceptual/cross_development/Configuring/configuring.html#//apple_ref/doc/uid/10000163i-CH1-SW2
      */
-    // Matches the oldest version supported in Xcode 7
-    String minVersionOsx = '10.4'
+    // Oldest OS X version that supports automatic reference counting (2009 onwards)
+    // Prevents Xcode error: "-fobjc-arc is not supported on versions of OS X prior to 10.6"
+    String minVersionOsx = '10.6'
```
advayDev1: Please update the CHANGELOG with this version too

advayDev1:
further comments + LGtM

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
brunobowden: moved to validateConfiguration

src/main/groovy/com/github/j2objccontrib/j2objcgradle/NativeCompilation.groovy
advayDev1: if you want to keep a comment, I'd put instead the following at `defineTarget` and remove this one:
```
// NOTE: Gradle native compilation fails to consider the clang args added here in its up-to-date-checks.
// Changing this configuration may require the user to do a clean build
// See issue #XX.
```

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
brunobowden: Updated CHANGELOG

Issue #528, opened 2015-10-23
Alpha release acceptance testing


advayDev1:
We hope to have an alpha (0.5) release today/tomorrow, and we'd like some help testing it.
We're hoping to get at minimum successful builds on:
- @brunobowden - multi-project + full configuration via cocoapods
- @advayDev1 - multi-project + native libraries + other power features
- @andrewfunston - project using Dagger/dependency injection/annotation processors

When this bug is updated with a particular commit SHA, please try building (and running on an iOS device if possible) your project and update this thread with the results.

Thanks!

brunobowden:
I also use cycleFinder as well.

advayDev1:
me too; we should turn that on by default someday :)

brunobowden:
Let's go with SHA 241b9df. That's PR #530 "minVersionOsx = 10.6 (new default)".

It's already working well for me.

@advayDev1 - did you run the overnight build like you talked about?

advayDev1:
no, I was waiting for the SHA.
let me kick things off now, since i wont be on my machine again until ~4pm, perhaps it will finish quicker

advayDev1:
Alright well in this span https://github.com/madvay/j2objc-gradle/compare/v0.4.9.4.0-alpha-fork...madvay:v0.4.9.9-alpha-fork we've broken something at least for me.  Investigating now, but I will have to be away in a bit.

advayDev1:
So after clean building some and incrementally building others (for speed), everything is working except:

1.  pretty obscure issue re: translate mappings (actually an undocumented breaking change in the CHANGELOG - which caused all my other issues; filing a PR).
2.  @brunobowden you need to decide whether this is an issue or not:  With project B depending on A, when you run './gradlew build', B's j2objcBuild will succeed before A's j2objcPodSpec or j2objcXcode (but B's j2objcXcode will occur after A's j2objcXcode, just all at the very end of the build).  This occurs because while j2objcAssemble depends on j2objcXcode (and therefore j2objcPodSpec),  j2objcBuild does not depend on j2objcAssemble (which is correct) and instead j2objcBuildDebug and j2objcBuildRelease rely on j2objcAssembleDebug and j2objcAssembleRelease.  There is no easy solution here except adding yet another layer of indirect tasks.  You can't make j2objcAssembleDebug|Release rely on j2objcXcode.

brunobowden:
[...]

PR #481, opened 2015-10-04
Wildcard prefix matching

brunobowden:
Fixes #430

Example:
```
prefixes.properties: com.example.dir.*=Wildcard
Test File:           com/example/dir/subdir/Test.java
Test Name:           WildcardTest
```

PR #531, opened 2015-10-23
Undocumented breaking change re #528

advayDev1:
This was the only alpha acceptance bug for me.


brunobowden:
LGTM

Also add this line in new functionality:

    Wilcard prefix matching #481

PR #532, opened 2015-10-23
j2objcAssemble{Debug|Release} depends on j2objcXcode

brunobowden:
- j2objcAssemble nolonger depends directly on j2objcXcode


brunobowden:
@advayDev1 - this isn't strictly necessary but it feels like a better place for j2objcXcode. Please share your feedback on this.

advayDev1:
If pod install does not rely on the libraries existing or anything like that, LGTM.  While I know the XcodeTask does not have those generated files as input files, I do not know the internals of pod install.

brunobowden:
I tested the `pod install` for two podspecs with dangling references and it worked just fine. I've add a comment to the task and will merge now.

PR #533, opened 2015-10-23
Gradle cache on CI build

brunobowden:
- Caches `$HOME/.gradle/caches/` and `$HOME/.gradle/wrapper/`

- Remove lock file to prevent upload of essentially unmodified cache

    $HOME/.gradle/caches/modules-2/modules-2.lock

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,7 @@ branches:
 sudo: false
 
 # Gradle Cache
-# Avoids uploading the cache after every build
+# Avoid uploading the cache after every build
 # http://docs.travis-ci.com/user/languages/android/#Default-Test-Command-for-Gradle
 before_cache:
   - rm -f $HOME/.gradle/caches/modules-2/modules-2.lock
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,6 +44,18 @@ branches:
 # We can be run in a container for improved performance.
 sudo: false
 
+# Gradle Cache
+# Avoids uploading the cache after every build
+# http://docs.travis-ci.com/user/languages/android/#Default-Test-Command-for-Gradle
+before_cache:
+  - rm -f $HOME/.gradle/caches/modules-2/modules-2.lock
+cache:
+  directories:
+    # TODO: match systemTest directories as well?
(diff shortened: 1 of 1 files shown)


advayDev1:
I don't think we should do this:
- Practically, you are only saving between 2 and 20 seconds out of a long tail of 36 minutes - less than one percent.

- We're building a gradle plugin, so the behavior of gradle is itself part of what we are testing.
In the example you pulled from (an Android app), you're not treating interactions with Gradle as something you need to test.  Each test should be completely hermetic, and the nuances of what happens between gradle runs shouldn't affect our CI builds.

- It is in fact part of the test that the libraries and what not we reference are available - if we for example we are using a Gradle api or dependency that is deleted or deprecated, or our system tests use a Jar that is similarly unavailable on jCenter, failing fast is the right thing to do.

brunobowden:
Fair points. Arguably we could use the cache for other purposes, e.g. localJ2objcDist but the value is marginal. Closing this.

advayDev1:
if we could "cache": localJ2objcDist, cocoa pods, and java 7 on OSX, we'd save about 4 minutes and it'd be actually worth it I think.  caching localJ2objcDist is a win of 1.5minutes (best case with 100% local disk hit rate).  caching an installed component like cocoa pods, brew, java 7 seems like a lot of work

brunobowden:
Looking at the logs, it appears that the j2objc release was downloaded within 3 seconds. Where are you getting 1.5 minutes from?

```
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   597    0   597    0     0   9921      0 --:--:-- --:--:-- --:--:-- 10854
100  127M  100  127M    0     0  36.4M      0  0:00:03  0:00:03 --:--:-- 40.7M
```

advayDev1:
My bad 90 seconds is Java 7.  40 seconds is download + unzip (much slower) of j2objc

brunobowden:
How are you measuring that? I couldn't see that from the logs.

advayDev1:
screen shot 2015-10-23 at 6 02 38 pm

PR #535, opened 2015-10-23
j2objc(Podspec|Xcode) disabled for translateOnlyMode

brunobowden:
- Necessary when running on Windows & Linux

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
@@ -833,6 +833,8 @@ class J2objcConfig {
             Utils.requireMacOSX('Native Compilation of translated code task')
         }
 
+        project.logger.info("J2objcPlugin: translateOnlyMode will disable most j2objc tasks")
+
         project.tasks.all { Task task ->
             String name = task.name
             // For convenience, disable all debug and/or release tasks if the user desires.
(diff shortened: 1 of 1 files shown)


brunobowden:
@advayDev1 - I haven't tried this on Windows or Linux but since we moved j2objcXcode to part of the build, it must be possible to disable it on that platform.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
```diff
@@ -853,7 +853,8 @@ class J2objcConfig {
                 // First pattern matches all native-compilation tasks.
                 // Second pattern matches plugin-specific tasks beyond translation.
                 if ((name =~ /^.*((J|j)2objc(Executable|StaticLibrary|SharedLibrary|Objc))$/).matches() ||
-                    (name =~ /^j2objc(Assemble|PackLibraries|Test)(Debug|Release)$/).matches()) {
+                    (name =~ /^j2objc(Assemble|PackLibraries|Test)(Debug|Release)$/).matches() ||
+                    (name =~ /^j2objc(Podspec|Xcode)$/).matches()) {
                     task.enabled = false
```
advayDev1: Can we add a project.logger.info("J2objcPlugin: Disabling $name in translateOnlyMode");

advayDev1:
lgtm + comment

brunobowden:
That'll spew out a lot of noise in the build. The user will already see `shared:j2objcXcode SKIPPED` anyway, so I don't feel this adds value. I've add a single info log that translateOnlyMode is enabled but that will only be shown once.

PR #536, opened 2015-10-23
Android CI build

brunobowden:
- Add allPlatforms to run-all.sh
- ANDROID_HOME env variable, Travis default: `/usr/local/android-sdk`
- Gradle allPlatforms/build.gradle => disable Android builds on OS X
- Fix run-all.sh so that without any parameter it will default to all

Travis config:
- Linux / oraclejdk7 => no tests (same as before)
- Linux / openjdk7 => TEST_SET=1 (also J2OBJC_TRANSLATE_ONLY)
- OS X => ANDROID_BUILD_DISABLED (using settings.gradle)

Android environment:
- android-21
- build-tools-21.1.2

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,11 +6,17 @@ matrix:
     - os: linux
       language: groovy
       jdk: oraclejdk7
-      env: USING_OS=linux
+      env:
+        # Trick to run no tests
+        - TEST_SET=NONE
+        - USING_OS=linux
     - os: linux
-      language: groovy
+      language: android
       jdk: openjdk7
-      env: USING_OS=linux
+      env:
+        - TEST_SET=1
+        - USING_OS=linux
+        - J2OBJC_TRANSLATE_ONLY=true
     # 'objective-c' forces a build on OS X.
     # Because we override the install and script commands
     # below, this works fine, even though we are actually
@@ -22,18 +28,25 @@ matrix:
       osx_image: xcode7
       language: objective-c
       env:
-       - TEST_SET=1
-       - USING_OS=osx
-       - USING_XCODE=7
+        - TEST_SET=1
+        - USING_OS=osx
+        - USING_XCODE=7
+        - ANDROID_BUILD_DISABLED=true
     # Split up the system tests into 2 sets to avoid hitting the
     # max build time.
     - os: osx
       osx_image: xcode7
       language: objective-c
       env:
-       - TEST_SET=2
-       - USING_OS=osx
-       - USING_XCODE=7
+        - TEST_SET=2
+        - USING_OS=osx
+        - USING_XCODE=7
+        - ANDROID_BUILD_DISABLED=true
+
+android:
+  components:
+    - build-tools-21.1.2
+    - android-21
 
 branches:
   only:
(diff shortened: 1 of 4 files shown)


brunobowden:
Android build on Linux is now confirmed as working. Here's part of the log:

    :app:assembleDebugUnitTest
    :app:testDebugUnitTest
    :app:preReleaseUnitTestBuild UP-TO-DATE
    :app:prepareReleaseUnitTestDependencies
    :app:processReleaseUnitTestJavaRes UP-TO-DATE
    :app:compileReleaseUnitTestJavaWithJavac
    :app:compileReleaseUnitTestSources
    :app:assembleReleaseUnitTest
    :app:testReleaseUnitTest
    :app:test
    :app:check
    :app:build

advayDev1:
Question: What are we proving by building an android project that depends on a plain old java shared project?  Is that a good use of build time?

brunobowden:
It's of limited value but it's only doing it for one project. I thought we should have a least one Android build, so that we know it's working across both that and iOS. At the moment, the plugin is only built on Linux and not used.

I'm open to suggestions on this. Perhaps only do the Android build on one Linux environment?

brunobowden:
Also, doing the Android builds on Linux, means that you can avoid it slowing down OS X builds.

brunobowden:
The Linux build would've caught the issue where we needed to disable j2objcXcode for translateOnlyMode. That's an example... but mysterious other cases like that which we don't anticipate are the reason to do this.

.travis.yml
```diff
@@ -4,13 +4,17 @@ language: generic
 matrix:
   include:
     - os: linux
-      language: groovy
+      language: android
       jdk: oraclejdk7
-      env: USING_OS=linux
+      env:
+        - TEST_SET=1
```
advayDev1: Can we keep one of the two linux runs without any system tests?
It was nice to be see if one build passed in 1-2 minutes.  Now it is 4+ minutes.

systemTests/run-all.sh
```diff
@@ -15,7 +15,8 @@
 # limitations under the License.
 
 
-set -ev
+# Fail if anything fails.
+set -uev
```
advayDev1: Does this work if $1 isn't set?  My intent was that with no argument specified, everything ran (like devs should locally)

advayDev1:
I can buy that.  I'm not opposed overall, but please do run the system tests only on one of the linux configs, that way we still get a 'sanity check' build passing in 1-2 minutes

LGTM + comments.

.travis.yml
brunobowden: I've turned off the tests for oraclejdk7 and left them on for openjdk7.

systemTests/run-all.sh
brunobowden: I encountered exactly this issue. Please take a look at the new version.

brunobowden:
Please note that you'll need to set ANDROID_HOME in your environment to use run-all.sh locally.

brunobowden:
I'm delighted to see that the build worked. The oraclejdk7 build was <2 minutes and the openjdk7 took 4.5 minutes.

PR #537, opened 2015-10-24
localJ2objcDist caching for Travis

brunobowden:
- Caches systemTests/common/localJ2objcDist
- Saves around 40 seconds for download and unzip


advayDev1:
If you push a no-op kick commit on this same PR and time goes down I'm worried...
For security, Travis should not modify the cache based on a PR since literally anyone can make a PR.  Travis says:
```
We fetch the repo’s cache on every build, including feature branches and pull requests.
There is one cache per branch and language version/compiler version/JDK version/Gemfile location/etc.
If a branch does not have its own cache yet, it will fetch the master branch cache.
Only modifications made to the cached directories from normal pushes are stored.
Currently Pull Requests will use the cache of the branch they are supposed to be merged into.
```

So a kick commit on this PR before you merge the caching PR into main possibly indicates an issue with Travis.

brunobowden:
Are you seeing a specific problem here? My impression is that this can't be fully tested until it's merged in to the main build.

advayDev1:
Yes, if Travis is correctly stating their security protections, correct.
However:
1.  On Linux, the logs claim to be caching both the JDK and the non-existent localJ2objcDist.  https://travis-ci.org/j2objc-contrib/j2objc-gradle/jobs/87148544
2.  On Mac, the cache seems to have no effect whatsoever - check the logs.  Perhaps their OSX virtual machines don't have this functionality yet.  https://travis-ci.org/j2objc-contrib/j2objc-gradle/jobs/87148546

advayDev1:
Confirmed, per http://docs.travis-ci.com/user/workers/container-based-infrastructure/ - caches work on Containers.  And only Linux uses Containers.  So this won't help us yet :(

brunobowden:
Fair enough. Closing this PR.

PR #539, opened 2015-10-25
convert podMethodName to valid Ruby

### LINKED ISSUE #538: '-' in project name creates Podfile that can't be installed
The reason for this is that '-' is an invalid character in a Ruby method name (the language used for the Cocoapods Podfile).

cc @andrewfunston

brunobowden:
- Replace all non-alphanumeric characters with underscore
- Fixes #538


brunobowden:
@andrewfunston - this is a fix for the issue you saw with '-' in a project name failing the build.

advayDev1:
lgtm

PR #541, opened 2015-10-25
README revamp

advayDev1:
* Make it clear what this plugin actually does up-front; it has become much more than just j2objc translation.
  For users of the plugin, the execution of J2ObjC is just one implementation detail really.
* Installation instructions are in the plugin already.
* Why Gradle 2.4 (vs others) is needed is also an implementation detail.
* Mentioning GWT is irrelevant to most people.  We could just as well mention web service implementation in Java, etc.
* Show a more normal shared/build.gradle - nearly everyone depends on *some* library.
* Use bullets to break up long text

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,11 +5,11 @@ so you can write an app's non-UI code (such as application logic and data models
 which is then shared by Android apps (natively Java) and iOS apps (using J2ObjC).
 
 The plugin:
-* translates your Java source code to Objective-C for iOS (iPhone/iPad) using [__J2ObjC__](https://github.com/google/j2objc), an open-source tool from Google
-* builds Objective-C static libraries and headers ready-to-use in Xcode
-* runs translated versions of your JUnit tests to ensure your code works in Objective-C form
-* handles multiple Java projects, external Java libraries \[1\], and existing Objective-C code you'd like to link in
-* configures Xcode projects to use your translated libraries, using CocoaPods (optionally)
+* Translates your Java source code to Objective-C for iOS (iPhone/iPad) using [__J2ObjC__](https://github.com/google/j2objc), an open-source tool from Google
+* Builds Objective-C static libraries and headers ready-to-use in Xcode
+* Runs translated versions of your JUnit tests to ensure your code works in Objective-C form
+* Handles multiple Java projects, external Java libraries \[1\], and existing Objective-C code you'd like to link in
+* Configures Xcode projects to use your translated libraries, using CocoaPods (optionally)
  
 The plugin is not affiliated with Google but was developed by former Google Engineers and others.
 Note that the plugin is currently in alpha; we may need to make breaking API changes
@@ -132,13 +132,13 @@ Please first check the [Frequently Asked Questions](FAQ.md).
 
 Next, search the [Issues](https://github.com/j2objc-contrib/j2objc-gradle/issues) for a similar
 problem.  If your issue is not addressed, please file a new Issue, including the following
-details (if you are comfortable sharing publicly as "Contribution(s)" per
-the [LICENSE](LICENSE)):
+details:
 - build.gradle file(s)
 - contents of Gradle build errors if any
 - version of J2ObjC you have installed
 
 If you are not comfortable sharing these, the community may not be able to help as much.
+Please note your bug reports will be treated as "Contribution(s)" per the [LICENSE](LICENSE).
 
 Mozilla's [Bug writing guidelines](https://developer.mozilla.org/en-US/docs/Mozilla/QA/Bug_writing_guidelines)
 may be helpful. Having public, focused, and actionable Issues
@@ -164,4 +164,4 @@ J2ObjC and libraries distributed with J2ObjC are under their own licenses.
 
 ### Footnotes
 
-[1].  <a href='#footnote1'>where source is appropriately licensed and available</a>
+[1]  <a href='#footnote1'>where source is appropriately licensed and available</a>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,13 +1,19 @@
...
-J2ObjC enables Java source to be part of an iOS application's build, no editing
-of the generated files is necessary. The goal is to write an app's non-UI code
-(such as application logic and data models) in Java, which is then shared by
-Android apps (natively Java), web apps (using GWT), and iOS apps (using J2ObjC).
+The __J2ObjC Gradle plugin__ enables Java source to be part of an iOS application's build
+so you can write an app's non-UI code (such as application logic and data models) in Java,
+which is then shared by Android apps (natively Java) and iOS apps (using J2ObjC).
+
+The plugin:
+* translates your Java source code to Objective-C for iOS (iPhone/iPad) using [__J2ObjC__](https://github.com/google/j2objc), an open-source tool from Google
```
brunobowden: Capitalize the first letter of each item. Repeat.

README.md
```diff
@@ -1,13 +1,19 @@
...
+* builds Objective-C static libraries and headers ready-to-use in Xcode
+* runs translated versions of your JUnit tests to ensure your code works in Objective-C form
+* handles multiple Java projects, external Java libraries (where source is appropriately licensed and available), and existing Objective-C code you'd like to link in
```
brunobowden: I'd suggest dropping the "(where source is appropriately licensed and available)" as that's something I would assume anyone would know about anyway.

README.md
```diff
@@ -1,13 +1,19 @@
...
+* configures your Xcode projects to use your translated code, without requiring manual editing of the generated files (optionally, when using CocoaPods)
```
brunobowden: Simplify to:

* Xcode project automatic configured to use your translated libraries, using CocoaPods

README.md
```diff
@@ -15,20 +21,20 @@ Android apps (natively Java), web apps (using GWT), and iOS apps (using J2ObjC).
...
 [Gradle Java plugin](https://docs.gradle.org/current/userguide/java_plugin.html).
-Starting as an empty project allows you to gradually shift over code from an existing
+Starting with an empty project allows you to gradually shift over code from an existing
 Android application. This is beneficial for separation between the application model
-and user interface. It also allows the shared project to be easily used server side as well.
+and user interface. It also allows the shared project to be easily used server-side as well.
 
-The Android app, shared Java project and Xcode should be sibling directories, i.e children
+The Android app, shared Java project and Xcode project should be sibling directories, i.e children
 of the same root level folder. Suggested folder names are `'android', 'shared' and 'ios'`
```
brunobowden: Should we replace 'android' with 'app' as that's what Android Studio chooses by default?

README.md
advayDev1: altered to be grammatically parallel, but done

README.md
advayDev1: If you look through j2objc-discuss, people seem to not realize that an APK alone does not suffice.
But I'll turn this into a footnote

README.md
```diff
@@ -37,8 +43,20 @@ plugins {
     id 'com.github.j2objccontrib.j2objcgradle' version '0.4.3-alpha'
 }
 
-// Plugin settings:
+dependencies {
+    compile 'com.google.j2objc:j2objc-annotations:0.9.8'
```
brunobowden: How are these needed? Don't libraries like Guava automatically use this? I see this as an advanced use case when you're dealing with CycleFinder... in which case this should go in the FAQ.

README.md
advayDev1: I don't think so: this is a single sentence with parallel verb clauses in bullet form.

README.md
advayDev1: I think we should leave it android for now.  My internal project was 'app' and 'ios-app' for a long time and it was super confusing

README.md
```diff
@@ -46,15 +64,15 @@ j2objcConfig {
 }
 ```
 
-Info on additional `j2objcConfig` settings are in
+For more complex situations like:
```
brunobowden: Move this section after `android/build.gradle`, so that the Quick Start build guidelines are together, then the advanced stuff comes afterwards.

README.md
advayDev1: True, I'll remove j2objc annotations.  But Guava and JUnit are pretty standard - I added them because having no dependencies in a build.gradle is at least something I've never seen.  We aren't illustrating what to add to an existing build.gradle; this section purports to tell you how to write a new build.gradle, where such things are normal.

README.md
advayDev1: done

README.md
```diff
@@ -103,6 +129,7 @@ be run as follows:
 
 Having issues with the plugin?
 Please first check the [Frequently Asked Questions](FAQ.md).
+
 Next, search the [Issues](https://github.com/j2objc-contrib/j2objc-gradle/issues) for a similar
 problem.  If your issue is not addressed, please file a new Issue, including the following
 details (if you are comfortable sharing publicly as "Contribution(s)" per
```
brunobowden: Though this comes from a previous PR, I think it's obvious that the report has to be public and redundant to mention it. I'd strike the section in parentheses and the line about "If you're are not comfortable share these..."

brunobowden:
LGTM with comments

README.md
advayDev1: on second thought, done.  Capitals look prettier :)

README.md
advayDev1: true it breaks the flow of the prose.  moved it further down.

advayDev1:
done and merged

PR #546, opened 2015-10-25
Faq 0.5 update

### LINKED ISSUE #542: README + FAQ update for 0.5
A market to let you know not to cut 0.5.
I'm fixing errors in the FAQ still

advayDev1:
one more!  fixes #542

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -31,7 +31,7 @@ Paste the results below, replacing existing contents.
 - [Why is my clean build failing?](#why-is-my-clean-build-failing)
 - [How do I include Java files from additional source directories?](#how-do-i-include-java-files-from-additional-source-directories)
 - [How do I develop with Swift?](#how-do-i-develop-with-swift)
-- [How do I solve 'File not found' import error in Xcode?](#how-do-i-solve-file-not-found-import-error-in-xcode)
+- [How do I manually configure my Xcode project to use the translated libraries?](#how-do-i-manually-configure-my-xcode-project-to-use-the-translated-libraries)
 - [How do I work with Package Prefixes?](#how-do-i-work-with-package-prefixes)
 - [How do I enable ARC for my translated Objective-C classes?](#how-do-i-enable-arc-for-my-translated-objective-c-classes)
 - [How do I call finalConfigure()?](#how-do-i-call-finalconfigure)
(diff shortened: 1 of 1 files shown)


FAQ.md
```diff
@@ -89,8 +89,9 @@ is opened in Xcode then CocoaPods will fail. This will appear as an Xcode build
     library not found for -lPods-IOS-APP-j2objc-shared
```
brunobowden: Change line 84 to somethig like:

the `xcodeProjectDir` in your shared `j2objcConfig` per the [Quick Start Guide](README.md#quick-start-guide).
The dependent Gradle projects (if any) should leave `xcodeProjectDir` as blank so that only one project
controls the Xcode updates.

FAQ.md
```diff
@@ -89,8 +89,9 @@ is opened in Xcode then CocoaPods will fail. This will appear as an Xcode build
...
 
 If you don't use CocoaPods, do not specify the `xcodeProjectDir` option;
```
brunobowden: Change to:

The above system uses Cocoapods and is the recommend way to build with Xcode. If you wish to avoid
using Cocoapods and configuring Xcode manually, then it can be done with...

FAQ.md
```diff
@@ -299,32 +300,29 @@ you'd like to access from Swift code. Also see the FAQ item on [file not found](
 ```
 
 
-### How do I solve 'File not found' import error in Xcode?
+### How do I manually configure my Xcode project to use the translated libraries?
 
-This is typically caused by a limitation of Xcode that expects all the source to be in a
-top-level directory. You need to use the `--no-package-directories` argument to flatten
-the hierarchy. The plugin will warn if two files are mapped to a conflicting name.
+If you do not want to use CocoaPods, you will need to modify your Xcode project's build settings.
```
brunobowden: Add line before:

Using Cocoapods is the recommended way to use the plugin. To configure Xcode manually...

brunobowden:
LGTM with comments

FAQ.md

FAQ.md
advayDev1: used 'quickest' - not 'recommended' since I don'

FAQ.md
advayDev1: done with 'quickest'

PR #550, opened 2015-10-26
Url

advayDev1:


brunobowden:
Why do we need to bring j2objccontrib.org in to it at all? Given that we're already so tightly integrated with GitHub, I don't see that changing... so I'm not sure what value this adds?

advayDev1:
Just easier to remember urls really

brunobowden:
I still think we're better off with the GitHub urls and people are more likely to be comfortable clicking on that. Unless we're building out the website, I don't think it has much value to add in another url.

advayDev1:
fair enough

PR #552, opened 2015-10-26
Resolve or suppress all technical debt warnings

advayDev1:
Covers everything from:
http://nemo.sonarqube.org/component_issues?id=com.github.j2objccontrib.j2objcgradle#resolved=false


brunobowden:
LGTM after you fix the build.

That was all the errors is reported? If so, then I think that's pretty good on our part.

advayDev1:
Yeah we fixed quite a few in the last few weeks.

PR #553, opened 2015-10-26
Correctly scope input and ouputs for AssembleLibrariesTask

### LINKED ISSUE #505: j2objcAssembleDebug fails UP-TO-DATE due to shared directory for debug / release build
For simple1 test, running `j2objcBuild` twice in succession shouldn't repeat running the same task twice. In this case, `j2objcAssembleDebug` fails the UP-TO-DATE on the 2nd build. It should pass. This occurs due to the debug and release builds sharing the same directory for the static libraries.

These folders need to be split, so that the debug and release libraries are written to separate directories.

```
:j2objcAssembleDebug (Thread[Daemon worker Thread 9,5,main]) started.
:j2objcAssembleDebug
Executing task ':j2objcAssembleDebug' (up-to-date check took 0.005 secs) due to:
  Input file /Users/brunobowden/dev/j2objc-gradle/systemTests/simple1/build/binaries/simple1-j2objcStaticLibrary/ios_arm64Release/libsimple1-j2objc.a has been added.
  Input file /Users/brunobowden/dev/j2objc-gradle/systemTests/simple1/build/binaries/simple1-j2objcStaticLibrary/ios_armv7Release/libsimple1-j2objc.a has been added.
  Input file /Users/brunobowden/dev/j2objc-gradle/systemTests/simple1/build/binaries/simple1-j2objcStaticLibrary/ios_x86_64Release/libsimple1-j2objc.a has been added.
```

advayDev1:
Fixes #505

TESTED=
- simple1 system test built twice idempotently
- then modified only debug files and it only rebuilt debug

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/AssembleLibrariesTask.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/AssembleLibrariesTask.groovy
@@ -19,9 +19,10 @@ package com.github.j2objccontrib.j2objcgradle.tasks
 import com.github.j2objccontrib.j2objcgradle.J2objcConfig
 import groovy.transform.CompileStatic
 import org.gradle.api.DefaultTask
+import org.gradle.api.file.FileTree
 import org.gradle.api.tasks.Input
-import org.gradle.api.tasks.InputDirectory
-import org.gradle.api.tasks.OutputDirectory
+import org.gradle.api.tasks.InputFiles
+import org.gradle.api.tasks.OutputDirectories
 import org.gradle.api.tasks.TaskAction
 
 /**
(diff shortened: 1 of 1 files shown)


advayDev1:
ptal @brunobowden

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/AssembleLibrariesTask.groovy
```diff
@@ -32,20 +33,42 @@ import org.gradle.api.tasks.TaskAction
...
     File srcLibDir
-
-    @InputDirectory
     File srcPackedLibDir
 
+    // Only treat as inputs the files of relevant buildType.
+    @InputFiles
+    FileTree getSrcLibs() {
+        return project.fileTree(dir: srcLibDir, include: "*$buildType/*.a") +
+               project.fileTree(dir: srcPackedLibDir, include: "*$buildType/*.a")
```
brunobowden: Should "*$buildType/*.a" be extracted since it's used in 3 places?

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/AssembleLibrariesTask.groovy
```diff
@@ -32,20 +33,42 @@ import org.gradle.api.tasks.TaskAction
...
     File getDestLibDirFile() { return J2objcConfig.from(project).getDestLibDirFile() }
 
+    // Only treat as outputs the directories of relevant buildType.
+    @SuppressWarnings('unused')
+    @OutputDirectories
+    List<File> getDestLibSubdirs() {
+        List<File> inputDirs = []
+        FileFilter filter = new FileFilter() {
+            @Override boolean accept(File pathname) {
+                return pathname.isDirectory() && pathname.name.endsWith(buildType)
```
brunobowden: Why this `isDirectory()` test? Add comment on this.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/AssembleLibrariesTask.groovy
```diff
@@ -32,20 +33,42 @@ import org.gradle.api.tasks.TaskAction
...
+            }
+        }
+        File[] filtered = srcLibDir.listFiles(filter)
+        if (filtered != null) {
```
brunobowden: When is this null? I'm not sure if that should ever happen. It might be that the Debug libraries aren't built but then shouldn't the assembleLibrariesDebug not be run either?

brunobowden:
LGTM with comments

In some ways, it would be easier splitting this in to a debug and release directory but I can see why that doesn't work as well.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/AssembleLibrariesTask.groovy
advayDev1: dont

advayDev1: er "done"

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/AssembleLibrariesTask.groovy
advayDev1: other stuff can be in there.  a .DSStore for example

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/AssembleLibrariesTask.groovy
advayDev1: It can legitimately be null when the parent directory doesn't exist. If the superior tasks are disabled, it would be correct for this task not to pack nothing.

When testing we don't actually create all these directories either.

Comment added.

advayDev1:
The directory naming structure follows the default one used by Gradle: namely [architecture][BuildType].  I wish it used hierarchy, but it doesn't.  Following this structure allows other Gradle native project to easily  link to the output of our plugin.

brunobowden:
Fair point on dependant projects. Agreed on keeping the structure as is.

PR #556, opened 2015-10-26
FAQ: How to build j2objc from xcode

### LINKED ISSUE #298: Building with j2objc-gradle within Xcode
My internal setup has both a) the final iOS/OSX app build-able by Gradle (with https://github.com/madvay/mixc-build) and b) the Java code buildable in Xcode.  While (a) needs to be separate from this plugin, I think (b) would be more appropriate to put here.  When you change your Java (say manually or just by git pull-ing the latest changes), the next Debug run in Xcode will first run both j2objc translation and j2objc-gradle native compilation before building the Xcode part of the project (with all of our usual partial translation/compilation magic).

In addition to convenience, the key benefit is that errors in your j2obc-gradle setup are much easier to debug.  Since j2objc's error messages match Xcode's format (thanks @tomball !), Xcode lets you jump straight into the Java file (for j2objc errors) or the resulting .m/.h transpiled file (for j2objc-gradle native compilation errors).

Of course, I don't use cocoa pods, so I'm not sure how to make our pod configure the resulting xcproj to do what mine does (my projects all use hand-rolled xcconfig files).

@brunobowden @confile - I'm going to need some pointers about pods if we want to do this.

advayDev1:
Fixes #298

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -32,6 +32,7 @@ Paste the results below, replacing existing contents.
 - [How do I include Java files from additional source directories?](#how-do-i-include-java-files-from-additional-source-directories)
 - [How do I develop with Swift?](#how-do-i-develop-with-swift)
 - [How do I manually configure my Xcode project to use the translated libraries?](#how-do-i-manually-configure-my-xcode-project-to-use-the-translated-libraries)
+- [How do I update my J2ObjC translated code from Xcode?](#how-do-i-update-my-j2objc-translated-code-from-xcode)
 - [How do I work with Package Prefixes?](#how-do-i-work-with-package-prefixes)
 - [How do I enable ARC for my translated Objective-C classes?](#how-do-i-enable-arc-for-my-translated-objective-c-classes)
 - [How do I call finalConfigure()?](#how-do-i-call-finalconfigure)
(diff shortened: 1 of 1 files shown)


FAQ.md
```diff
@@ -329,6 +330,37 @@ In each case, if the setting has existing values, append the ones above.
...
+  #!/bin/sh
+  # Don't contaminate the build with Xcode env vars.
+  unset DEVELOPER_DIR
+  # CUSTOMIZE: Add any needed paths here.
+  export PATH=/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin
+  if [ "$ACTION" != "clean" ]; then
+      # CUSTOMIZE: List the j2objcBuild task for each J2ObjC project.
+      ./gradlew shared:j2objcBuild -q
+  else
+      echo "Skipping Gradle Build for Clean"
```
brunobowden: What about other actions, such as build?

FAQ.md
```diff
@@ -329,6 +330,37 @@ In each case, if the setting has existing values, append the ones above.
 Wherever `shared` appears above, duplicate that value for every J2ObjC project used,
 including all transitive dependencies.
 
+
+### How do I update my J2ObjC translated code from Xcode?
+
+You can add the J2ObjC build as an External Build System target:
```
brunobowden: Add line before...

It is suggested to first develop and debug in Android, then after that is working to build in iOS. If you want to develop directly from Java to iOS, then the J2ObjC build can be initiated from Xcode as an External Build System target:

FAQ.md
advayDev1: Not sure what you mean.  For all actions except Clean, the project will be builtt.

FAQ.md
advayDev1: adjusted and done

FAQ.md
```diff
@@ -329,6 +330,42 @@ In each case, if the setting has existing values, append the ones above.
...
+
+It is suggested to first develop and debug in Android, then after that is working to
+build in iOS. If you want to update your generated Objective-C code while working in
+Xcode, then you can add the J2ObjC build as an External Build System target:
+
+1.  Select the main iOS project.  File -> New Target -> Other -> External Build System.
+    Set the build tool as `/bin/sh`, and name the target j2objcGradle.
+2.  In the 'External Build Tool Configuration' screen that appears, set the following values:
+  * Build Tool: `/bin/sh`
+  * Arguments: `./gradlePreBuild.sh`
```
brunobowden: How about labeling it `./gradleSharedJ2objcBuild.sh` since that's closer to the command being run.

FAQ.md
brunobowden: Got it. I misread it before.

brunobowden:
LGTM with comment

FAQ.md

PR #558, opened 2015-10-27
Podspec output to build/j2objcOutputs

### LINKED ISSUE #549: Podspecs output should be build/j2objcOutputs directory
We should not put any files in project/build/.  They should be in project/build/something.

brunobowden:
- Debug and release podspecs output to build/j2objcOutputs
- destPodspecDir to specify output directory for pod specs
- prevent podspec paths that traverse parent
- unit tests updated for these changes
- Fixes #549


advayDev1:
Can you please add to CHANGELOG.md, esp. if a migration step will be required for users from 0.5.0-alpha to our next release? 

LGTM otherwise

brunobowden:
Since the podspecs were already in the build directory, this will be a seamless move that likely no one will notice.

As for CHANGELOG.md, I thought it would be easier to wait until a release, review the recent changes and then select what's appropriate. I'd prefer not to get in the habit of trying to update that for each PR unless the release is close to completion.

PR #563, opened 2015-11-04
Fix documentation links from plugin

brunobowden:
Fixes a number of broken documentation links for the plugin.

All the github urls have been checked.


advayDev1:
lgtm

PR #564, opened 2015-11-06
Allow spaces in Xcode targets

brunobowden:
- Allow spaces in Xcode targets
- Continue to reject spaces for command line arguments
- Additionally reject arguments that are only whitespace
- For watchOS, change unsupported exception to warning for development


brunobowden:
Turns out that many Xcode targets have spaces by default, particularly the watch app and extensions

advayDev1:
As a naked boolean, methods like appendNoWhitespaceArgs and appendWhitespaceArgs (and similarly named verify... methods) might be more readable.  Your call, LGTM either way.

brunobowden:
I tried both ways and slightly preferred this. Though I see points in both directions.

PR #565, opened 2015-11-08
Config to only add j2objc_shared method to Podfile

### LINKED ISSUE #561: Config to only add j2objc_shared method to Podfile and not update targets
For a more complex Podfile with nested Ruby code, it may not be possible for the plugin to parse and correctly update the Podfile. In that case, it should have an option to:

1. Plugin only add / updates the `def j2objc_shared` method and doesn't make any other changes
1. Allow the user to manually edit the Podfile as needed

saruye:
@brunobowden 
I started working on issue #561. This is not finished. I just opened this PR already to get some feedback if I am on the right track. 

My idea was: Add an extra property that can disable updating the targets. Default will be as it is. If the property is set, it will disable all checks for targets and only add/update the j2objc method to the Podfile. Is that the intention?

My questions: 
naming: since you renamed the ticket I guess you prefer the naming to be something about "only to add the method" instead of "disabling updating of the target"? in the task I used updateTargets now as a boolean since that seemed more straight forward. Is that ok?
So far the method is always added before the target instructions. Where should it go if this cannot be found? Is it possible to just add it at the beginning of the file? Maybe before the 'source ...' ?

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
@@ -664,8 +664,6 @@ class J2objcConfig {
      */
     String xcodeProjectDir = null
 
-    boolean onlyAddJ2ObjcToPodfile = false
-
     /**
      * iOS app and test Xcode targets to link to the generated libraries.
      *
@@ -713,6 +711,16 @@ class J2objcConfig {
     void xcodeTargetsWatchos(String... xcodeTargetsWatchos) {
         appendArgs(this.xcodeTargetsWatchos, 'xcodeTargetsWatchos', xcodeTargetsWatchos)
     }
+
+    /**
+     * Allows manual config of the Podfile (default is false).
+     *
+     * When set to true, this allows manual configuring of the Podfile targets.
+     * This is necessary when your Podfile is too complex to be automatically
+     * updated. When used, you must also set xcodeTargets{Ios|Osx|Watchos)
+     * to empty.
+     */
+    boolean xcodeTargetsManualConfig = false
 
 
     protected boolean finalConfigured = false
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
@@ -664,6 +664,8 @@ class J2objcConfig {
      */
     String xcodeProjectDir = null
 
+    boolean onlyAddJ2ObjcToPodfile = false
+
     /**
      * iOS app and test Xcode targets to link to the generated libraries.
      *
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy
@@ -49,6 +49,8 @@ class XcodeTask extends DefaultTask {
     public static final String targetNamedRegex = /^\s*target\s+'TARGET'\s+do\s*$/
     public static final String podMethodStartRegex = /^\s*((def\s*j2objc_)|(# J2ObjC Gradle Plugin)).*/
     public static final String endRegex = /^\s*end\s*/
+    public static final String sourceLineRegex = /^source 'http.*Specs.git'$/  // https://guides.cocoapods.org/syntax/podfile.html#podfile
+    public static final String platformLineRegex = /^platform :.*/
 
     @Input @Optional
     String getXcodeProjectDir() { return J2objcConfig.from(project).xcodeProjectDir }
@@ -320,6 +322,41 @@ class XcodeTask extends DefaultTask {
     }
 
     /**
+     * Replaces lines between two regexps
+     */
+    @VisibleForTesting
+    static List<String> regexReplaceLines(List<String> podfileLines,
+                                        String startRegex, String endRegex, List<String> newPodFileLines) {
+        List<String> result = new ArrayList<>()
+        boolean active = false
+        boolean endOfRegex = false
+
+        for (line in podfileLines) {
+            if (!active && (line =~ startRegex).find()) {
+                active = true
+                result.addAll(newPodFileLines)
+            }
+            if (active && (line =~ endRegex).find()) {
+                active = false
+                endOfRegex = true
+            }
+            if (!active && !endOfRegex) {
+                result.add(line)
+            }
+
+            if(endOfRegex){
+                endOfRegex=false
+            }
+        }
+        if (active) {
+            throw new InvalidUserDataException(
+                    "Failed to find endRegex: ${Utils.escapeSlashyString(endRegex)}\n" +
+                            podfileLines.join('\n'))
+        }
+        return result
+    }
+
+    /**
      * Insert new lines in to podfile between startRegex to endRegex.
      *
      * Throws error for no match or multiple matches.
@@ -372,6 +409,116 @@ class XcodeTask extends DefaultTask {
     }
 
     /**
+     * Searches for the line where to insert new methods into the podfile
+     */
+    @VisibleForTesting
+    static int findInsertLine(List<String> podfileLines) {
+
+        boolean noneCommentLineFound =false
+        boolean platformStatementFound =false
+        boolean sourceStatementFound =false
+
+        int counter = 0
+        int insertLineNumber = 0
+        for (line in podfileLines) {
+
+            // if a target statement is found. insert before this statement
+            if(line =~ targetStartRegex) {
+                return counter
+            }
+
+            counter++
+
+            // else insert after platform statement (platform :ios, '6.0')
+            if(platformStatementFound) continue
+            if(line =~ platformLineRegex) {
+                insertLineNumber = counter
+                platformStatementFound = true
+                continue
+            }
+
+            // else insert after platform statement (source 'https://github.com/CocoaPods/Specs.git')
+            if(sourceStatementFound) continue
+            if(line =~ sourceLineRegex) {
+                insertLineNumber = counter
+                sourceStatementFound = true
+                continue
+            }
+
+            // else insert at first none commented line
+            if(noneCommentLineFound) continue
+
+            if(!line.trim().startsWith("#")){
+                insertLineNumber = counter-1
+                noneCommentLineFound = true
+            }
+
+        }
+
+        return insertLineNumber
+    }
+
+    /**
+     * Insert new lines in to podfile at a certain line
+     * @param surroundWithEmptyLines regulates if inserted lines should be surrounded with empty ones
+     *
+     */
+    @VisibleForTesting
+    static List<String> insertAtLineNumber(List<String> podfileLines,List<String> insertLines, int line, boolean surroundWithEmptyLines) {
+        List<String> result = new ArrayList<>()
+        result.addAll(podfileLines.subList(0,line))
+
+        assert podfileLines.size() >= line
+
+        boolean insertAtTheEnd = podfileLines.size()==line
+        boolean insertAtTheBeginning = line==0
+
+        if(surroundWithEmptyLines){
+            if(!insertAtTheBeginning){
+                if(!isBlankLine(podfileLines,line-1)){
+                    result.add('')
+                }
+                if(isBlankLine(podfileLines,line)){
+                    line++
+                }
+            }
+            if(!insertAtTheEnd){
+                insertLines.add('')
+                if(isBlankLine(podfileLines,line)){
+                    line++
+                }
+            }
+        }
+        result.addAll(insertLines)
+        if(!insertAtTheEnd){
+            result.addAll(podfileLines.subList(line,podfileLines.size()))
+        }
+        return result
+    }
+
+    static boolean isBlankLine(List<String> podfileLines,int line){
+        if(line>=podfileLines.size()){
+            return false
+        }
+        return podfileLines.get(line).trim()==''
+    }
+
+    /**
+     * Checks if Podfile contains a certain regex
+     *
+     */
+    @VisibleForTesting
+    static boolean containsRegex(List<String> podfileLines, String startRegex) {
+
+        for (line in podfileLines) {
+            if ((line =~ startRegex).find()) {
+                return true
+            }
+        }
+        return false
+    }
+
+    /**
      * Modify in place the existing podfile.
      */
     @VisibleForTesting
@@ -405,14 +552,14 @@ class XcodeTask extends DefaultTask {
 
         boolean xcodeTargetsAllEmpty =
                 xcodeTargetDetails.xcodeTargetsIos.isEmpty() &&
-                        xcodeTargetDetails.xcodeTargetsOsx.isEmpty() &&
-                        xcodeTargetDetails.xcodeTargetsWatchos.isEmpty()
+                xcodeTargetDetails.xcodeTargetsOsx.isEmpty() &&
+                xcodeTargetDetails.xcodeTargetsWatchos.isEmpty()
 
         if (xcodeTargetsManualConfig) {
             if (!xcodeTargetsAllEmpty) {
                 throw new InvalidUserDataException(
                         "Xcode targets must all be blank when using xcodeTargetsManualConfig.\n" +
-                                "Please update j2objcConfig by removing xcodeTargetsIos, xcodeTargetsOsx & xcodeTargetsWatchos")
+                        "Please update j2objcConfig by removing xcodeTargetsIos, xcodeTargetsOsx & xcodeTargetsWatchos")
             }
         } else {
             // xcodeTargetsManualConfig = false  (default)
@@ -421,19 +568,17 @@ class XcodeTask extends DefaultTask {
             verifyTargets(xcodeTargetDetails.xcodeTargetsOsx, podfileTargets, 'xcodeTargetsOsx')
             verifyTargets(xcodeTargetDetails.xcodeTargetsWatchos, podfileTargets, 'xcodeTargetsWatchos')
 
-            if (xcodeTargetDetails.xcodeTargetsIos.isEmpty() &&
-                    xcodeTargetDetails.xcodeTargetsOsx.isEmpty() &&
-                    xcodeTargetDetails.xcodeTargetsWatchos.isEmpty()) {
+            if (xcodeTargetsAllEmpty) {
                 // Give example for configuring iOS as that's the common case
                 throw new InvalidUserDataException(
                         "You must configure the xcode targets for the J2ObjC Gradle Plugin.\n" +
-                                "It must be a subset of the valid targets: '${podfileTargets.join("', '")}'\n" +
-                                "\n" +
-                                "j2objcConfig {\n" +
-                                "    xcodeTargetsIos 'IOS-APP', 'IOS-APPTests'  // example\n" +
-                                "}\n" +
-                                "\n" +
-                                "Can be optionally configured for xcodeTargetsOsx and xcodeTargetsWatchos\n")
+                        "It must be a subset of the valid targets: '${podfileTargets.join("', '")}'\n" +
+                        "\n" +
+                        "j2objcConfig {\n" +
+                        "    xcodeTargetsIos 'IOS-APP', 'IOS-APPTests'  // example\n" +
+                        "}\n" +
+                        "\n" +
+                        "Can be optionally configured for xcodeTargetsOsx and xcodeTargetsWatchos\n")
             }
             newPodfileLines = updatePodfileTargets(newPodfileLines, podspecDetailsList, xcodeTargetDetails)
         }
@@ -449,8 +594,8 @@ class XcodeTask extends DefaultTask {
             if (! podfileTargets.contains(xcodeTarget)) {
                 throw new InvalidUserDataException(
                         "Invalid j2objcConfig { $xcodeTargetsName '$xcodeTarget' }\n" +
-                                "Must be one of the valid targets: '${podfileTargets.join("', '")}'\n" +
-                                "NOTE: if your Podfile is too complex, you may need to use xcodeTargetsManualConfig")
+                        "Must be one of the valid targets: '${podfileTargets.join("', '")}'\n" +
+                        "NOTE: if your Podfile is too complex, you may need to use xcodeTargetsManualConfig")
             }
         }
     }
@@ -459,10 +604,6 @@ class XcodeTask extends DefaultTask {
     static List<String> updatePodMethods(
             List<String> podfileLines, List<PodspecDetails> podspecDetailsList, File podfile) {
 
-        // strip all old methods
-        // Note: use of preserveEndLine=true so that the targetStartRegex isn't removed
-        List<String> newPodfileLines =
-                regexStripLines(podfileLines, false, podMethodStartRegex, targetStartRegex, /.*/)
 
         // create new methods
         List<String> insertLines = new ArrayList<>()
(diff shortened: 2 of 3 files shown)


src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
```diff
@@ -664,6 +664,8 @@ class J2objcConfig {
      */
     String xcodeProjectDir = null
 
+    boolean onlyAddJ2ObjcToPodfile = false
```
brunobowden: 1) Rename so that it's more clearly associated with the xcodeTargetsIos config:

    xcodeTargetsManualConfig

2) Add the javadoc to it indicating what it does:

    /**
     * Allows manual config of the Podfile (default is false).
     *
     * When set to true, this allows manual configuring of the Podfile targets.
     * This is necessary when your Podfile is too complex to be automatically
     * updated. When used, you must also set xcodeTargets{Ios|Osx|Watchos)
     * to empty.
     */

3) move this after the xcodeTargetsWatchos setting, so it's the last one listed.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy
```diff
@@ -53,6 +53,9 @@ class XcodeTask extends DefaultTask {
     @Input @Optional
     String getXcodeProjectDir() { return J2objcConfig.from(project).xcodeProjectDir }
 
+    @Input
+    boolean isOnlyAddJ2ObjcToPodfile() { return J2objcConfig.from(project).onlyAddJ2ObjcToPodfile }
```
brunobowden: Follow the existing convention and use a "get" prefix on the existing name, with first letter capitalized:

    getXcodeTargetsManualConfig

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy
```diff
@@ -374,14 +377,14 @@ class XcodeTask extends DefaultTask {
     @VisibleForTesting
     static void writeUpdatedPodfileIfNeeded(
             List<PodspecDetails> podspecDetailsList,
-            XcodeTargetDetails xcodeTargetDetails,
+            XcodeTargetDetails xcodeTargetDetails, boolean updateTargets,
```
brunobowden: Follow existing conventions and put new parameter on separate line.

Also, rename to `xcodeTargetsManualConfig`.... I strongly believe in consistent naming so that it's easier for a developer to understand code as they're reading it.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy
```diff
@@ -374,14 +377,14 @@ class XcodeTask extends DefaultTask {
...
             File podfile) {
 
         List<String> oldPodfileLines = podfile.readLines()
         List<String> newPodfileLines = new ArrayList<String>(oldPodfileLines)
 
         newPodfileLines = updatePodfile(
-                newPodfileLines, podspecDetailsList, xcodeTargetDetails, podfile)
+                newPodfileLines, podspecDetailsList, xcodeTargetDetails,updateTargets, podfile)
```
brunobowden: Need space after comma.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy
```diff
@@ -393,34 +396,38 @@ class XcodeTask extends DefaultTask {
     static List<String> updatePodfile(
             List<String> podfileLines,
             List<PodspecDetails> podspecDetailsList,
-            XcodeTargetDetails xcodeTargetDetails,
+            XcodeTargetDetails xcodeTargetDetails,boolean updateTargets,
```
brunobowden: Parameter on new line.

Rename to `xcodeTargetsManualConfig`.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy
```diff
@@ -393,34 +396,38 @@ class XcodeTask extends DefaultTask {
...
             File podfile) {
 
-        List<String> podfileTargets = extractXcodeTargets(podfileLines)
-        verifyTargets(xcodeTargetDetails.xcodeTargetsIos, podfileTargets, 'xcodeTargetsIos')
-        verifyTargets(xcodeTargetDetails.xcodeTargetsOsx, podfileTargets, 'xcodeTargetsOsx')
-        verifyTargets(xcodeTargetDetails.xcodeTargetsWatchos, podfileTargets, 'xcodeTargetsWatchos')
+        if(updateTargets){
```
brunobowden: Need space between `if` and `(`.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy
```diff
@@ -200,7 +203,7 @@ class XcodeTask extends DefaultTask {
                 getXcodeTargetsIos(), getXcodeTargetsOsx(), getXcodeTargetsWatchos(),
                 getMinVersionIos(), getMinVersionOsx(), getMinVersionWatchos())
 
-        writeUpdatedPodfileIfNeeded(podspecDetailsList, xcodeTargetDetails, podfile)
+        writeUpdatedPodfileIfNeeded(podspecDetailsList, xcodeTargetDetails,!isOnlyAddJ2ObjcToPodfile(), podfile)
```
brunobowden: By using a similar variable name (and meaning) throughout the code, you should avoid things like negation. When you make a change like this, a developer has to think "it's X over here" then "it's not-X over here". That's a significant addition in mental complexity for someone else reading the code that you should avoid.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy
```diff
@@ -393,34 +396,38 @@ class XcodeTask extends DefaultTask {
...
+                        "Can be optionally configured for xcodeTargetsOsx and xcodeTargetsWatchos\n")
+            }
         }
 
         // update pod methods
         List<String> newPodfileLines = updatePodMethods(podfileLines, podspecDetailsList, podfile)
 
         // update pod targets
-        newPodfileLines = updatePodfileTargets(newPodfileLines, podspecDetailsList, xcodeTargetDetails)
+        if(updateTargets){
```
brunobowden: Combine this section with the previous `if (updateTargets)` section. For mental complexity, that conditional should all be together. Then likely put line 425, the "updatePodMethods" call before that section again.

This is how it is currently:
```
if (a) {
   foo1()
}

bar()

if (a) {
   foo2()
}
```

See how much simpler it is to understand this instead:

```
bar()

if (a) {
   foo1()
   foo2()
}
```

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTaskTest.groovy
```diff
@@ -465,7 +465,7 @@ class XcodeTaskTest {
         List<String> newPodfileLines = XcodeTask.updatePodfile(
                 podfileLines,
                 podspecDetailsProj,
-                xcodeTargetDetailsIosAppOnly,
+                xcodeTargetDetailsIosAppOnly, true,
```
brunobowden: If there's already a convention of putting each parameter on a new line, then do it. It's confusing to mix styles, you should adopt the same local style that already exists.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTaskTest.groovy
```diff
@@ -485,7 +485,39 @@ class XcodeTaskTest {
...
                 newPodfileLines,
                 podspecDetailsProj,
-                xcodeTargetDetailsIosAppOnly,
+                xcodeTargetDetailsIosAppOnly, true,
+                new File('/SRC/ios/Podfile'))
+        assert expectedPodfileLines == newPodfileLines
+    }
+
+    @Test
+    void testUpdatePodfile_onlyAddMethod() {
```
brunobowden: Rename to: `testUpdatePodfile_xcodeTargetsManualConfig`

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTaskTest.groovy
```diff
@@ -509,7 +541,7 @@ class XcodeTaskTest {
         List<String> newPodfileLines = XcodeTask.updatePodfile(
                 podfileLines,
                 podspecDetailsProj,
-                xcodeTargetDetails,
+                xcodeTargetDetails, true,
```
brunobowden: Adopt local style of new line for each parameter

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTaskTest.groovy
```diff
@@ -539,7 +571,7 @@ class XcodeTaskTest {
         newPodfileLines = XcodeTask.updatePodfile(
                 newPodfileLines,
                 podspecDetailsProj,
-                xcodeTargetDetails,
+                xcodeTargetDetails, true,
```
brunobowden: Adopt local style of new line for each parameter

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy
```diff
@@ -393,34 +396,38 @@ class XcodeTask extends DefaultTask {
...
+                // Give example for configuring iOS as that's the common case
+                throw new InvalidUserDataException(
+                        "You must configure the xcode targets for the J2ObjC Gradle Plugin.\n" +
+                        "It must be a subset of the valid targets: '${podfileTargets.join("', '")}'\n" +
+                        "\n" +
+                        "j2objcConfig {\n" +
+                        "    xcodeTargetsIos 'IOS-APP', 'IOS-APPTests'  // example\n" +
+                        "}\n" +
+                        "\n" +
+                        "Can be optionally configured for xcodeTargetsOsx and xcodeTargetsWatchos\n")
```
brunobowden: Add additional line:

    This check can be disabled with "xcodeTargetsManualConfig = true"

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy
```diff
@@ -393,34 +396,38 @@ class XcodeTask extends DefaultTask {
...
+            }
```
brunobowden: Rewrite logic so that it forces xcode targets to be empty when doing manual config. It should be either or when configuring this.

```
boolean xcodeTargetsAllEmpty =
        xcodeTargetDetails.xcodeTargetsIos.isEmpty() &&
        xcodeTargetDetails.xcodeTargetsOsx.isEmpty() &&
        xcodeTargetDetails.xcodeTargetsWatchos.isEmpty()

if (xcodeTargetsManualConfig) {
    if (!xcodeTargetsAllEmpty) {
        throw new InvalidUserDataException(
                "Xcode targets must all be blank when using xcodeTargetsManualConfig.\n" +
                "Please update j2objcConfig by removing xcodeTargetsIos, xcodeTargetsOsx & xcodeTargetsWatchos")
    }
} else {
    // xcodeTargetsManualConfig = false  (default)
    List<String> podfileTargets = extractXcodeTargets(podfileLines)
    verifyTargets(...
    if (xcodeTargetsAllEmpty) {
        existing exception
    }
}
```

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy
```diff
@@ -393,34 +396,38 @@ class XcodeTask extends DefaultTask {
...
         List<String> newPodfileLines = updatePodMethods(podfileLines, podspecDetailsList, podfile)
 
         // update pod targets
-        newPodfileLines = updatePodfileTargets(newPodfileLines, podspecDetailsList, xcodeTargetDetails)
+        if(updateTargets){
+             newPodfileLines = updatePodfileTargets(newPodfileLines, podspecDetailsList, xcodeTargetDetails)
+        }
 
         return newPodfileLines
     }
```
brunobowden: For the `verityTargets` method below (I have to comment on this line as I can't comment in the review on a line that hasn't been modified). Within that method, add another line to explain how to handle a complex Podfile. My supposition is that in that scenario, this is where it will fail, so it's the most useful spot to add an explanation.

    private static verifyTargets(List<String> xcodeTargets, List<String> podfileTargets, xcodeTargetsName) {
        xcodeTargets.each { String xcodeTarget ->
            if (! podfileTargets.contains(xcodeTarget)) {
                throw new InvalidUserDataException(
                        "Invalid j2objcConfig { $xcodeTargetsName '$xcodeTarget' }\n" +
                        "Must be one of the valid targets: '${podfileTargets.join("', '")}'\n" +
                        "NOTE: if your Podfile is too complex, you may need to use xcodeTargetsManualConfig")
            }
        }
    }

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTaskTest.groovy
```diff
@@ -485,7 +485,39 @@ class XcodeTaskTest {
...
+                "",
+                "target 'IOS-APP' do",
+                "end"]
+        assert expectedPodfileLines == newPodfileLines
+
+        // Second time around is a no-op
+        newPodfileLines = XcodeTask.updatePodfile(
+                newPodfileLines,
+                podspecDetailsProj,
+                xcodeTargetDetailsIosAppOnly, false,
```
brunobowden: New line for new parameter

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTaskTest.groovy
```diff
@@ -485,7 +485,39 @@ class XcodeTaskTest {
...
                 new File('/SRC/ios/Podfile'))
         assert expectedPodfileLines == newPodfileLines
     }
```
brunobowden: Add another test where it guides the user on how to handle a complex, non-standard Podfile:

```
testUpdatePodfile_complexPodfileNeedsXcodeTargetsManualConfig() {
    List<String> podfileLines = [
        // Construct target name by concatenation, in a way that automated detection fails
        "target ('IOS-' + 'APP') do",
        "end"]

    expectedException.expect(InvalidUserDataException.class)
    expectedException.expectMessage('NOTE: if your Podfile is too complex, you may need to use xcodeTargetsManualConfig')

    // Second time around is a no-op
    newPodfileLines = XcodeTask.updatePodfile(
        newPodfileLines,
        podspecDetailsProj,
        xcodeTargetDetailsIosAppOnly,
        false,
        new File('/SRC/ios/Podfile'))
}
```

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTaskTest.groovy
```diff
@@ -583,11 +615,30 @@ class XcodeTaskTest {
                 [],
                 new XcodeTask.XcodeTargetDetails(
                         ['TARGET-DOES-NOT-EXIST'], [], [],
-                        '6.0.0', '10.6.0', '1.0.0'),
+                        '6.0.0', '10.6.0', '1.0.0'), true,
                 null)
     }
 
     @Test
+    void testUpdatePodfileNoTargets_onlyAddMethod() {
```
brunobowden: The first part of the test name is the method under test, so the "NoTargets" part should not be there.

This doesn't seem to add much value compared to the existing `testUpdatePodfile_xcodeTargetsManualConfig` test. I'm not sure if it's worth having it.

A better test would be a more complex but still valid Podfile. Perhaps based on your own which I presume meant that you came across this problem. In any case, if you add that, then it should likely be next to the `testUpdatePodfile_xcodeTargetsManualConfig` test, perhaps named `testUpdatePodfile_xcodeTargetsManualConfigComplexPodfile`.

brunobowden:
Hi @saruye, thanks for your work on this. I think that generally this is on the right track. I came across this when looking at some complex Podfiles, so I thought that eventually people would need it.

To answer your questions:

My questions: 
naming: since you renamed the ticket I guess you prefer the naming to be something about "only to add the method" instead of "disabling updating of the target"?

- I wasn't consciously thinking about this. I generally think of naming a setting being the most important for someone using the plugin, then secondly important for someone reading the code as a developer. In this case, I think it's helpful to associate it with the xcodeTargetsIos-like settings... then also explain when it's most likely needed (e.g. the verifyTargets exception).

in the task I used updateTargets now as a boolean since that seemed more straight forward. Is that ok?

- Boolean is fine and makes sense. For a developer reading the code, I find tremendous value to have consistent naming throughout the stack. The same name from the config to the internal variable is very helpful. That's lost when you change names and meanings.

So far the method is always added before the target instructions. Where should it go if this cannot be found?

- Fair point. Part of the value for this is the ability to remove and update the methods when they change over time. By placing the code before the target, there's an ability to easily wipe and replace it. If there's no target then that becomes harder.

Two thoughts:

1) Probably need a START and END of the custom shared methods. So if this is found, then it's updated in place, regardless of where it's moved to - not sure if Ruby requires you to declare methods before they're used.

```
# J2ObjC Gradle Plugin - DO NOT MODIFY from here to similar line below"
# This block may be moved as a complete unit around the Podfile
...
# J2ObjC Gradle Plugin - DO NOT MODIFY from here to similar line above"
```

2) When trying to place the shared code, use the existing if any, then try to do it before the first targets, if both of those aren't possible, then do it after the end of the first comment section.

Is it possible to just add it at the beginning of the file? Maybe before the 'source ...' ?

- I think you're on the right track. I don't know you mean by `source...`, this is what I see in my Podfile:

```
# Uncomment this line to define a global platform for your project
# platform :ios, '6.0' asdf 

use_frameworks!
```

So I think that line 3 being the first line that isn't a comment could work. As above though, if the code has been moved elsewhere in the file, then it should be updated in place.

superarne:
I tried to implement all your feedback. I didn't get to implement the actual updating logic though.  My idea is to replace the regexStripLines with somethingLike regexReplaceLines and then update the insert logic only to run if nothing was replaced and to insert it at the beginning if no targetStartRegex is found. Try to get to it sometime this week.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy
```diff
@@ -394,33 +398,48 @@ class XcodeTask extends DefaultTask {
...
-                    "It must be a subset of the valid targets: '${podfileTargets.join("', '")}'\n" +
-                    "\n" +
-                    "j2objcConfig {\n" +
-                    "    xcodeTargetsIos 'IOS-APP', 'IOS-APPTests'  // example\n" +
-                    "}\n" +
-                    "\n" +
-                    "Can be optionally configured for xcodeTargetsOsx and xcodeTargetsWatchos\n")
+        boolean xcodeTargetsAllEmpty =
+                xcodeTargetDetails.xcodeTargetsIos.isEmpty() &&
+                        xcodeTargetDetails.xcodeTargetsOsx.isEmpty() &&
```
brunobowden: We've generally avoided the style of extra indentation for multiple lines. So lines 408 and 409 should start at the same place as line 407.

This applies several times below as well.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy
```diff
@@ -394,33 +398,48 @@ class XcodeTask extends DefaultTask {
...
+        } else {
+            // xcodeTargetsManualConfig = false  (default)
+            List<String> podfileTargets = extractXcodeTargets(podfileLines)
+            verifyTargets(xcodeTargetDetails.xcodeTargetsIos, podfileTargets, 'xcodeTargetsIos')
+            verifyTargets(xcodeTargetDetails.xcodeTargetsOsx, podfileTargets, 'xcodeTargetsOsx')
+            verifyTargets(xcodeTargetDetails.xcodeTargetsWatchos, podfileTargets, 'xcodeTargetsWatchos')
+
+            if (xcodeTargetDetails.xcodeTargetsIos.isEmpty() &&
+                    xcodeTargetDetails.xcodeTargetsOsx.isEmpty() &&
+                    xcodeTargetDetails.xcodeTargetsWatchos.isEmpty()) {
```
brunobowden: LInes 424-426 should re-use the `xcodeTargetsAllEmpty` variable you created above.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTaskTest.groovy
```diff
@@ -46,6 +46,9 @@ class XcodeTaskTest {
                 'PROJ',
                 new File('/SRC/PROJ/BUILD/j2objc-PROJ-debug.podspec'),
                 new File('/SRC/PROJ/BUILD/j2objc-PROJ-release.podspec'))]
+    XcodeTask.XcodeTargetDetails xcodeTargetDetailsEmpty =
+            new XcodeTask.XcodeTargetDetails(
+                    [], [], [],null,null,null)
```
brunobowden: Add spaces after commas.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTaskTest.groovy
```diff
@@ -486,6 +490,41 @@ class XcodeTaskTest {
...
+                false,
+                new File('/SRC/ios/Podfile'))
+        assert expectedPodfileLines == newPodfileLines
+    }
+
+    @Test
+    void testUpdatePodfile_xcodeTargetsManualConfig() {
+        List<String> podfileLines = [
+                "target 'IOS-APP' do",
+                "end"]
```
brunobowden: I think this is better demonstrated with a more complex Podfile example. Where there loops or other strange ruby logic that is impossible to update normally. That's the important example to have here.

brunobowden:
@superarne, @saruye - are you the same user? It's just a little more confusing when you're using different accounts.

This is definitely coming along. Definitely make sure that all the unit tests pass. When it's ready, ping me again all I'll go do a final review.

saruye:
hi. yeah that was me. I was logged in with a wrong account. sry :/ 
I implemented the logic now for adding the generated method to the podfile. I split it into two parts. One that looks where to put it and the other one that does the insert trying to make sure it has empty lines before and after. I did update one or two existing tests where I thought the new behavior with empty lines is nicer. I added a test with a Podfile which looks similar to ours. This and all old tests pass. 
Is there by any chance a codestyle file for intellij that I could import? 
Could you have a look at me code and my assumptions regarding where to insert the new lines and if you would like more/less/other tests. Don't bother with the style problems for now. I will fix those once you are fine with the general logic. Also if there is something too ungroovy let me know.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy
```diff
@@ -317,6 +322,41 @@ class XcodeTask extends DefaultTask {
     }
 
     /**
+     * Replaces lines between two regexps
+     */
+    @VisibleForTesting
+    static List<String> regexReplaceLines(List<String> podfileLines,
+                                        String startRegex, String endRegex, List<String> newPodFileLines) {
```
brunobowden: This should be a variation of the regexInsertLines method. Just extend regexInsertLines, to add a `boolean replace` parameter. If true, it should skip copying the old lines. It avoid having duplication of similar code.

brunobowden:
@saruye - I took elements of your code and then reworked it for PR #582. With that in mind, I'm closing this PR but I'd still appreciate your input on the other one. Thank you for helping out here.

PR #573, opened 2015-12-16
FAQ for "Undefined symbols for architecture"

brunobowden:
- Commonly occurs due to Cocoapods flakiness

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -43,6 +43,7 @@ Paste the results below, replacing existing contents.
 - [Cycle Finder Advanced Setup](#cycle-finder-advanced-setup)
 - [How do I develop on Windows or Linux?](#how-do-i-develop-on-windows-or-linux)
 - [How do I fix missing required architecture linker warning?](#how-do-i-fix-missing-required-architecture-linker-warning)
+- [How do I fix Undefined symbols for architecture linker warning?](#how-do-i-fix-undefined-symbols-for-architecture-linker-warning)
 - [How do I solve the Eclipse error message Obtaining Gradle model...?](#how-do-i-solve-the-eclipse-error-message-obtaining-gradle-model)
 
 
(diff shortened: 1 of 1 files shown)


FAQ.md
```diff
@@ -561,6 +562,44 @@ j2objcConfig {
...
+### How do I fix `Undefined symbols for architecture` linker warning?
+
+This usually occurs with `x86_64` for the simulator and `arm64` when running
+on a device. The error commonly looks like this:
+
+Undefined symbols for architecture x86_64:
+  "_OBJC_METACLASS_$_MyClassMethodName", referenced from:
+     _OBJC_METACLASS_$_MethodName in MyClass.o
+
+It can usually be fixed by manually running the Cocoapods install command
```
advayDev1: Markdown syntax highlighter thinks there are mismatched underscores here so the whole rest of the document is italics.

Either the x86_64 needs to be in quotes, or you can do what I did in a different source file, add a blank __-__ in a  which seems to reset it

advayDev1:
lgtm with comment

devisnik:
:+1: 
I did a `pod update` to fix the issue, which does a little less if I understand it correctly.
https://github.com/CocoaPods/CocoaPods/issues/760#issuecomment-46300500

brunobowden:
It could well be that "pod update" fixes the issue.... but it's better for
reliability to be consistent. When people are doing "pod update" and "pod
install", then you have two paths to investigate and understand. If people
do one path and one path only, you can have far greater confidence that it
works.

FAQ.md
brunobowden: Fixed by quoting

tjerkw:
I dont use Cocoa Pods, and i still have this problem:

`
Undefined symbols for architecture x86_64:
  "_OBJC_CLASS_$_TWalletDataWalletOrder", referenced from:
      objc-class-ref in AppDelegate.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
`

However the class is defined, the header file and .m files are included... So I still dont understand this problem.

Issue #572, opened 2015-12-16
Argument list too long


himamis:
I have a huge project with over 3000 files in obscured long package names.
Because of this, TranslateTask fails, with error 'Argument list too long'.
As far as I understand this is a limitation of the Mac os of having ARG_MAX as 262144 (`getconf ARG_MAX` or `sysctl -A | grep kern.argmax`).
Is there a solution to this problem? (Maybe translating in multiple steps, or giving jar with sources as argument)

build.grade file
```groovy
buildscript {
	repositories {
		maven {
			url 'https://plugins.gradle.org/m2/'
		}
	}
	dependencies {
		classpath 'gradle.plugin.com.github.j2objccontrib.j2objcgradle:j2objc-gradle:0.5.0-alpha'
	}
}

configure([project(':common'), project(':ggbjdk')]) {
	// inject j2objc gradle plugin into common projects
	apply plugin: 'com.github.j2objccontrib.j2objcgradle'

	j2objcConfig {
		autoConfigureDeps true

		xcodeProjectDir '../ios/JavaClasses'

		finalConfigure()
	}
}

task wrapper(type: Wrapper) {
	gradleVersion = '2.8'
}
```

Nothing is printed as error, but if I take the command that is called and copy paste it into terminal, I get the `Argument list too long` error.

j2objc version 0.9.8.2.1

brunobowden:
Hi @hinamis,
Thanks for doing your research before posting this issue. The plugin will normally pass through the error but I can imagine it has difficulties in this particular case. I think there's two possible workarounds and also likely a way of doing a proper fix in the plugin. Given that I like neither of the workarounds, it would be very helpful to get a PR from you to do the fix in the plugin. If my guesses are correct, then the fix shouldn't be too difficult. I'd also appreciate you doing the fix as you'll be able to test it properly as well.

So here are the workarounds followed by the potential proper plugin fix:

1) Reduce the average filepath:
Given the 262,144 limit (I presume this is a character limit?) and 3,000 files, the obvious workaround is keep the average filepath to less than 85 characters (262,144 / 3,000 = 87). If you're root directly is `/Users/SomethingLongThatCanBeMadeSmaller`, then that's easy. Otherwise you could replace the package names throughout the whole project but that's a big pain.

2) Split in to multiple projects:
[...]

PR #574, opened 2016-01-04
Arbitrary arg length

himamis:
This pull request fixes issue #572 .
A long term fix has been implemented, using a temporary file.
The test have also been modified to accommodate for the changes in the command line of the TranslateTask.

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTask.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTask.groovy
@@ -252,7 +252,7 @@ class TranslateTask extends DefaultTask {
                 project.files(getTranslateSourcepaths()),
                 project.files(getGeneratedSourceDirs())
         ])
-        doTranslate(sourcepathDirs, srcGenMainDir, translateArgs, mainSrcFilesChanged)
+        doTranslate(sourcepathDirs, srcGenMainDir, translateArgs, mainSrcFilesChanged, "mainSrcFilesArgFile")
 
         // Translate test code. Tests are never built with --build-closure; otherwise
         // we will get duplicate symbol errors.
@@ -269,7 +269,7 @@ class TranslateTask extends DefaultTask {
                 project.files(getTranslateSourcepaths()),
                 project.files(getGeneratedSourceDirs())
         ])
-        doTranslate(sourcepathDirs, srcGenTestDir, testTranslateArgs, testSrcFilesChanged)
+        doTranslate(sourcepathDirs, srcGenTestDir, testTranslateArgs, testSrcFilesChanged, "testSrcFilesArgFile")
     }
 
     int deleteRemovedFiles(List<String> removedFileNames, File dir) {
@@ -292,7 +292,7 @@ class TranslateTask extends DefaultTask {
     }
 
     void doTranslate(UnionFileCollection sourcepathDirs, File srcDir, List<String> translateArgs,
-                     FileCollection srcFilesToTranslate) {
+                     FileCollection srcFilesToTranslate, String srcFilesArgFilename) {
         int num = srcFilesToTranslate.getFiles().size()
         logger.info("Translating $num files with j2objc...")
         if (srcFilesToTranslate.getFiles().size() == 0) {
@@ -318,6 +318,27 @@ class TranslateTask extends DefaultTask {
         String classpathArg = Utils.joinedPathArg(classpathFiles) +
                               Utils.pathSeparator() + "${project.buildDir}/classes"
 
+        // Source files arguments
+        List<String> srcFilesArgs = []
+        int srcFilesArgsCharCount = 0
+        for (File file in srcFilesToTranslate) {
+            String filePath = file.getPath()
+            srcFilesArgs.add(filePath)
+            srcFilesArgsCharCount += filePath.length() + 1
+        }
+
+        // Handle command line that's too long
+        // Allow up to 2,000 characters for command line excluding src files
+        // http://docs.oracle.com/javase/7/docs/technotes/tools/windows/javac.html#commandlineargfile
+        if (srcFilesArgsCharCount + 2000 > Utils.maxArgs()) {
+            File srcFilesArgFile = new File(getTemporaryDir(), srcFilesArgFilename);
+            FileWriter writer = new FileWriter(srcFilesArgFile);
+            writer.append(srcFilesArgs.join('\n'));
+            writer.close()
+            // Replace src file arguments by referencing file
+            srcFilesArgs = ["@${srcFilesArgFile.path}".toString()]
+        }
+
         ByteArrayOutputStream stdout = new ByteArrayOutputStream()
         ByteArrayOutputStream stderr = new ByteArrayOutputStream()
 
@@ -338,8 +359,9 @@ class TranslateTask extends DefaultTask {
                 }
 
                 // File Inputs
-                srcFilesToTranslate.each { File file ->
-                    args file.path
+                srcFilesArgs.each { String arg ->
+                    // Can be list of src files or a single @/../srcFilesArgFile reference
+                    args arg
                 }
 
                 setStandardOutput stdout
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
@@ -695,4 +695,25 @@ class Utils {
         }
         return files
     }
+
+
+    // Max number of characters for OS command line
+    static int maxArgs() {
+        if (isMacOSX()) {
+            // http://www.in-ulm.de/~mascheck/various/argmax/
+            return 262144
+        }
+        if (isWindows()) {
+            // Assume Windows XP or later which has max of 8191
+            // https://support.microsoft.com/en-us/kb/830473
+            return 8191
+        }
+        if (isLinux()) {
+            // v2.6.23 (released 2007) or later limit is 1/4 of stack size,
+            // so Linux is presumed to have no limit
+            // http://www.in-ulm.de/~mascheck/various/argmax/
+            return Integer.MAX_VALUE
+        }
+        assert false
+    }
 }
--- a/src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTaskTest.groovy
+++ b/src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTaskTest.groovy
@@ -38,10 +38,9 @@ class TranslateTaskTest {
         (proj, j2objcHome, j2objcConfig) = TestingUtils.setupProject(
                 new TestingUtils.ProjectConfig(applyJavaPlugin: true, createJ2objcConfig: true))
 
+        // The files in these folders are created by the 'genNonIncrementalInputs' method
         proj.file('src/main/java/com/example').mkdirs()
-        proj.file('src/main/java/com/example/Main.java').write("// Fake")
         proj.file('src/test/java/com/example').mkdirs()
-        proj.file('src/test/java/com/example/Verify.java').write("// Fake")
     }
 
     // TODO: add java source files to the test cases
@@ -61,7 +60,7 @@ class TranslateTaskTest {
                 '-d', '/PROJECT_DIR/build/j2objcSrcGenMain',
                 '-sourcepath', '/PROJECT_DIR/src/main/java:/PROJECT_DIR/build/classes/main',
                 '-classpath', '/J2OBJC_HOME/lib/j2objc_annotations.jar:/J2OBJC_HOME/lib/j2objc_guava.jar:/J2OBJC_HOME/lib/j2objc_junit.jar:/J2OBJC_HOME/lib/jre_emul.jar:/J2OBJC_HOME/lib/javax.inject-1.jar:/J2OBJC_HOME/lib/jsr305-3.0.0.jar:/J2OBJC_HOME/lib/mockito-core-1.9.5.jar:/J2OBJC_HOME/lib/hamcrest-core-1.3.jar:/J2OBJC_HOME/lib/protobuf_runtime.jar:/PROJECT_DIR/build/classes',
-                '/PROJECT_DIR/src/main/java/com/example/Main.java'
+                '/PROJECT_DIR/src/main/java/com/example/Main1.java'
         ],
         // expectedWindowsExecutableAndArgs
         [
@@ -75,7 +74,7 @@ class TranslateTaskTest {
                 '-d', '/PROJECT_DIR/build/j2objcSrcGenTest',
                 '-sourcepath', '/PROJECT_DIR/src/main/java:/PROJECT_DIR/src/test/java:/PROJECT_DIR/build/classes/main',
                 '-classpath', '/J2OBJC_HOME/lib/j2objc_annotations.jar:/J2OBJC_HOME/lib/j2objc_guava.jar:/J2OBJC_HOME/lib/j2objc_junit.jar:/J2OBJC_HOME/lib/jre_emul.jar:/J2OBJC_HOME/lib/javax.inject-1.jar:/J2OBJC_HOME/lib/jsr305-3.0.0.jar:/J2OBJC_HOME/lib/mockito-core-1.9.5.jar:/J2OBJC_HOME/lib/hamcrest-core-1.3.jar:/J2OBJC_HOME/lib/protobuf_runtime.jar:/PROJECT_DIR/build/classes',
-                '/PROJECT_DIR/src/test/java/com/example/Verify.java'
+                '/PROJECT_DIR/src/test/java/com/example/Verify1.java'
         ],
         // expectedWindowsExecutableAndArgs
         [
@@ -104,7 +103,7 @@ class TranslateTaskTest {
                 '-d', '/PROJECT_DIR/build/j2objcSrcGenMain',
                 '-sourcepath', '/PROJECT_DIR/src/main/java;/PROJECT_DIR/build/classes/main',
                 '-classpath', '/J2OBJC_HOME/lib/j2objc_annotations.jar;/J2OBJC_HOME/lib/j2objc_guava.jar;/J2OBJC_HOME/lib/j2objc_junit.jar;/J2OBJC_HOME/lib/jre_emul.jar;/J2OBJC_HOME/lib/javax.inject-1.jar;/J2OBJC_HOME/lib/jsr305-3.0.0.jar;/J2OBJC_HOME/lib/mockito-core-1.9.5.jar;/J2OBJC_HOME/lib/hamcrest-core-1.3.jar;/J2OBJC_HOME/lib/protobuf_runtime.jar;/PROJECT_DIR/build/classes',
-                '/PROJECT_DIR/src/main/java/com/example/Main.java'
+                '/PROJECT_DIR/src/main/java/com/example/Main1.java'
         ],
         // expectedWindowsExecutableAndArgs
         [
@@ -118,7 +117,7 @@ class TranslateTaskTest {
                 '-d', '/PROJECT_DIR/build/j2objcSrcGenTest',
                 '-sourcepath', '/PROJECT_DIR/src/main/java;/PROJECT_DIR/src/test/java;/PROJECT_DIR/build/classes/main',
                 '-classpath', '/J2OBJC_HOME/lib/j2objc_annotations.jar;/J2OBJC_HOME/lib/j2objc_guava.jar;/J2OBJC_HOME/lib/j2objc_junit.jar;/J2OBJC_HOME/lib/jre_emul.jar;/J2OBJC_HOME/lib/javax.inject-1.jar;/J2OBJC_HOME/lib/jsr305-3.0.0.jar;/J2OBJC_HOME/lib/mockito-core-1.9.5.jar;/J2OBJC_HOME/lib/hamcrest-core-1.3.jar;/J2OBJC_HOME/lib/protobuf_runtime.jar;/PROJECT_DIR/build/classes',
-                '/PROJECT_DIR/src/test/java/com/example/Verify.java'
+                '/PROJECT_DIR/src/test/java/com/example/Verify1.java'
         ],
         // expectedWindowsExecutableAndArgs
         [
@@ -160,7 +159,7 @@ class TranslateTaskTest {
                 '-sourcepath', "/PROJECT_DIR/src/main/java:/PROJECT_DIR/REL-SOURCEPATH:$absSourcePath:/PROJECT_DIR/build/classes/main:/PROJECT_DIR/REL-GENPATH:$absGenPath",
                 '-classpath', "/PROJECT_DIR/REL-CLASSPATH:$absClassPath:/J2OBJC_HOME/lib/J2OBJC-LIB1:/J2OBJC_HOME/lib/J2OBJC-LIB2:/PROJECT_DIR/build/classes",
                 '-ARG1', '-ARG2',
-                '/PROJECT_DIR/src/main/java/com/example/Main.java'
+                '/PROJECT_DIR/src/main/java/com/example/Main1.java'
         ],
         // expectedWindowsExecutableAndArgs
         [
@@ -174,7 +173,7 @@ class TranslateTaskTest {
                 '-sourcepath', "/PROJECT_DIR/src/main/java:/PROJECT_DIR/src/test/java:/PROJECT_DIR/REL-SOURCEPATH:$absSourcePath:/PROJECT_DIR/build/classes/main:/PROJECT_DIR/REL-GENPATH:$absGenPath",
                 '-classpath', "/PROJECT_DIR/REL-CLASSPATH:$absClassPath:/J2OBJC_HOME/lib/J2OBJC-LIB1:/J2OBJC_HOME/lib/J2OBJC-LIB2:/PROJECT_DIR/build/classes",
                 '-ARG1', '-ARG2',
-                '/PROJECT_DIR/src/test/java/com/example/Verify.java'
+                '/PROJECT_DIR/src/test/java/com/example/Verify1.java'
         ],
         // expectedWindowsExecutableAndArgs
         [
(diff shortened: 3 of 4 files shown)


src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTaskTest.groovy
```diff
@@ -61,7 +61,7 @@ class TranslateTaskTest {
                 '-d', '/PROJECT_DIR/build/j2objcSrcGenMain',
                 '-sourcepath', '/PROJECT_DIR/src/main/java:/PROJECT_DIR/build/classes/main',
                 '-classpath', '/J2OBJC_HOME/lib/j2objc_annotations.jar:/J2OBJC_HOME/lib/j2objc_guava.jar:/J2OBJC_HOME/lib/j2objc_junit.jar:/J2OBJC_HOME/lib/jre_emul.jar:/J2OBJC_HOME/lib/javax.inject-1.jar:/J2OBJC_HOME/lib/jsr305-3.0.0.jar:/J2OBJC_HOME/lib/mockito-core-1.9.5.jar:/J2OBJC_HOME/lib/hamcrest-core-1.3.jar:/J2OBJC_HOME/lib/protobuf_runtime.jar:/PROJECT_DIR/build/classes',
-                '/PROJECT_DIR/src/main/java/com/example/Main.java'
+                '@/PROJECT_DIR/build/tmp/j2objcTranslate/inputFiles'
```
brunobowden: Do you see how this loses something in the validation? Previously, it would check for `Main.java` but now that no-longer happens. Be very careful that you're not regressing the test. Please make two changes:

1) For short command lines, no temporary file should be used - so revert this test to what it was before. That preserves the existing behaviour and avoid the complexity and risk of the temporary file.

2) Add a specific test for the long command line. That'll need a normal test as usual with the `@/PROJECT_DIR/build/tmp/j2objcTranslate/inputFiles`, then an additional test that reads `inputFiles` and verifies the contents.

himamis: I'll make the changes. I also have a question:

1) What do you think, what should consist of a long command line? I thought having a an argument list with a length greater than 250000 is what should be called long (as noted before, for Mac OS X this is 262144). We could also tighten this number and make it around 100000.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTask.groovy
```diff
@@ -318,6 +318,14 @@ class TranslateTask extends DefaultTask {
         String classpathArg = Utils.joinedPathArg(classpathFiles) +
                               Utils.pathSeparator() + "${project.buildDir}/classes"
 
+        File inputFiles = new File(getTemporaryDir(), "inputFiles");
+        FileWriter writer = new FileWriter(inputFiles);
+        for (File file in srcFilesToTranslate) {
+            writer.append(file.getPath());
+            writer.append("\n");
+        }
+        writer.close();
```
brunobowden: The temporary file should only be used if it's necessary:

    // Source files arguments
    List<String> srcFilesArgs = []
    srcFilesArgsCharCount = 0
    for (File file in srcFilesToTranslate) {
        String filePath = file.getPath()
        srcFilesArgs.append(filePath)
        srcFilesArgsCharCount += filePath.count() + 1
    }

    // Handle command line that's too long
    // Allow up to 2,000 characters for command line excluding src files
    if (srcFilesArgsCharCount + 2000 > Utils.maxArgs()) {
        File srcFilesArgsFile = new File(getTemporaryDir(), "srcFilesArgsFile");
        FileWriter writer = new FileWriter(srcFilesArgsFile);
            writer.append(srcFilesArgs.join('\n'));
        }
    }

Add the following to Utils.groovy:

   // Max number of character for OS command line
   static int maxArgs() {
        if (Utils.isMacOSX()) {
            // http://www.in-ulm.de/~mascheck/various/argmax/
            return 262144
        }
        if (Utils.isWindows()) {
            // need to add good reference for this
            return 8192
        }
        if (Utils.isLinux()) {
            // v2.6.23 (released 2007) or later limit is 1/4 of stack size,
            // so Linux is presumed to have no limit
            // http://www.in-ulm.de/~mascheck/various/argmax/
            return Integer.MAX_VALUE
        }
        assert false
    }

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/UtilsTest.groovy
```diff
@@ -842,4 +844,18 @@ class UtilsTest {
...
+        assert Utils.isLongArgument("short argument")
+        Utils.resetLongArgumentLength()
+        assert !Utils.isLongArgument("still short argument")
+        StringBuilder longString = new StringBuilder("very long argument");
+        while (longString.length() < 250000) {
+            longString.append((CharSequence) longString)
+        }
+        assert Utils.isLongArgument(longString.toString())
+    }
 }
```
brunobowden: Tests should be os specific:

    @Test
    testMaxArgs_Linux() {
        Utils.setFakeOSMacOSX()
        assert Integer.MAX_VALUE == Utils.maxArgs()
    }

    @Test
    testMaxArgs_OSX() {
        Utils.setFakeOSMacOSX()
        assert 262144 == Utils.maxArgs()
    }

    @Test
    testMaxArgs_Windows() {
        Utils.setFakeOSWindows()
        assert 8192 == Utils.maxArgs()
    }

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTask.groovy
```diff
@@ -318,6 +318,21 @@ class TranslateTask extends DefaultTask {
...
+            File inputFiles = new File(getTemporaryDir(), "inputFiles")
+            FileWriter writer = new FileWriter(inputFiles)
+            for (File file in srcFilesToTranslate) {
+                writer.append(file.path)
+                writer.append("\n")
+            }
+            writer.close()
+            inputFilesPath = inputFiles.path
+        }
+
```
brunobowden: The temporary file should only be used if it's necessary. Here's an updated version:

    // Source files arguments
    List<String> srcFilesArgs = []
    srcFilesArgsCharCount = 0
    for (File file in srcFilesToTranslate) {
        String filePath = file.getPath()
        srcFilesArgs.append(filePath)
        srcFilesArgsCharCount += filePath.count() + 1
    }

    // Handle command line that's too long
    // Allow up to 2,000 characters for command line excluding src files
    if (srcFilesArgsCharCount + 2000 > Utils.maxArgs()) {
        File srcFilesArgsFile = new File(getTemporaryDir(), "srcFilesArgsFile");
        FileWriter writer = new FileWriter(srcFilesArgsFile);
            writer.append(srcFilesArgs.join('\n'));
        }
        // Replace src file arguments by referencing file
        srcFilesArgsFile = ["@${srcFilesArgsFile.getPath()}"]
    }

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTask.groovy
```diff
@@ -338,8 +353,12 @@ class TranslateTask extends DefaultTask {
                 }
 
                 // File Inputs
-                srcFilesToTranslate.each { File file ->
-                    args file.path
+                if (argumentIsLong) {
+                    args "@${inputFilesPath}"
```
brunobowden: With the new version, it doesn't need any if check:

    // File Inputs
    srcFilesArgs.each { String arg ->
        // Can be list of src files or a single @/../srcFilesArgsFile reference
        args arg
    }

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
```diff
@@ -695,4 +695,36 @@ class Utils {
...
+
+
+    private static boolean alwaysLongArgumentLength = false
+
+    /* This allows simpler testing of long arguments.
+     *
+     * The resetLongArgumentLength() should be called in the test unit @After method to isolate test state.
+     */
+    @VisibleForTesting
+    static void setAlwaysLongArgumentLength() {
```
brunobowden: Try to limit modification of functionality as this is harder to test. Better to supply a legitimately long list of arguments and test that rather than make it pretend that.

There is a counter example in this file with the setFakeOS... methods but there is no other way around that. This is necessary to make the tests self-contained when run on different OSs. They're also carefully written so that running tests on one OS, test all the other OSs at the same time. So it's difficult for a build to pass locally and then fail on another OS.

brunobowden: In this case, I don't think any modification of functionality is needed. Just use the setFakeOS to pretend it's on a system with a limit (incase you're developing on Linux which practically has no limit), then create a real command line of sufficient length and test it for real.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTaskTest.groovy
```diff
@@ -189,6 +190,53 @@ class TranslateTaskTest {
         mockProjectExec.verify()
     }
 
+    @Test
+    void translate_LongArgumentList() {
```
brunobowden: Rename to `translate_maxArgsExceeded` as that more precisely describes the situation.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTaskTest.groovy
```diff
@@ -189,6 +190,53 @@ class TranslateTaskTest {
...
+        Utils.setAlwaysLongArgumentLength()
```
brunobowden: Instead use the fake os method to artificially test OS X. Though you could test Windows, I think it's better to test OS X as this is the most common os for development.

    // Artificially enforce max arg limit on most common OS for development
    Utils.setFakeOSMacOSX()

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTaskTest.groovy
```diff
@@ -189,6 +190,53 @@ class TranslateTaskTest {
...
+        TranslateTask j2objcTranslate = (TranslateTask) proj.tasks.create(
+                name: 'j2objcTranslate', type: TranslateTask) {
+            srcGenMainDir = proj.file(proj.file('build/j2objcSrcGenMain').absolutePath)
+            srcGenTestDir = proj.file(proj.file('build/j2objcSrcGenTest').absolutePath)
```
brunobowden: I'm thinking about how to fake a larger number of files. To do this properly will be a little trickier, let me think about this for the moment...

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTaskTest.groovy
```diff
@@ -189,6 +190,53 @@ class TranslateTaskTest {
...
+        assert proj.file("/build/tmp/j2objcTranslate/inputFiles").text.endsWith(
+                new File("src/test/java/com/example/Verify.java").path + "\n"
+        )
+
+        mockProjectExec.verify()
+    }
+
 
     // Utility Method
     private static IncrementalTaskInputs genNonIncrementalInputs(Project proj) {
```
brunobowden: So here's the trick. Modify this function to generate a larger number of files. This defaults to creating a single Main.java and Verify.java. However if fileCountMultiplier = 100, then it creates 100 copies of each file, all with sequential numerical suffix. This requires modifying the other tests from `Main.java` to `Main1.java` but that's easy enough.

    private static IncrementalTaskInputs genNonIncrementalInputs(Project proj, int fileCountMultiplier = 1) {
       ...
          @Override
          void outOfDate(Action<? super InputFileDetails> outOfDateAction) {
              List<String> srcFiles = []
              (1..fileCountMultiplier).each { int idx =>
                  srcFiles.append("src/main/java/com/example/Main${idx}.java")
                  srcFiles.append("src/test/java/com/example/Verify${idx}.java")
              }
              proj.files(srcFiles).each {
              ...

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTaskTest.groovy
```diff
@@ -189,6 +190,53 @@ class TranslateTaskTest {
...
+                '@/PROJECT_DIR/build/tmp/j2objcTranslate/inputFiles'
+        ],
+        // expectedWindowsExecutableAndArgs
+        [
+                'java',
+                '-jar',
+                '/J2OBJC_HOME/lib/j2objc.jar'
+        ])
+
+        j2objcTranslate.translate(genNonIncrementalInputs(proj))
```
brunobowden: Modify to translate a large number of files in order to trigger the limit. It's tricky to know how many files are needed as you don't know the length of the path to the temporary directory. So you have to overshoot:

    // Translate large number of files to exceed max args
    // srcFilesArgsCharCount = 40 (min chars for file path assumption) * 2 (main & test) * 3,500
    //                                       = 280,000  (this is greater than 262144, which is max args on OS X)
    j2objcTranslate.translate(genNonIncrementalInputs(proj, 3500))

    // Verify arg file
    List<String> srcFiles = proj.file("/build/tmp/j2objcTranslate/inputFiles").readLines()
    assert srcFiles.count() == 7000
    assert srcFiles[0] == proj.file('src/test/java/com/example/Main0.java').path
    assert srcFiles[1] == proj.file('src/test/java/com/example/Main1.java').path
    assert srcFiles[3499] == proj.file('src/test/java/com/example/Main3500.java').path
    assert srcFiles[3500] == proj.file('src/test/java/com/example/Verify0.java').path
    assert srcFiles[6999] == proj.file('src/test/java/com/example/Verify3500.java').path

I think that's enough verification to be confident that this works.

brunobowden:
@himamis - this should be enough feedback for you to fix this up. Thanks for working on this. Ping me when you've worked through this. There's a couple of other things for you to do:

1. Collapse the commits to a single commit - the contributor guide has notes on this
1. Make sure it passes on all the CI builds on each different platform
1. Add a note in the comment indicating how long the test takes to run (you can see it in the test output). If it's over a second then we should understand that and see if we can make it faster. I believe it should be pretty quick though.

himamis:
@brunobowden please let me know if the changes are passable, then I'll prepare the pull request for submission.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTaskTest.groovy
```diff
@@ -189,18 +187,87 @@ class TranslateTaskTest {
...
-                           'src/test/java/com/example/Verify.java').each {
+                List<String> srcFiles = []
+                (1..fileCountMultiplier).each { int idx ->
+                    srcFiles.add("src/main/java/com/example/Main${idx}.java")
+                    srcFiles.add("src/test/java/com/example/Verify${idx}.java")
+                }
+                proj.files(srcFiles).each {
                     final File file ->
+                    // Create the files
+                    file.write("// Fake")
```
brunobowden: Take this line out.

The file never needs to be written to disk in order for the test to work. In fact, writing a large number of files will slow down the test dramatically without adding anything to the quality of the testing. An important part of unit tests is that they should avoid anything like writing or reading to a file. In this case, just passing a list of "changed files" is enough for this test to verify the translate task.

brunobowden: Actually, I see the code that you moved here that was writing the file in another place. Can you check if it needs to write the file at all? I will test this myself also.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
```diff
@@ -695,4 +695,26 @@ class Utils {
...
+    // Max number of characters for OS command line
+    static int maxArgs() {
+        if (isMacOSX()) {
+            // http://www.in-ulm.de/~mascheck/various/argmax/
+            return 262144
+        }
+        if (isWindows()) {
+            // https://support.microsoft.com/en-us/kb/830473
+            // this value applies to Windows XP and above
+            // for Windows 2000 or NT 4.0 it is 2047
```
brunobowden: Put the assumption first, then references afterwards:

    // Assume Windows XP or later which has max of 8191
    // https://support.microsoft.com/en-us/kb/830473

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTaskTest.groovy
```diff
@@ -189,18 +187,87 @@ class TranslateTaskTest {
         mockProjectExec.verify()
     }
 
+    @Test
+    // This test takes 8 seconds on average on a local dev computer
+    void translate_MaxArgsExceeded() {
+        // Artificially enforce max arg limit on most common OS for development
+        Utils.setFakeOSMacOSX()
```
brunobowden: Let's us Windows instead as that'll force a much faster test:

    // Windows is used as max args is low, which makes for a faster test
    Utils.setFakeOSWindows()

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTaskTest.groovy
```diff
@@ -189,18 +187,87 @@ class TranslateTaskTest {
...
+        // Translate large number of files to exceed max args
+        // srcFilesArgsCharCount = 40 (min chars for file path assumption) * 2 (main & test) * 3,500
+        //                                       = 280,000  (this is greater than 262144, which is max args on OS X)
```
brunobowden: Align the `=` sign on both lines.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTaskTest.groovy
```diff
@@ -189,18 +187,87 @@ class TranslateTaskTest {
...
+        int fileCount = 3500
```
brunobowden: `fileCount = 120` as that'll work for Windows. Update numbers on previous lines.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTaskTest.groovy
```diff
@@ -189,18 +187,87 @@ class TranslateTaskTest {
...
+
+        MockProjectExec mockProjectExec = new MockProjectExec(proj, j2objcHome)
+        mockProjectExec.demandExecAndReturn([
+                '/J2OBJC_HOME/j2objc',
+                '-d', '/PROJECT_DIR/build/j2objcSrcGenMain',
+                '-sourcepath', '/PROJECT_DIR/src/main/java:/PROJECT_DIR/build/classes/main',
+                '-classpath', '/J2OBJC_HOME/lib/j2objc_annotations.jar:/J2OBJC_HOME/lib/j2objc_guava.jar:/J2OBJC_HOME/lib/j2objc_junit.jar:/J2OBJC_HOME/lib/jre_emul.jar:/J2OBJC_HOME/lib/javax.inject-1.jar:/J2OBJC_HOME/lib/jsr305-3.0.0.jar:/J2OBJC_HOME/lib/mockito-core-1.9.5.jar:/J2OBJC_HOME/lib/hamcrest-core-1.3.jar:/J2OBJC_HOME/lib/protobuf_runtime.jar:/PROJECT_DIR/build/classes',
+                '@/PROJECT_DIR/build/tmp/j2objcTranslate/mainSrcFilesArgsFile'
+        ],
+                // expectedWindowsExecutableAndArgs
```
brunobowden: Align this comment with the square brackets.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTaskTest.groovy
```diff
@@ -189,18 +187,87 @@ class TranslateTaskTest {
...
-    private static IncrementalTaskInputs genNonIncrementalInputs(Project proj) {
+    private static IncrementalTaskInputs genNonIncrementalInputs(Project proj, int fileCountMultiplier = 1) {
         IncrementalTaskInputs incrementalTaskInputs = new IncrementalTaskInputs() {
             @Override
             boolean isIncremental() { return false }
 
             @Override
             void outOfDate(Action<? super InputFileDetails> outOfDateAction) {
-                proj.files('src/main/java/com/example/Main.java',
-                           'src/test/java/com/example/Verify.java').each {
```
brunobowden: Add a simple assert:

    assert fileCountMultiplier >= 1

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTaskTest.groovy
```diff
@@ -189,18 +187,87 @@ class TranslateTaskTest {
...
+                List<String> srcFiles = []
+                (1..fileCountMultiplier).each { int idx ->
+                    srcFiles.add("src/main/java/com/example/Main${idx}.java")
+                    srcFiles.add("src/test/java/com/example/Verify${idx}.java")
+                }
```
brunobowden: As mentioned just below, try seeing if it works when you just write one file instead of all of them. If this does work, then keep the fakeOS as Mac (you don't need to change it to Windows):

    List<String> srcFiles = []
    (1..fileCountMultiplier).each { int idx ->
        srcFiles.add("src/main/java/com/example/Main${idx}.java")
        srcFiles.add("src/test/java/com/example/Verify${idx}.java")
    }
    // Write first main and test files, so there's something to delete, which the test needs
    proj.file(srcFiles[0]).write("// Fake")
    proj.file(srcFiles[1]).write("// Fake")

If this doesn't work, then just go ahead with the switch to Windows to make it faster.

himamis:
On Windows the test fails, if the files are not there:
```
junit.framework.AssertionFailedError: No call to 'delete' expected at this point. Still 1 call(s) to 'exec' expected.
```
Good news is that with Windows 240 files are enough, and with that the test is under 1 second on my machine.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTaskTest.groovy
```diff
@@ -189,18 +187,87 @@ class TranslateTaskTest {
         mockProjectExec.verify()
     }
 
+    @Test
+    // This test takes 8 seconds on average on a local dev computer
```
brunobowden: Remove this comment

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTaskTest.groovy
```diff
@@ -189,18 +187,87 @@ class TranslateTaskTest {
...
+    void translate_MaxArgsExceeded() {
+        // Artificially enforce max arg limit on most common OS for development
```
brunobowden: Update this comment explaining why you used Windows and the faster speed, e.g. "(test takes <1 second compared to 8 seconds on OS X)"

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTaskTest.groovy
```diff
@@ -189,18 +187,87 @@ class TranslateTaskTest {
...
+        Utils.setFakeOSWindows()
+        // Translate large number of files to exceed max args
+        // srcFilesArgsCharCount = 40 (min chars for file path assumption) * 300
```
brunobowden: What about the factor of 2 for main vs. test files? The line should be:

    // srcFilesArgsCharCount = 40 (min chars for file path assumption) * 2 (main and test file) * 110
    //                       = 8,800  (this is greater than 8,191, which is max args on Windows XP and above)

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTaskTest.groovy
```diff
@@ -189,18 +187,87 @@ class TranslateTaskTest {
...
+        //                       = 12,000  (this is greater than 8,191, which is max args on Windows XP and above)
+        int fileCount = 240
```
brunobowden: This line should match the choice you made on line 196. In this case, I think the 110 should work.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTask.groovy
```diff
@@ -318,6 +318,26 @@ class TranslateTask extends DefaultTask {
...
+        }
+
+        // Handle command line that's too long
+        // Allow up to 2,000 characters for command line excluding src files
+        if (srcFilesArgsCharCount + 2000 > Utils.maxArgs()) {
+            File srcFilesArgsFile = new File(getTemporaryDir(), srcFilesArgsFilename);
+            FileWriter writer = new FileWriter(srcFilesArgsFile);
+            writer.append(srcFilesArgs.join('\n'));
+            writer.close()
+            // Replace src file arguments by referencing file
```
brunobowden: Add extra line referencing Java documentation:

    // http://docs.oracle.com/javase/7/docs/technotes/tools/windows/javac.html#commandlineargfile

brunobowden:
@himamis - looks like the PR is close to complete. I see the same issue as you that it fails without any files. I put in a comment about trying it with just a main and test file. See if that works.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTaskTest.groovy
```diff
@@ -39,9 +39,7 @@ class TranslateTaskTest {
                 new TestingUtils.ProjectConfig(applyJavaPlugin: true, createJ2objcConfig: true))
 
         proj.file('src/main/java/com/example').mkdirs()
-        proj.file('src/main/java/com/example/Main.java').write("// Fake")
         proj.file('src/test/java/com/example').mkdirs()
-        proj.file('src/test/java/com/example/Verify.java').write("// Fake")
```
brunobowden: Add comment:

    // File in this folder are created by 'genNonIncrementalInputs' method

himamis:
Sorry, my comment was not clear enough. Without any files, all of the translate tests fail, with just Main1 and Verify1 only the translate_MaxArgsExceeded test fails.

brunobowden:
Fair enough. In which case, leave it as `setFakeOSWindows` and creating all
the files. Then add a comment, in the genNonIncrementalInputs method that:

    // TODO: prefer not to create all these temporary files just to run the
test

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTaskTest.groovy
himamis: The main and the test sources are translated separately, so we need to have enough files in 
at least one of the source sets, to trigger the usage of the file args for that source set.
For speeding up the test, we could only generate files for the main or test source set and then check if the args file was used.

brunobowden: Ahh.... fair point. In that case, you're good. Last two issues I can see
are:

1) Adding this comment as mentioned before:

    //
http://docs.oracle.com/javase/7/docs/technotes/tools/windows/javac.html#commandlineargfile

2) Collapse the commits:

https://github.com/j2objc-contrib/j2objc-gradle/blob/master/CONTRIBUTING.md#preparing-your-pull-request-for-submission

himamis: The comment has been added. I'll collapse my commits into a single one now.

himamis:
@brunobowden The commits have been squashed into a single commit. Please let me know, if there is anything else I need to do.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTask.groovy
```diff
@@ -252,7 +252,7 @@ class TranslateTask extends DefaultTask {
                 project.files(getTranslateSourcepaths()),
                 project.files(getGeneratedSourceDirs())
         ])
-        doTranslate(sourcepathDirs, srcGenMainDir, translateArgs, mainSrcFilesChanged)
+        doTranslate(sourcepathDirs, srcGenMainDir, translateArgs, mainSrcFilesChanged, "mainSrcFilesArgsFile")
```
brunobowden: Replace "argsFile" with "argFile" to match the naming in the Java documentation. There's 13 instances of this.

Apologies I didn't catch this earlier:
http://docs.oracle.com/javase/7/docs/technotes/tools/windows/javac.html#commandlineargfile

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTaskTest.groovy
```diff
@@ -189,18 +188,88 @@ class TranslateTaskTest {
...
 
             @Override
             void outOfDate(Action<? super InputFileDetails> outOfDateAction) {
-                proj.files('src/main/java/com/example/Main.java',
-                           'src/test/java/com/example/Verify.java').each {
-                    final File file ->
+                assert fileCountMultiplier >= 1
+                List<String> srcFiles = []
+                (1..fileCountMultiplier).each { int idx ->
+                    srcFiles.add("src/main/java/com/example/Main${idx}.java")
```
brunobowden: Replace `${idx}` with `$idx` to match style guide:

https://github.com/j2objc-contrib/j2objc-gradle/blob/master/CONTRIBUTING.md#style-guidelines

brunobowden:
Two minor nits and then we're done. Ping me once you've fixed them. I'll merge this in once all the CI builds pass. Thanks for your patience @himamis.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/TranslateTaskTest.groovy
himamis: The problem is that idx is followed by a ```.java```, and that is understood as a property of idx and gives the error:
```
No such property: java for class: java.lang.Integer
```
So either ```${idx}.java``` or ```$idx\\.java```?

brunobowden: I agree about `${idx}.java` - much easier for someone else to understand.

If that's the case, then the style guidelines look incorrect. It currently suggests:

    String message = "the count is $object.count"    // incorrect as it only prints "$object"

Hmmm... not so important though.

himamis:
@brunobowden The ```argsFile``` have been replaced with ```argFile```. If there is anything else, let me know. Thank you in helping fixing this!

brunobowden:
You're welcome, thank you too. I'll merge this in once the CI builds pass.

One last thing... Have you tested this to see if it makes your own build work now? The docs describe how you can test a locally built plugin for that build which was failing before:

https://github.com/j2objc-contrib/j2objc-gradle/blob/master/CONTRIBUTING.md#testing-on-your-own-gradle-project

Confirming that it works would make sure that this is going to solve your problem. I wanted to avoid the situation of releasing a new plugin version and then you find it doesn't solve your problem.

himamis:
I've tested it, and I can confirm that with my project it works.

brunobowden:
That's great @himamis. I'll merge this in the morning and then push a new public version for your delectation.

himamis:
Thank you @brunobowden. It was fun working on this project! :smile:

PR #576, opened 2016-01-07
FAQ Style fixup + build OS target

brunobowden:
- Fix code language style
- Add file path to code examples (e.g. shared/build.gradle)
- Add “how to” for building against specific OS version

- Change minOsVersion => minVersionOs (matching j2objcConfig setting)

### DIFF
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/PodspecTask.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/PodspecTask.groovy
@@ -132,7 +132,7 @@ class PodspecTask extends DefaultTask {
     @VisibleForTesting
     static String genPodspec(String podname, String publicHeadersDir, String resourceDir,
                              String libDirIos, String libDirOsx, String libDirWatchos,
-                             String minIosVersion, String minOsxVersion, String minWatchosVersion,
+                             String minVersionIos, String minVersionOsx, String minVersionWatchos,
                              String libName, String j2objcHome) {
 
         // Relative paths for content referenced by CocoaPods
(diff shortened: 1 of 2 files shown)


brunobowden:
@ph1lb4 - added an FAQ to explain how to set the OS. Though perhaps we should default to a more modern OS, e.g. iOS 8.0 for example.

@himamis - please take a look.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/PodspecTask.groovy
```diff
@@ -176,9 +176,9 @@ class PodspecTask extends DefaultTask {
                "    'LIBRARY_SEARCH_PATHS' => '$j2objcHome/lib'\n" +
                "  }\n" +
                 // http://guides.cocoapods.org/syntax/podspec.html#deployment_target
-               "  spec.ios.deployment_target = '$minIosVersion'\n" +
-               "  spec.osx.deployment_target = '$minOsxVersion'\n" +
-               "  spec.watchos.deployment_target = '$minWatchosVersion'\n" +
+               "  spec.ios.deployment_target = '$minVersionIos'\n" +
```
himamis: The style doesn't match the one in the guidelines:
```"target = '$minVersionIos'\n"``` instead of ```"target = $minVersionIos\n"```.

ph1lb4:
@brunobowden great idea to put it in the FAQ's

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/PodspecTask.groovy
brunobowden: @himamis - not sure I understand this. This line is for the output to the podspec, which is read by Cocoapods and has it's own parsing requirements. Infact, the podspec is essentially Ruby, so it needs the quotes around strings (single or double). See Cocoapods docs with example at the bottom:

https://guides.cocoapods.org/syntax/podspec.html#deployment_target

himamis: Sorry, I've forgot to check the context, you're right.

advayDev1:
ex post facto lgtm

Issue #482, opened 2015-10-04
 object file was built for newer iOS version (8.3) than being linked (8.0).


advayDev1:
We arbitrarily chose 8.3 as the OS version for our native compilation step.  This should be configurable via j2objcConfig.

cc @kirkvogen

ph1lb4:
how do you actually specify the minimum version in the build.gradle file?
Simply writing `minIosVersion = '8.0'` in the j2objcConfig method does not work

brunobowden:
`minIosVersion` exists in the code but isn't a setting for j2objcConfig. I think you meant this instead:

    j2objcConfig {
        minVersionIos '8.0'
    }

You can see the name of the setting and the default value of "6.0" here:

https://github.com/j2objc-contrib/j2objc-gradle/blob/master/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy#L636

ph1lb4:
Thanks, that was the confusion!

PR #584, opened 2016-01-10
Default minVersion: iOS => 8.3, watchOS => 2.0

brunobowden:
- Helps with avoiding #482


brunobowden:
@ph1lb4 - this should help you avoid this problem in the future

ph1lb4:
Thanks! We did set it by hand in the meantime which also worked fine.

brunobowden:
v0.6.0 of the plugin has been published, please try this out and report back any issues

Issue #588, opened 2016-01-19
Podfile configurations customization


danieldickison:
Instead of the current hardcoded `:configuration => ['Debug']` and `:configuration => ['Release']`, it would be nice if these configuration names were customizable. We have a separate configuration for beta releases on HockeyApp and my current workaround is to edit the Podfile so the j2objc method includes `:configuration => ['Release', 'Preview']`

I think this would go somewhere around here (but I'm not very familiar with groovy):
https://github.com/j2objc-contrib/j2objc-gradle/blob/459d2c70a8fbabc26a440cc3ad284d7d0bf7758e/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy#L487

brunobowden:
@danieldickison - this sounds reasonable. Your pointing to an old version of the code from October so best to use something more current. For example:

https://github.com/j2objc-contrib/j2objc-gradle/blob/master/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy#L669

To do this, you'll need to add something to J2objcConfig.groovy to allow someone to specify the configurations:

    List<String> podfileDebugConfigurations = ['Debug']
    List<String> podfileReleaseConfigurations = ['Release']

Please add the following example to both the J2objcConfig.groovy comments AND the FAQ:

    j2objcConfig {
        podfileReleaseConfigurations += ['Preview']
        ...
    }

Within XcodeTask.groovy, you need to add reading of the config around line 91 (see prior lines). Then at line 669, print the array with "[${podfileReleaseConfigurations.join(',')}]" and update the unit tests.

Send me the PR and I'll take a look. This should be pretty straight forward.

Bruno

brunobowden:
Add the configuration to line 728 in J2objcConfig.groovy.

danieldickison:
Thanks for the pointers. Just created pull request #590 with your suggested changes.

danieldickison:
Fixed in #590

PR #590, opened 2016-01-20
Customization of Xcode build configuration names

danieldickison:
Patch for issue #588.

Adds two new configs, `podfileDebugConfigurations` and `podfileReleaseConfigurations`, to allow customization of the Xcode build configuration names that are specified in the Podfile as linking with the debug and release translated libs. These default to the previously hardcoded `['Debug']` and `['Release']`.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -32,6 +32,7 @@ Paste the results below, replacing existing contents.
 - [Why is my clean build failing?](#why-is-my-clean-build-failing)
 - [How do I include Java files from additional source directories?](#how-do-i-include-java-files-from-additional-source-directories)
 - [How do I develop with Swift?](#how-do-i-develop-with-swift)
+- [How do I specify additional Xcode build configurations?](#how-do-i-specify-additional-xcode-build-configurations)
 - [How do I manually configure the Cocoapods Podfile?](#how-do-i-manually-configure-the-cocoapods-podfile)
 - [How do I manually configure my Xcode project to use the translated libraries?](#how-do-i-manually-configure-my-xcode-project-to-use-the-translated-libraries)
 - [How do I update my J2ObjC translated code from Xcode?](#how-do-i-update-my-j2objc-translated-code-from-xcode)
@@ -333,6 +334,27 @@ you'd like to access from Swift code.
 // Included from `shared/build/j2objcOutputs/src/main/objc`
 #import "MyClassOne.h"
 #import "MyClassTwo.h"
+```
+
+
+### How do I specify additional Xcode build configurations?
+
+Set `xcodeDebugConfigurations` and `xcodeReleaseConfigurations` to specify which build
+configurations should link the debug and release builds of the translated libs, respectively.
+These default to `['Debug']` and `['Release']` which correspond to Xcode's default build
+configuration names. Setting either of these to an empty array will omit the respective pod
+line from the "pod method".
+
+For example, if you have a build configuration called "Beta" for sending to internal testers
+and one called "Preview" for doing ad-hoc distribution:
+
+```gradle
+// File: shared/build.gradle
+j2objcConfig {
+    xcodeDebugConfigurations += ['Beta']
+    xcodeReleaseConfigurations += ['Preview']
+    ...
+}
 ```
 
 
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
@@ -725,6 +725,33 @@ class J2objcConfig {
      */
     boolean xcodeTargetsManualConfig = false
 
+    /**
+     * The Xcode build configurations which should link to the generated debug libraries.
+     * If set to an empty array, the Debug configuration will be omitted from the "pod method".
+     * <p/>
+     * For example:
+     * <pre>
+     * j2objcConfig {
+     *     xcodeDebugConfigurations += ['Beta']
+     *     ...
+     * }
+     * </pre>
+     */
+    List<String> xcodeDebugConfigurations = ['Debug']
+
+    /**
+     * The Xcode build configurations which should link to the generated release libraries.
+     * If set to an empty array, the Release configuration will be omitted from the "pod method".
+     * <p/>
+     * For example:
+     * <pre>
+     * j2objcConfig {
+     *     xcodeReleaseConfigurations += ['Preview']
+     *     ...
+     * }
+     * </pre>
+     */
+    List<String> xcodeReleaseConfigurations = ['Release']
 
     protected boolean finalConfigured = false
     /**
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
@@ -511,6 +511,14 @@ class Utils {
         return null
     }
 
+    static String toQuotedList(List<String> listString) {
+        if (listString.isEmpty()) {
+            return ''
+        } else {
+            return "'${listString.join("','")}'"
+        }
+    }
+
     @VisibleForTesting
     static String projectExecLog(
             ExecSpec execSpec, ByteArrayOutputStream stdout, ByteArrayOutputStream stderr,
--- a/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy
+++ b/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy
@@ -87,6 +87,15 @@ class XcodeTask extends DefaultTask {
     @Input
     String getMinVersionWatchos() { return J2objcConfig.from(project).minVersionWatchos }
 
+    // TODO: remove '2' suffix
+    // The '2' suffix here is purely to get around what appears to be a groovy compiler bug.
+    // groovy output includes: "BUG! exception in phase 'instruction selection'"
+    @Input
+    List<String> getXcodeDebugConfigurations2() { return J2objcConfig.from(project).xcodeDebugConfigurations }
+    @Input
+    List<String> getXcodeReleaseConfigurations2() { return J2objcConfig.from(project).xcodeReleaseConfigurations }
+
+
     @OutputFile
     File getPodfileFile() {
         return project.file(new File(getXcodeProjectDir(), '/Podfile'))
@@ -101,11 +110,16 @@ class XcodeTask extends DefaultTask {
         String projectName
         File podspecDebug
         File podspecRelease
+        List<String> xcodeDebugConfigurations
+        List<String> xcodeReleaseConfigurations
 
-        PodspecDetails(String projectNameIn, File podspecDebugIn, File podspecReleaseIn) {
+        PodspecDetails(String projectNameIn, File podspecDebugIn, File podspecReleaseIn,
+                       List<String> xcodeDebugConfigurationsIn, List<String> xcodeReleaseConfigurationsIn) {
             projectName = projectNameIn
             podspecDebug = podspecDebugIn
             podspecRelease = podspecReleaseIn
+            xcodeDebugConfigurations = xcodeDebugConfigurationsIn
+            xcodeReleaseConfigurations = xcodeReleaseConfigurationsIn
         }
 
         String getPodMethodName() {
@@ -259,7 +273,8 @@ class XcodeTask extends DefaultTask {
         logger.debug("${proj.getName()} project podspecs: " +
                      "${j2objcPodspec.getPodNameDebug()}, ${j2objcPodspec.getPodNameRelease()}")
         podspecs.add(new PodspecDetails(proj.getName(),
-                j2objcPodspec.getPodspecDebug(), j2objcPodspec.getPodspecRelease()))
+                j2objcPodspec.getPodspecDebug(), j2objcPodspec.getPodspecRelease(),
+                getXcodeDebugConfigurations2(), getXcodeReleaseConfigurations2()))
 
         J2objcConfig j2objcConfig = proj.getExtensions().getByType(J2objcConfig)
         j2objcConfig.getBeforeProjects().each { Project beforeProject ->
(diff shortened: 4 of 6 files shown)


FAQ.md
```diff
@@ -336,6 +337,24 @@ you'd like to access from Swift code.
 ```
 
 
+### How do I specify additional Xcode build configurations?
```
brunobowden: It's a podfile configuration rather than Xcode, so title it:

    How do I specify additional Podfile build configurations for Xcode?

Then also update the title and link in the TOC.

brunobowden: Actually, this is a configuration inside Xcode as well? I'm wondering if this should be named xcodeDebugConfigurations and xcodeReleaseConfigurations instead? Please comment on this.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy
```diff
@@ -101,11 +107,16 @@ class XcodeTask extends DefaultTask {
         String projectName
         File podspecDebug
         File podspecRelease
+        List<String> debugConfigurations
+        List<String> releaseConfigurations
 
-        PodspecDetails(String projectNameIn, File podspecDebugIn, File podspecReleaseIn) {
+        PodspecDetails(String projectNameIn, File podspecDebugIn, File podspecReleaseIn,
+                       List<String> debugConfigurationsIn, List<String> releaseConfigurationsIn) {
```
brunobowden: Use variable name `podfileReleaseConfigurations`... I like to reuse the name so a single search finds all the relevant code and comments, rather than having to follow the chain of logic.

danieldickison:
Not sure what the best name for this would be: the build configurations are things defined in the Xcode project, but the Podfile references those configuration names. I guess that's the same situation as target names and those configs are named e.g. `xcodeTargetsIos`, so I can rename these to be `xcodeDebugConfigurations` and `xcodeReleaseConfigurations` if you prefer.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy
```diff
@@ -113,6 +124,14 @@ class XcodeTask extends DefaultTask {
             return "j2objc_$projectName".toString().replaceAll(/[^a-zA-Z0-9]/, '_')
         }
 
+        String getJoinedDebugConfigurations() {
+            return debugConfigurations.collect {"'$it'"}.join(', ')
```
brunobowden: Again reuse the `podfileDebugConfigurations` so a single search finds all the references. You can append or prepend that variable name as needed.

Extract this to a utility method in Utils.groovy and add an appropriate unit test:

    def toQuotedList(List<String> listString) {
        return "'${listString.toArray().join("','")}'"
    }

This does the magic that you're looking for ;-) If you're doing stuff with iterators in the future you need to add explicit types. See the style guidelines on this.

https://github.com/j2objc-contrib/j2objc-gradle/blob/master/CONTRIBUTING.md#style-guidelines

brunobowden: Then it can be:

    podMethodLines.add("    pod '$podspecDebugName', :configuration => [${Utils.toQuotedList(xcodeDebugConfigurations))}], :path => '$pathDebug'".toString())

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTaskTest.groovy
```diff
@@ -1107,6 +1112,27 @@ class XcodeTaskTest {
...
+                        new File('/SRC/PROJ/BUILD/j2objc-PROJ-release.podspec'),
+                        ['Debug', 'Beta'], ['Release', 'Preview'])
+        List<String> podMethodLines = XcodeTask.podMethodLines(
+                podspecDetails,
+                new File('/SRC/ios/Podfile'))
+
+        List<String> expectedLines = [
+                "def j2objc_PROJ",
+                "    pod 'j2objc-PROJ-debug', :configuration => ['Debug', 'Beta'], :path => '../PROJ/BUILD'",
+                "    pod 'j2objc-PROJ-release', :configuration => ['Release', 'Preview'], :path => '../PROJ/BUILD'",
```
brunobowden: Good, this is the test you need to have.

brunobowden:
More important than what I prefer is what makes best sense for other engineers. I had mistakenly thought this was only for the Podfile... but since this is an Xcode configuration then it feel like it should be consistent with the other settings. So xcodeDebugConfigurations and xcodeReleaseConfigurations sounds like a better choice. Feel free to push back if you feel differently.

If you're good with this, update the methods, tests and FAQ and it should be good for a final review.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy
danieldickison: Is there a reason for the `toArray()` there, as opposed to calling join on the List directly? I.e.

    return "'#{listString.join("','")}'"

danieldickison:
Yup I agree the xcode* naming makes more sense now that I've seen more of the context. I'll make those changes and update the pull request.

I'm now getting a bizarre build error from groovy on identical code that previously compiled :/ I'll update when I get this sorted, but I'm curious if you've seen something like this.

```* What went wrong:
Execution failed for task ':compileGroovy'.
> BUG! exception in phase 'instruction selection' in source unit '/Users/daniel/bc/j2objc-gradle/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy' Declaring class for method call to 'java.util.List getXcodeDebugConfigurations()' declared in com.github.j2objccontrib.j2objcgradle.tasks.XcodeTask was not matched with found receiver com.github.j2objccontrib.j2objcgradle.tasks.XcodeTask$PodspecDetails. This should not have happened!
```

brunobowden:
I didn't think it worked without `toArray()` but bonus points if you
successfully make it simpler. Try clean before another build.

Ping me when it's ready and I'll do a final review.

danieldickison:
@brunobowden that's ready for review.

As mentioned in the commit message, I couldn't get around the strange groovy error so I left a workaround in, renaming the offending methods with a "2" suffix. In theory those methods could be renamed back without the "2" suffix, but I'm unable to get that to compile locally on my machine even after clean and wiping out gradle cache directories (I guess there must be a groovy compiler cache directory I'm missing since it persists after rebooting).

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy
```diff
@@ -87,6 +87,14 @@ class XcodeTask extends DefaultTask {
     @Input
     String getMinVersionWatchos() { return J2objcConfig.from(project).minVersionWatchos }
 
+    // The '2' suffix here is purely to get around what appears to be a groovy compiler bug.
```
brunobowden: Make this a TODO to revert to the proper name without '2'

FAQ.md
```diff
@@ -336,6 +337,24 @@ you'd like to access from Swift code.
...
+configurations should link the debug and release builds of the translated libs, respectively.
+These default to `['Debug']` and `['Release']` which correspond to Xcode's default build
+configuration names.
+
+For example, if you have a build configuration called "Preview" for doing ad-hoc distribution:
+
+```gradle
+// File: shared/build.gradle
+j2objcConfig {
+    xcodeReleaseConfigurations += ['Preview']
```
brunobowden: If you think it fits, have both configs:

    xcodeDebugConfigurations += ['Beta']
    xcodeReleaseConfigurations += ['Preview']

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy
```diff
@@ -666,8 +680,8 @@ class XcodeTask extends DefaultTask {
         // Either update pod line in place or add line if pod doesn't exist
         List<String> podMethodLines = new ArrayList<>()
         podMethodLines.add("def ${podspecDetails.getPodMethodName()}".toString())
-        podMethodLines.add("    pod '$podspecDebugName', :configuration => ['Debug'], :path => '$pathDebug'".toString())
-        podMethodLines.add("    pod '$podspecReleaseName', :configuration => ['Release'], :path => '$pathRelease'".toString())
+        podMethodLines.add("    pod '$podspecDebugName', :configuration => [${Utils.toQuotedList(podspecDetails.xcodeDebugConfigurations)}], :path => '$pathDebug'".toString())
```
brunobowden: These lines are getting along, so extract it:

    String xcodeDebugConfigs = Utils.toQuotedList(podspecDetails.xcodeDebugConfigurations)

Then in the string:

    "    pod '$podspecDebugName', :configuration => [$xcodeDebugConfigs], :path => '$pathDebug'".toString()

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/UtilsTest.groovy
```diff
@@ -498,6 +498,13 @@ class UtilsTest {
     }
 
     @Test
+    void testToQuotedString() {
+        assert "'a','b','c'" == Utils.toQuotedList(['a', 'b', 'c'])
+        assert "'abc'" == Utils.toQuotedList(['abc'])
```
brunobowden: Add another case:

    assert "''" == Utils.toQuotedList([''])

I'm wondering if `Utils.toQuotedList([])` should be a empty string or not? I think it should be to distinguish it from the case above. That means changing the last test and changing the method also.

    assert "" == Utils.toQuotedList([])

brunobowden:
These should be my final comments.

For the `2` method, try deleting the gradle cache:
https://github.com/j2objc-contrib/j2objc-gradle/blob/master/FAQ.md#start-here-for-debugging-aka-its-not-working-aka-dont-panic

Add copyright grant:
https://github.com/j2objc-contrib/j2objc-gradle/blob/master/CONTRIBUTING.md#conditions-for-accepting-pull-requests

Collapse the commits:
https://github.com/j2objc-contrib/j2objc-gradle/blob/master/CONTRIBUTING.md#preparing-your-pull-request-for-submission

brunobowden:
Finally, what should happen for `xcodeDebugConfigurations = []`, i.e. no configurations at all? Should that give an error? Will Cocoapods accept it? Try that out and add a comment explaining whatever logic you suggest.

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/UtilsTest.groovy
danieldickison: Yeah, I can see this being ok either way (is the method always expected to return a quoted string or is an empty string acceptable?) but having the empty string return value does work better for treating the output as the contents of a Ruby array literal so I'll go with that.

danieldickison:
Re: `xcodeDebugConfigurations = []`

If the Podfile contains `:configurations => []`, the pod dependency gets applied to _all_ configurations. This would be unexpected, so I propose that the entire pod line is excluded when the array is empty.

On the other hand, CocoaPods reports an error when `:configurations => ['']`, which I think is fine since, in fact, an empty string is an invalid configuration name.

brunobowden:
I feel that `toQuotedList([])` should be an empty string.

It makes sense that for an empty list, the Podfile line is removed
entirely. Add that to the FAQ and the comment in J2objcConfig.

danieldickison:
Ok I think I incorporated all the comments. I went ahead and deleted the "2" suffix from those method names even though I wasn't able to solve the mystery groovy/gradle error locally (I tried deleting the various cache directories, etc, to no avail. Very strange).

danieldickison:
Looks like the AppVeyor failure is the same groovy error I'm seeing locally. I'll put the "2" workaround back in there for now, but I'm at a loss as to why these particular method declarations fail.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/Utils.groovy
```diff
@@ -511,6 +511,11 @@ class Utils {
         return null
     }
 
+    static String toQuotedList(List<String> listString) {
+        if (listString.isEmpty()) return ''
+        else return "'${listString.join("','")}'"
```
brunobowden: We've adopted the style of separating conditional logic over multiple lines:

    if (listString.isEmpty()) {
        return ''
    } else {
        return "'${listString.join("','")}'"
    }

src/test/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTaskTest.groovy
```diff
@@ -1107,6 +1112,43 @@ class XcodeTaskTest {
...
+                podspecDetails,
+                new File('/SRC/ios/Podfile'))
+        List<String> expectedLines = [
+                "def j2objc_PROJ",
+                "    pod 'j2objc-PROJ-debug', :configuration => ['Debug','Beta'], :path => '../PROJ/BUILD'",
+                "    pod 'j2objc-PROJ-release', :configuration => ['Release','Preview'], :path => '../PROJ/BUILD'",
+                "end"]
+        assert expectedLines == podMethodLines
+
+        // Empty configurations list should omit pod line, while an empty string should be included
```
brunobowden: This should be separate tests:

1. testUpdatePodMethods_customConfigurations
1. testUpdatePodMethods_emptyConfigurations
1. testUpdatePodMethods_emptyStringConfigurations

For the last case with `['']`, I think that should be an error. Is there any way it can be valid for a Podfile? I certainly think Xcode wouldn't allow `''` as a valid configuration name.

brunobowden: If it's an error, then you should throw an exception and then catch it in test. See this example elsewhere in the file:

    expectedException.expect(InvalidUserDataException.class)

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
```diff
@@ -725,6 +725,31 @@ class J2objcConfig {
...
+     *     xcodeDebugConfigurations += ['Beta']
+     *     ...
+     * }
+     * </pre>
+     */
+    List<String> xcodeDebugConfigurations = ['Debug']
+
+    /**
+     * The Xcode build configurations which should link to the generated release libraries.
+     * <p/>
```
brunobowden: Add comment:

    If set to an empty array, then the Debug configuration will be omitted from the "pod method"

brunobowden:
Given the flakiness, it makes sense to retain the `2`. Very strange and
probably fixed in a later version of Gradle.

src/main/groovy/com/github/j2objccontrib/j2objcgradle/J2objcConfig.groovy
danieldickison: Made these changes, though I skipped adding a check for empty config names. That seems like a contrived test case and I think having CocoaPods fail on it is good enough if someone were to try using an empty name.

brunobowden:
@danieldickison - thanks for a great change with good code. I'm happy to have merged this in. We'll probably work on some other fixes for J2ObjC 1.0 before spinning a new version.

danieldickison:
Sounds good. Thanks for the code reviews!

Issue #496, opened 2015-10-07
Can't Run j2objcXcode Task


MohammadHejazi:
I tried many things to make this task work but always get:

    Execution failed for task ':common:j2objcXcode'.
    org.gradle.api.InvalidUserDataException: Command Line Failed: pod install
    org.gradle.process.internal.ExecException: A problem occurred starting process 'command 'pod''

    Please install CocoaPods to use j2objcXcode (https://cocoapods.org):
        sudo gem install cocoapods

I tried reinstalling CocoaPods, RubyGems, Ruby and RVM, using system Ruby, using rbenv instead of RVM with no success. Everything works in terminal but not inside Android Studio.

The problem seems to be with task environment, but how could this problem be resolved?

brunobowden:
Is there any surrounding debug information which gives a clue? There should
be some surrounding lines with "Standard Output:", "Error Output:",
"Working Dir:" and "Cause:". Also try the "--debug" flag on the gradlew
command.

I'm wondering if this is a permissions issue. So check that as well. If you
run "pod install" from your "ios" directory, does it work correctly as
expected?

MohammadHejazi:
I don't think it's a permissions issue, but environment issue mostly because it can't find pod executable to run the command.

I checked the environment in the log, and it's different from the one in terminal and doesn't contain any reference to ruby stuff. So maybe this is the problem, but I wonder how it works for other people?

I enabled the debug option and this is the output:
```
[DEBUG] [org.gradle.api.Task] Pod exists at path: ../iOS
[DEBUG] [org.gradle.api.Task] XcodeTask - projectExec - pod install:
[INFO] [org.gradle.process.internal.DefaultExecHandle] Starting process 'command 'pod''. Working directory: /TestJ2OBJC/iOS Command: pod install
[DEBUG] [org.gradle.process.internal.DefaultExecHandle] Environment for process 'command 'pod'': {SHELL=/bin/bash, EnvAgent_vars=..., ANDROID_NDK_ROOT=..., TMPDIR=..., __CF_USER_TEXT_ENCODING=..., PATH=/usr/bin:/bin:/usr/sbin:/sbin, XPC_FLAGS=0x0, USER=..., HOME=..., XPC_SERVICE_NAME=com.google.android.studio.4832, LOGNAME=..., Apple_PubSub_Socket_Render=..., SSH_AUTH_SOCK=...}
[...]

PR #592, opened 2016-01-28
Update FAQ to resolve pod install command failure

danielrees18:
Addresses the problem discussed in issue #496

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -47,7 +47,8 @@ Paste the results below, replacing existing contents.
 - [How do I develop on Windows or Linux?](#how-do-i-develop-on-windows-or-linux)
 - [How do I fix missing required architecture linker warning?](#how-do-i-fix-missing-required-architecture-linker-warning)
 - [How do I fix Undefined symbols for architecture linker warning?](#how-do-i-fix-undefined-symbols-for-architecture-linker-warning)
-- [How do I solve the Eclipse error message Obtaining Gradle model...?](#how-do-i-solve-the-eclipse-error-message-obtaining-gradle-model)
+- [How do I fix the Eclipse error message Obtaining Gradle model...?](#how-do-i-fix-the-eclipse-error-message-obtaining-gradle-model)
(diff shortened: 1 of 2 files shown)


FAQ.md
```diff
@@ -48,6 +48,7 @@ Paste the results below, replacing existing contents.
 - [How do I fix missing required architecture linker warning?](#how-do-i-fix-missing-required-architecture-linker-warning)
 - [How do I fix Undefined symbols for architecture linker warning?](#how-do-i-fix-undefined-symbols-for-architecture-linker-warning)
 - [How do I solve the Eclipse error message Obtaining Gradle model...?](#how-do-i-solve-the-eclipse-error-message-obtaining-gradle-model)
+- [How do I resolve Command Line Failed: pod install](#how-do-i-resolve-command-line-failed-pod-install)
```
brunobowden: Change to "How do I fix..." to make it consistent with previous FAQ items. This requires updating the title in the item also.

Also, please also replace the "solve" with "fix" on the previous question.

brunobowden:
You should also update the error message, so that when others encounter this issue, then it direct the user to the FAQ item. Then the FAQ item needs to be expanded to explain that it's one of two errors: 1) Cocoapods not installed and 2) The path issue you explained.

Here's the section to update:
https://github.com/j2objc-contrib/j2objc-gradle/blob/master/src/main/groovy/com/github/j2objccontrib/j2objcgradle/tasks/XcodeTask.groovy#L246

Here's what to replace it with:
```
String message =
    exception.toString() +
    '\n' +
    'Please see this FAQ item for help:\n'
    'https://github.com/j2objc-contrib/j2objc-gradle/blob/master/FAQ.md#how-do-i-resolve-command-line-failed-pod-install'
```

brunobowden:
Please see this notice about the future of the J2ObjC Gradle Plugin:
https://groups.google.com/forum/#!topic/j2objc-discuss/jQ_eyW9nxmQ

brunobowden:
Closing pull request as the plugin is no longer being actively maintained. You're welcome to fork the project.

PR #618, opened 2016-05-19
J2ObjC 1.0.2 Support

kirkvogen:
Added support for J2ObjC version 1.0.2 and Guava. J2ObjC is [phasing out built-in Guava support](https://groups.google.com/d/msg/j2objc-discuss/9rBpyQ97ECM/xzOs1t0pAgAJ). As part of this phase out, the location of the headers has changed slightly, which cause it to not work with j2objc-gradle plugin. The plugin could have been update to take into account this new header location. Instead, this version of the plugin no longer uses the built-in Guava, since J2ObjC can compile and link the vanilla Guava.

The rest of the changes largely involved upgrading version numbers in plugin code so that it will accept newer versions of J2ObjC.

The system test in systemTests/multiProject1 exists to show that a project can depend on a J2ObjC-provided dependency. It currently used built-in Guava. Since this version of j2objc-gradle no longer supports built-in Guava, this system test no longer uses Guava as its built-in, but instead uses Google Protocol Buffers. Writing a test that uses Protocol Buffers was challenging. The original plan was to write a test that does serialization, but I wasn't able to get this to work as it appears that the J2ObjC is missing some headers for Protocol Buffers. Next, I tried using the WiringFormat class, since it has a little utility method that could be used for the test. This option failed as the Objective-C compiler had trouble with macros reference in the WiringFormat.h header. In the end, the classes basically do a Class.forName() to show that the code in Protocol Buffers can successfully be loaded. This at least shows that the built-in Protocol Buffers can be listed as a Gradle dependency and used successfully.


brunobowden:
Please see this notice about the future of the J2ObjC Gradle Plugin:
https://groups.google.com/forum/#!topic/j2objc-discuss/jQ_eyW9nxmQ

brunobowden:
Closing pull request as the plugin is no longer being actively maintained. You're welcome to fork the project.

